PR #64, opened 2018-01-05
Lazily Require Chokidar

lukeed:
Chokidar has a heavy footprint & is an unnecessary price to pay in production since it's unused. The benefit is that initial boot time is reduced... _slightly_ but it's still a small win.

I discovered the same benefit during Taskr/Fly development: https://github.com/lukeed/taskr/issues/236


Rich-Harris:
Makes total sense — thank you 👍

PR #65, opened 2018-01-05
Add init option to control prefetching of routes

EmilTholin:
It might be a useful addition to add options to the `init` function. I think being able to control the initial prefetching of all routes would be powerful, because I like to e.g. add inlined image previews in my JS files, which results in quite a large initial download if all JS is fetched.


EmilTholin:
Closing this in favor of the much more versatile solution suggested in https://github.com/sveltejs/sapper/issues/160.

Rich-Harris:
Thanks for the inspiration, and sorry I never got round to providing feedback on this PR — I liked the idea, just wanted to avoid 'optionitis', and couldn't see the (in retrospect fairly obvious) solution until today!

EmilTholin:
No problem! Totally agree on the 'optionitis' problem, your solution is much better.

Issue #62, opened 2018-01-04
Remove async/await


Rich-Harris:
Currently, Sapper requires Node 8, which means you can't run it on AWS Lambda and what-have-you

EmilTholin:
Would you like to remove async/await from the tests as well?

Rich-Harris:
I suppose we'd need to, if we're to guard against regressions. If we're not testing on Node 6 it'd be very easy to accidentally stop supporting it

thgh:
I would vote to use `@std/esm` instead. Not sure what the performance impact would be on Lambda.

Rich-Harris:
@thgh `@std/esm` is just about supporting `import` and `export`, it doesn't have any bearing on whether `async` and `await` can be used. This issue should be closed though, it was fixed in #67

PR #66, opened 2018-01-05
Adds a `sapper extract` CLI command, which scrapes the server to run …

### LINKED ISSUE #9: Export static site
For static sites (e.g. svelte.technology), it might be nice to have an equivalent of `next export`. Could be done like so:

```js
// scripts/export.js
const exporter = require('sapper/export');

const pages = [
  '/',
  '/guide',
  '/repl',
  '/blog'
].concat(getListOfBlogPostsSomehow());

exporter.export(pages, 'dist').then(() => {
  console.log('done');
});
```

That would run `server.js` in a separate process and basically just `curl` those URLs into `dist`, and copy over the `assets` folder and the files generated by the compiler.

freedmand:
…as a static website starting at the site's root. Resolves #9

TESTED=Basic unit test ensuring relevant routes are added.


lib/utils/extract.js
```diff
@@ -0,0 +1,223 @@
...
+ * well as copying assets.
+ * @param {?Array<string>=} includeUrls If non-null, a set of additional URLs to
+ *     scrape in the extraction. This should only be set if there are routes
+ *     that cannot be reached from the root.
+ * @param {?Array<string>=} excludeUrls If non-null, a set of URLs to avoid
+ *     scraping in the extraction.
+ * @param {number=} extractionDir The directory in which to place the extracted
+ *     output.
+ */
+module.exports = async function(includeUrls = null, excludeUrls = null,
```
EmilTholin: Super cool stuff! There is an [issue to move away from `async/await`](https://github.com/sveltejs/sapper/issues/62). Maybe it's best to preemptively not use it?

test/common/test.js
```diff
@@ -306,6 +308,57 @@ function run(env) {
...
+						'api/blog/index.html',
+
+						'blog/a-very-long-post/index.html',
+						'api/blog/a-very-long-post/index.html',
+
+						'blog/how-can-i-get-involved/index.html',
+						'api/blog/how-can-i-get-involved/index.html',
+
+						'blog/how-is-sapper-different-from-next/index.html',
+							'api/blog/how-is-sapper-different-from-next/index.html',
```
EmilTholin: A little funky indentation.

lib/utils/extract.js
freedmand: Thanks! Hmm, there are a lot of `await` calls in Sapper already. I'm happy to change out my `await` and `async` calls, but do you think the onus of not being able to run on platforms like AWS Lambda because of the Node version is on the platforms to fix?

await_in_sapper

freedmand:
Just realized this is currently not producing a navigable website in the `test/app` directory. Working to fix now

lib/utils/extract.js
EmilTholin: @Rich-Harris has more insight on that than me, but I can definitely see the appeal of supporting lower versions of Node. 
I just have all the `async/await` in `test.js` left to replace in a PR, so no worries!

freedmand: I've removed `async/await` in my changes to keep you from having to do the work :)

lukeed:
FWIW, my PR fails within Travis the same way & it didn't change anything in relation to that test. Maybe it's off?

freedmand:
If you look back in the pull request [history](https://travis-ci.org/sveltejs/sapper/pull_requests), Travis is mostly failing, so I agree something seems off. I'm confident my PR doesn't contribute new failures.

(On that note, there seems to be some amount of non-deterministic behavior in the tests. When my computer processor was overloaded, some of the tests sporadically failed, maybe since they're based on timeouts?)

Rich-Harris:
Thanks very much for this, and apologies for leaving it unattended a while.

I'm actually having a few issues getting it to work correctly — if you serve the `test/app/dist` directory and click around, it fails to navigate. Possibly a few things going on here:

* Changing `/foo` to `/foo/`, backed by `/foo/index.html`, might be confusing the router
* Things like `/api/blog/what-is-sapper` (which contains JSON data) is being written as `/api/blog/what-is-sapper/index.html` which is incorrect

Will keep chipping away at it (I added some commits to this PR already, mostly so that it doesn't try to extract without first building the site in prod mode), but noting it in here in case it you already know the answer 😀

freedmand:
I'll have time to look into it more in a few hours. I suspect it may depend on the static router. I think `python -m SimpleHTTPServer` and Netlify were cool redirecting things automatically so that it would work. Are you using `serve` or something else (I haven't looked into other things)?

Rich-Harris:
I'm using `serve`, yeah. FYI I'm investigating a slightly different approach that involves using the server's entry point and generating the pages *without* spinning up a server — early results are promising, and it should be blazing fast!

freedmand:
That would be great. There is no reason to spin up a server (while researching this I was shocked there's no *non-hacky* way to get Express to feed you pages without consuming a port). For this PR, I figured getting a messy approach first that can be iteratively improved would be a good starting point.

Still need to figure a good solution for having the extract pipeline know about server routes that are called in JS without any extra config options (like the `/api/` routes in the example project) -- does anyone have thoughts on that?

freedmand:
Just to be clear, would the server actually run frontend JS code to figure out what it needs? Let me know if I can help

Rich-Harris:
My current thinking is that we *wouldn't* run client-side code, because that opens a big can of worms. Instead, we rely on pages specifying their data dependencies in `preload`.

In time, we could probably add a way of configuring additional resources to export, on top of the ones that are automatically discovered.

I opened a new PR on top of this one, since that was the easiest way to make additional changes — #88

PR #67, opened 2018-01-05
Remove async await

EmilTholin:
https://github.com/sveltejs/sapper/issues/62


EmilTholin:
Yeah, everything is working locally with `node@6.12.3`, but Travis is not to happy. It had some issues in my other PR as well, which had a few lines difference, so not quite sure what's wrong.

PR #71, opened 2018-01-06
[WIP] Fix Tests

lukeed:
Just a little attempt to see if this approach works better within Travis.

Good idea to always use `.end()` after `.evaluate()` as it ensures the queue is done before continuing. Also continuing the chain to `assert.equal` from within a `then` that follows.

The idea is that Travis is getting bottle-necked during the promise-chain. The chain resolves sooner than it should (why?) then the `await`ed value is incorrect.

If this works, I'll update the other possible offenders.

### DIFF
--- a/test/app/routes/show-url.html
+++ b/test/app/routes/show-url.html
@@ -2,10 +2,8 @@
 
 <script>
 	export default {
-		preload(foo) {
-			let url = foo.url;
-			console.log('> i am here', url, foo);
-			return { url };
+		preload({ url }) {
+			if (url) return { url };
 		}
 	};
 </script>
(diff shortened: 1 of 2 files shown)


lukeed:
Looks like the `preload` function is being called twice. 🤔 

Linking to the [second call](https://travis-ci.org/sveltejs/sapper/builds/325912847#L1354). The first one has the correct value.

lukeed:
Lol, I've done things I'm not proud of with Nightmare...

I sorta stopped on it, so WIP is still meant to be there. Travis was just being super strange & that double `preload` was throwing me off.

Feel free to close though, I can always pick this back up later. Otherwise I'll rebase/fix conflicts when can flush this out.

PR #72, opened 2018-01-07
Explicitly load style-loader/css-loader deps

nolanlawson:
This appears to fix sveltejs/sapper-template#27. It's based on the proposed workaround in https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/456#issuecomment-340287722

The downside is that the production size of `main.js` appears to increase from 4.2KB to 6.2KB. On the other hand... async style loading actually works. 😅 I don't see any more console errors while clicking around in the production app.

### DIFF
--- a/webpack/config.js
+++ b/webpack/config.js
@@ -6,7 +6,12 @@ module.exports = {
 	client: {
 		entry: () => {
 			return {
-				main: entry.client
+				main: [
+					entry.client,
+					// workaround for https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/456
+					'style-loader/lib/addStyles',
+					'css-loader/lib/css-base'
+				]
 			};
 		},
 


nolanlawson:
Hm, on second thought, this appears to fix the console errors but doesn't actually enable async style loading. I have some style in a subcomponent that isn't loaded when it's loaded asynchronously... this needs more investigation.

nolanlawson:
Nevermind, that other issue appears to be unrelated. I'll try to find a repro. In any case, this PR allows the app to actually load subpages, which is a more serious bug than not being able to load async styles.

nolanlawson:
Opened a separate issue, and it is indeed a separate issue: https://github.com/sveltejs/sapper/issues/79

Rich-Harris:
Thank you! Top sleuthing, not sure I would ever have figured that out. Seems a bit unfortunate, will have to see if there's a way around the extra bytes.

Issue #77, opened 2018-01-09
Support DELETE requests


jacobmischka:
I imagine (though I'm not sure) that Sapper supports all of the request types that express supports, but trying to export a DELETE handler currently won't work because `delete` is a reserved keyword, so `export function delete(){}` won't work.

Is there a clever workaround to trick js into letting me use `delete`? If not, could we add one to Sapper?

I thought maybe webpack harmony magic would let this work, but it doesn't seem to:

```js
export default {
	delete(req, res) {
		console.log('delete');
	}
};

```

Thanks! And great work as always, Sapper is fun.

lukeed:
Wouldn't you need to do this?

```js
export function delete(req, res) {
 // 
}
```

In CJS terms, your example would be available as `exports.default.delete` when you want `exports.delete`.

EmilTholin:
That's right @lukeed, but what you have written throws `Module parse failed: Unexpected token` because `delete` is a reserved keyword.

jacobmischka:
@lukeed Ah of course, sorry, that's what I meant and what I tried.

Crisfole:
I think most people handle this by using `_delete`, `del`, `remove` or `destroy`. Other more colorful options include `annihilate`, `smash`, `crush`.

Rich-Harris:
good point! I think `del` is probably the best option here:

```js
export function del(req, res) {
  // ...
}
```

Any objections?

jacobmischka:
Seems like the best option to me!

Rich-Harris:
Released 0.4 with `export function del(...){...}` support thanks to @EmilTholin

PR #80, opened 2018-01-10
Handle DELETE requests with del export

EmilTholin:
`export function delete(req, res) { /* ... */ }` throws a `Module parse failed: Unexpected token` error because `delete` is a reserved keyword.

This PR changes the handler of `DELETE` requests from `delete` to `del`, following the discussion in https://github.com/sveltejs/sapper/issues/77.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -177,7 +177,11 @@ function get_route_handler(fn) {
 					}
 
 					else {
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you! Added a test in https://github.com/sveltejs/sapper/commit/e3c047831a0edba9957a6482bd0cea228586dc09

PR #93, opened 2018-01-19
Hydrate on first load, not destroy

chexxor:
I installed the svelte-hackernews project and put a trace on this line and saw that `hydrate` is false on the first load. The `component` var in the runtime is still undefined, so `!!component` evaluates to `false`. To hydrate on the first load, when `component` is `undefined`, it should be `!component`.

Can squash, if you merge.


Rich-Harris:
Thank you! Have added a test for this — it seems Svelte's hydration doesn't quite work as expected in slotted content, which we'll have to look into.

PR #95, opened 2018-01-19
Remove Express Helpers

### LINKED ISSUE #94: Remove Express-specific Utilities
Express isn't listed as a `dependency` nor a `peer-dependency`, plus I know Sapper is meant to work with any HTTP server 😉 

There are a few instances where Sapper is using things like `res.set`, which is an Express shorthand, instead of `res.setHeader`.

Tracking all fo them down, then I'll open a PR to fix.

lukeed:
Replaces the accidental use of Express helpers with the native versions~

---

Closes #94


Rich-Harris:
thank you!

PR #96, opened 2018-01-20
Identify js and html route clashes

mrkishi:
This modifies `create_routes.js` slightly in order to catch clashing non-dynamic routes.

It's been also added to the unit tests.

(this would previously fail at runtime with multiple similarly named exports in the generated js)


Rich-Harris:
thank you!

Issue #92, opened 2018-01-17
sapper upgrade


Rich-Harris:
#91 proposes changing the required project structure so that `2xx.html` becomes `page.html`, and `4xx.html` and `5xx.html` are moved into the `routes` folder.

#86 removed `%sapper.main%` in favour of `%sapper.scripts%`.

These probably aren't the last times we'll ever change our minds about how something should be organised. Changes like these can be a pain for people upgrading, and in a worst case scenario can discourage them from doing so.

While I'm not a fan of the walled garden/black box/'toolkit' approach, for reasons I've [articulated](https://github.com/sveltejs/sapper/issues/68) [elsewhere](https://twitter.com/Rich_Harris/status/953005486969966595), it is nice to provide as seamless an upgrade path as possible.

One possible approach would be a `sapper upgrade` command that detects any deprecated constructs and fixes them with codemods etc:

```bash
$ yarn start

templates/2xx.html
<script src='%sapper.main%'> is unsupported — use %sapper.scripts% (without the <script> tag) instead
https://github.com/sveltejs/sapper/issues/86
   34 |        initialise the router -->
 > 35 |   <script src='%sapper.main%'></script>
      |                ^

Run `sapper upgrade` to fix this issue automatically.
```

Rich-Harris:
Since `degit` is just based on git repos, you could either pull the most recent master branch of sveltejs/sapper-template, or manually look at the commits to see what's changed. But `sapper upgrade` will be in the next release (I just merged #120)

Rich-Harris:
Closing, as this is implemented (though there's plenty more that could be done in terms of specific upgrade tasks)

Issue #91, opened 2018-01-17
Move error handling into pages


Rich-Harris:
...and out of templates. Great idea from @mrkishi — if our `routes` directory had `4xx.html` and `5xx.html` (or more specific codes, not that we're currently set up to use them) files, then we could use those in place of the current `templates/xxx.html` files. This has a number of advantages:

* No duplication — we can just have a single `templates/page.html`
* Those pages can themselves use components
* It gives us a sensible way to handle client-side navigation errors, rather than, err... discarding them which is what we currently do (yeah I know)

It would also simplify the internals slightly.

Rich-Harris:
A related thought. Currently, `server.js` is a CommonJS module, and the Sapper middleware loads the 'server app' generated from the contents of `routes` via the webpack config.

It turns out that's not ideal for a few reasons. Firstly, it's slightly confusing that `server.js` and `templates/main.js` are completely different beasts with different rules. Secondly, changes to `server.js` don't cause a server restart, because it's exempt from the filewatching mechanism. Thirdly, it adds a lot of complexity around (for example) interacting with user data generated by things like Passport — I've found myself throwing stuff onto the `req` object just to pass it into my server routes, because my server.js and my individual routes can't access shared modules for managing stuff in a more intelligent way. (My example is probably extremely specific, but I think it hints at a more general issue.)

So I'd like to experiment with a different, simpler structure:

```
├ package.json
├ server.js
├ webpack.client.config.js
├ webpack.server.config.js
├ app
│ ├ client.js
│ ├ server.js
│ ├ service-worker.js
│ └ template.html
├ assets
│ ├ # your files here
├ routes
│ ├ # your routes here
```

`app/server.js` would be able to import and interact with the same modules as `routes/**/*.[html|js]`.

I think that would be a little more intuitive than the status quo, and might yield benefits in terms of startup performance (for Lambda-type environments). There'd be some gotchas to work through though, e.g. we don't want to accidentally bundle `chokidar` in prod mode. And it's still not clear what we'd do about `service-worker.js` — does that get processed by webpack? We probably don't want a *third* webpack config...

ansarizafar:
Nextjs is now has Universal Webpack feature https://zeit.co/blog/next5. I think we can also simplify webpack configuration by using a similar feature.

LorbusChris:
[...]

PR #101, opened 2018-01-21
Restructure codebase

Rich-Harris:
This is designed to put the codebase on a somewhat more stable footing. The existing structure is completely nonsensical, which I suppose is what happens when you're figuring things out as you go along.

Now, instead of things like routes and templates being automatically created once the relevant modules are required (resulting in an unpredictable order of execution, and making things harder to test), everything is a bit more functional and self-explanatory. All the file-watching code, for example, now resides in one place (in `middleware`). Also, the entire codebase is now written in TypeScript (though there are lots of missing types still).

`sapper/core.js` exposes various APIs that can be used by the CLI and the middleware. (In principle these APIs could be used by other third party consumers, though that's not necessarily the intention, and they won't be documented any time soon.) Would like to figure out a way to import stuff internally from `sapper/core.js` to get all the TypeScript goodness without ending up duplicating stuff across bundles.

Templates are still a bit weird, because it makes sense to implement #91 (and #92, by extension) rather than refactoring something that's about to be destroyed.


runtime.js.map
```diff
@@ -0,0 +1 @@
...CAAC,CAAC;CACjB;AAED,IAAI,MAAe,CAAC;AAEpB,cAAqB,OAAa,EAAE,OAAgB;IACnD,MAAM,GAAG,OAAO,CAAC;IACjB,MAAM,GAAG,OAAO,CAAC;IAEjB,IAAI,CAAC,MAAM,EAAE;QACZ,MAAM,CAAC,gBAAgB,CAAC,OAAO,EAAE,YAAY,CAAC,CAAC;QAC/C,MAAM,CAAC,gBAAgB,CAAC,UAAU,EAAE,eAAe,CAAC,CAAC;;QAGrD,MAAM,CAAC,gBAAgB,CAAC,YAAY,EAAE,2BAA2B,CAAC,CAAC;QACnE,MAAM,CAAC,gBAAgB,CAAC,WAAW,EAAE,2BAA2B,CAAC,CAAC;QAElE,MAAM,GAAG,IAAI,CAAC;KACd;IAED,UAAU,CAAC;QACJ,IAAA,oBAAgC,EAA9B,cAAI,EAAE,cAAI,CAAqB;QAEvC,IAAM,WAAW,GAAG,IAAI,IAAI,QAAQ,CAAC,aAAa,CAAC,IAAI,CAAC,CAAC;QACzD,cAAc,CAAC,GAAG,CAAC,GAAG,WAAW;YAChC,EAAE,CAAC,EAAE,CAAC,EAAE,CAAC,EAAE,WAAW,CAAC,qBAAqB,EAAE,CAAC,GAAG,EAAE;YACpD,YAAY,EAAE,CAAC;QAEhB,OAAO,CAAC,YAAY,CAAC,EAAE,EAAE,EAAE,GAAG,EAAE,EAAE,EAAE,EAAE,IAAI,CAAC,CAAC;QAC5C,QAAQ,CAAC,IAAI,GAAG,CAAC,MAAM,CAAC,QAAQ,CAAC,IAAI,CAAC,EAAE,GAAG,CAAC,CAAC;KAC7C,CAAC,CAAC;CACH;AAED,cAAqB,IAAY,EAAE,IAA8B;IAA9B,qBAAA,EAAA,SAAS,YAAY,EAAE,KAAK,EAAE;IAChE,IAAI,QAAQ,CAAC,IAAI,GAAG,CAAC,IAAI,EAAE,MAAM,CAAC,QAAQ,CAAC,IAAI,CAAC,EAAE,IAAI,CAAC,EAAE;QACxD,IAAI,OAAO;YAAE,OAAO,CAAC,IAAI,CAAC,YAAY,GAAG,cAAc,GAAG,WAAW,CAAC,CAAC,EAAE,EAAE,EAAE,GAAG,EAAE,EAAE,EAAE,EAAE,IAAI,CAAC,CAAC;KAC9F;SAAM;QACN,MAAM,CAAC,QAAQ,CAAC,IAAI,GAAG,IAAI,CAAC;KAC5B;CACD;;;;"}
```
Conduitry: I imagine this should also be gitignored?

src/core/export.ts
```diff
@@ -60,7 +63,7 @@ module.exports = function() {
 		return fetch(url, opts);
 	};
 
-	app.use(sapper());
+	app.use(require('./middleware')()); // TODO this is filthy
```
lukeed: Looks like this missed a level, should be `../middleware`.

runtime.js.map
Rich-Harris: yep! fixed

src/core/export.ts
Rich-Harris: it's meant to be this way in fact — the built files are all in the project root (`/core.js`, `/middleware.js`), so while it looks incorrect from the perspective of the source code, it's correct from the perspective of the built code. Ideally it would `import` the middleware but it was finicky so I cheated, hence the TODO

lukeed: Got it 👍 On a more recent pass I realized that & forgot to remove comment~

PR #104, opened 2018-01-26
Fix DOMException when location.hash is invalid selector

thgh:
While building an app that uses location.hash extensively, the following error occurred often:
`Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#' is not a valid selector.`

querySelector throws on invalid selectors, but getElementById doesn't throw on invalid ids.


Rich-Harris:
thank you! wasn't aware of this, nice fix

PR #105, opened 2018-01-26
Fix route not found when url is /?

thgh:
The regex should always remove the question mark, even when there is nothing behind it.

Alternative code without regex:
`req.url.slice(0, (req.url + '?').indexOf('?'))`


Rich-Harris:
thanks!

PR #111, opened 2018-01-31
fixes api route default content-type

samhatchett:
fixes issue where api routes were being defaulted to text/html. Page routes should be text/html, but api routes could be json, zip files, etc., and express does some type-guessing to assist in case the user code does not specify the content-type.


samhatchett:
likely to conflict with PR #101

Rich-Harris:
thank you! remarkably, it merged perfectly. Not bad, git

PR #113, opened 2018-02-05
Consolidate Rollup config file

lukeed:
Personal preference, mostly. I feel it makes it easier to see what's going on at a glance, with key/unique information in the initial array.


Rich-Harris:
nice, thanks

Rich-Harris:
(will probably move that webpack file at some point, so it's not an anomaly that requires its own special handling. is only like that for historical reasons)

PR #114, opened 2018-02-05
Fix: Remove HMR attempts in Prod~!

lukeed:
In production apps, there were still attempts made to `/__webpack_hmr`, which meant that the dev-entry was being included.

The short-fix was to modify the build script: `NODE_ENV=production sapper build` ... but that's still ugly & not the intended.

I added some debuggers and saw that the `dev` block of `create_app` was still running. Then took a look at the compiled `cli.js` output:

screen shot 2018-02-05 at 2 09 04 pm

The `dev` check is made before manually setting the `NODE_ENV`. Also, because of Rollup's "import hoisting" (is that the term?) moving [this line](https://github.com/sveltejs/sapper/blob/master/src/cli/index.ts#L4) to the top of the file, ahead of the `dev` import, didn't affect the compiled output.

Simply converting `config.dev` to `config.isDev()` solves this, deferring the check until needed.

Now, `sapper build && sapper start` works as expected, while preserving HMR for dev-mode. 🎉

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -1,10 +1,11 @@
 import { build, export as exporter } from 'sapper/core.js';
-import { dest, dev, entry, src } from '../config';
+import { dest, entry, isDev, src } from '../config';
 
 process.env.NODE_ENV = 'production';
 
 const cmd = process.argv[2];
 const start = Date.now();
+const dev = isDev();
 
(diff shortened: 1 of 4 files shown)


Rich-Harris:
I wish I'd *thunk* of this

...

...

[tumbleweed]

PR #119, opened 2018-02-13
Better CLI

Rich-Harris:
For now, this basically just adds a `--help` message and proper argument parsing (with mri, naturally). Just laying the groundwork for future improvements, such as adding options to `sapper export` and `sapper build`, and adding a `sapper upgrade` command.

Heeding @lukeed's point about not aggressively rolling everything up (if you're building, you don't need `sapper export` dependencies like cheerio), it'd be nice to use dynamic imports and code-splitting to load commands lazily. Rollup will need some changes before that's possible though — specifically, we need more control over how chunks are named. Will try and remember to raise an issue (and/or PR).


lukeed:
Nice 👏 Not trying to just plug my shiz but if you're gonna do a full CLI, `sade` might make more sense. You can lazily require from within each action block.

Rich-Harris:
oh, nice! i'll file that under 'future work' though, as right now I'm just trying to get to a good place to start implementing stuff like #91 (which I need for the day job)

PR #124, opened 2018-02-15
Absolute path resolution

arxpoetica:
Just inlining tests with (potential) changes made over here: https://github.com/sveltejs/sapper-template/pull/36

### DIFF
--- a/test/app/routes/_components/Layout.html
+++ b/test/app/routes/_components/Layout.html
@@ -5,7 +5,7 @@
 </main>
 
 <script>
-	import Nav from './Nav.html';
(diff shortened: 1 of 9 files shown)


Rich-Harris:
Thanks for the PR — I'm going to close this and https://github.com/sveltejs/sapper-template/pull/36 though. Changing module resolution logic is a strong opinion that I don't think belongs in Sapper itself: it has some tangible downsides, such as making it impossible to use things like eslint-plugin-import or VSCode's intellisense.

But this is exactly why the project structure encourages you to tinker with the webpack config, so that we can easily support this sort of thing without baking it into the framework.

PR #151, opened 2018-03-03
try and narrow down test failure

Rich-Harris:


test/app/export/about/index.html
```diff
@@ -0,0 +1,62 @@
+<!doctype>
```
johnmuhl: Missing ` html`?

Rich-Harris: yep! good catch, thanks. unrelated to these test failures but a mistake nonetheless

Issue #116, opened 2018-02-06
Don't assume /client/** assets are JS


Rich-Harris:
Moving https://github.com/sveltejs/sapper/issues/109#issuecomment-362298418 into its own separate issue — [this line](https://github.com/sveltejs/sapper/blob/7a2ed16884ebfbe4497f2bc2b458ad7dc66f95d7/src/middleware/index.ts#L63) (and the related line in `connect_prod`) prevents certain webpack plugins from being used

EmilTholin:
Could this be solved by removing that line of code you linked, and [make `opts.type` optional](https://github.com/sveltejs/sapper/blob/7a2ed16884ebfbe4497f2bc2b458ad7dc66f95d7/src/middleware/index.ts#L140), or is that to naive?

Rich-Harris:
Related: Sapper will actually generate a script tag for all route assets, even non-JS ones...

https://gitter.im/sveltejs/svelte?at=5a99baf56fba1a703abf5043

jacobmischka:
Is this indeed fixed in #153? Should this be closed now?

PR #153, opened 2018-03-04
various stability improvements

Rich-Harris:
This started out as an attempt to improve the reliability of HMR when service workers are involved, but ended up being a bit more wide-ranging:

* Changes to the service worker are reflected without needing to restart the server
* Ditto for `app/template.html`. Additionally, editing that file will cause the page to reload
* hmr-client.js is now sapper-dev-client.js, reflecting the fact that it's not just for HMR any more
* Template handling is simpler
* Generated assets are cached in memory, rather than read from disk for each request

The HMR problem turned out to be that the service worker intercepted the EventSource connection incorrectly. It needs to be fixed in sapper-template, which will happen once this is released.


Rich-Harris:
This also fixes #116

PR #158, opened 2018-03-04
vastly better logging

Rich-Harris:
...or 'that feeling when @lukeed has already solved your problem'. This substantially improves the logging:

* warnings are actually shown (they were squelched before). Related: https://github.com/sveltejs/svelte-loader/issues/35
* errors and warnings are deduped (e.g. a component has a syntax error — we don't want to print the frame for both the server *and* the client compiler
* invalidations are printed to the console
* output is nicely colourised

There are lots of ways it could be improved further. It would be nice to separate server console logs from Sapper console logs, for example (some kind of split screen?). For people who want it, it would be nice to show the details of the webpack output. It would be very cool to integrate with something like [Jarvis](https://github.com/zouhir/jarvis) or [electron-webpack-dashboard](https://github.com/FormidableLabs/electron-webpack-dashboard), though that might be tricky when there are multiple webpack configs. (Maybe it makes sense to have a dedicated Sapper electron app?) For now though, this is a decent leap forward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "mkdirp": "^0.5.1",
     "mri": "^1.1.0",
     "node-fetch": "^1.7.3",
+    "pretty-ms": "^3.1.0",
     "relative": "^3.0.2",
     "require-relative": "^0.8.7",
     "rimraf": "^2.6.2",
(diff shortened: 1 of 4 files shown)


lukeed:
😅 happy to help lol. There's a plugin you can use directly if you want: `webpack-messages` -- I use that version personally.

Jarvis doesn't do multi compilers yet, but we can bang that out.

PR #161, opened 2018-03-04
add preloadRoutes function

### LINKED ISSUE #160: Programmatically preload secondary routes
At present, as soon as the first route is rendered, Sapper preloads the modules for all your secondary routes. In a lot of cases that makes sense, but not always.

Perhaps it should be a function that the user can choose whether or not to run:

```js
import { init, preloadRoutes } from 'sapper/runtime.js';
import { routes } from './manifest/client.js';

init(document.querySelector('#sapper'), routes).then(() => {
  preloadRoutes();
});
```

With no arguments, it would preload all routes. Or, you could provide an array of routes:

```js
// could replace '[slug]' with '*' or 'whatever', it'd still match
preloadRoutes(['/about', '/blog/[slug]']);
```

Not sure how you'd specify error pages in the second case. Maybe `'4xx'` and `'5xx'` get special treatment.

Rich-Harris:
closes #160

### DIFF
--- a/src/runtime/index.ts
+++ b/src/runtime/index.ts
@@ -59,9 +59,6 @@ function render(Component: ComponentConstructor, data: any, scroll: ScrollPositi
 			detach(start);
 			detach(end);
 		}
-
-		// preload additional routes
-		routes.reduce((promise: Promise<any>, route) => promise.then(route.load), Promise.resolve());
 	}
 
 	component = new Component({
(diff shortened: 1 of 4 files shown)


src/runtime/index.ts
```diff
@@ -268,3 +265,15 @@ export function goto(href: string, opts = { replaceState: false }) {
 		window.location.href = href;
 	}
 }
+
+export function preloadRoutes(pathnames: string[]) {
+	if (!routes) throw new Error(`You must call init() first`);
+
+	return routes
+		.filter(route => {
+			return !pathnames || pathnames.some(pathname => route.pattern.test(pathname));
```
EmilTholin: Will this take care of `4xx` and `5xx`, or is that to be added at a later stage?

Rich-Harris: Just me being forgetful 😬 thanks! fix incoming

Issue #169, opened 2018-03-05
select available port


Rich-Harris:
It would make sense if Sapper automatically selected an available port (starting at 3000, since that's what we're currently using) when doing `sapper dev` or `sapper start`.

I have a situation though where I have multiple Sapper apps that need to communicate with each other, so the ports need to be known ahead of time. For that reason it would also make sense if we could select a port with `sapper dev --port 4567` (or `-p 4567`), and Sapper would bail if it wasn't available.

thgh:
I'm running multiple sapper instances with `"dev": "PORT=4000 sapper dev"`. It works properly 👍  a

Can those occurences of 3000 be replaced by `process.env.PORT || 3000` ? Or is that not needed?

Rich-Harris:
ah yeah, there's a couple of places where things are hardcoded that shouldn't be. Good catch. And yeah, I guess we may as well treat `PORT=4000 sapper dev` as equivalent to `sapper dev -p 4000`. Reopening this

Rich-Harris:
Fixed by #184

PR #170, opened 2018-03-05
better CLI

### LINKED ISSUE #168: show stats after build/export
maybe along with instructions for running the site

Rich-Harris:
Fixes #168 and #169, using another @lukeed jam ([Sade](https://github.com/lukeed/sade)) which gives us a much nicer CLI.

`sapper dev`, `sapper start` and `sapper export` will all now pick available ports as necessary, unless (for `dev` and `start`) you specify a `--port` or `-p`, in which case it will try to use that port and print a useful error if it can't.

It also emits stats following the build, and prints instructions on how to start the app.


src/utils.ts
```diff
@@ -0,0 +1,10 @@
+import * as fs from 'fs';
+
+export function exists(file: string) {
```
lukeed: Can just use `fs.existsSync` here

src/cli/index.ts
```diff
@@ -1,89 +1,114 @@
...
+	.describe('Export your app as static files (if possible)')
+	.action((dest = 'export') => {
+		console.log(`> Building...`);
 
-	case 'export':
 		process.env.NODE_ENV = 'production';
+		process.env.SAPPER_DEST = '.sapper/.export';
 
-		const export_dir = opts._[1] || 'export';
+		const start = Date.now();
```
lukeed: You can use `process.hrtime` if you want to be a little more exact. Roughly the same though

src/cli/index.ts
```diff
@@ -1,89 +1,114 @@
...
 
-		break;
+prog.command('export [dest]')
+	.describe('Export your app as static files (if possible)')
+	.action((dest = 'export') => {
+		console.log(`> Building...`);
 
-	case 'export':
 		process.env.NODE_ENV = 'production';
+		process.env.SAPPER_DEST = '.sapper/.export';
```
lukeed: More for my own curiosity, but why not set `SAPPER_DEST` from within `export()`? Especially since the `dest` is variable

src/cli/index.ts
```diff
@@ -1,89 +1,114 @@
...
-});
+const prog = sade('sapper');
 
-if (opts.help) {
-	const rendered = help
-		.replace('<@version@>', pkg.version)
-		.replace(/^(.+)/gm, (m: string, $1: string) => /[#>]/.test(m) ? $1 : `  ${$1}`)
-		.replace(/^# (.+)/gm, (m: string, $1: string) => chalk.bold.underline($1))
-		.replace(/^> (.+)/gm, (m: string, $1: string) => chalk.cyan($1));
+prog.version(pkg.version);
```
lukeed: Definitely personal preference, but I like to recommend assigning all _global_ aspects into the initial assignment. New commands get their own blocks, and that consistent spacing helps separate things with larger CLIs.

```js
const prog = sade('sapper').version(pkg.version);`
```

src/utils.ts
Rich-Harris: huh, I thought it was deprecated. Turns out that's just `fs.exists`. Nice!

src/cli/index.ts
Rich-Harris: when `pretty-ms` is involved I don't think it'll make any difference. I generally avoid `process.hrtime` because, well, the signature is batshit

src/cli/index.ts
Rich-Harris: It's used by `build` first — it sends the webpack output to `.sapper/.export`, and *then* `sapper export` runs the code therein to generate the `export` directory. So it has to be set before either thing happens

src/cli/index.ts
Rich-Harris: Ah, nice. I didn't realise it was chainable, I suppose that makes sense!

lukeed: Aye. You can chain the whole thing if you want, but that'll get messy

Rich-Harris:
This PR ended up including a bunch of changes to the tests. They weren't very well structured before, and it turned out they had certain assumptions (about the build directory etc) that weren't necessarily true — in fact, the `dev` tests weren't testing dev behaviour at all.

Getting dev tests working will require jumping through a few more hoops, so I've just commented them out for now. Everything else is a bit neater and more robust.

PR #184, opened 2018-03-10
Fix hard-coded port

Rich-Harris:
This removes the hard-coded port in `sapper dev`, and makes these equivalent:

```bash
PORT=4567 npx sapper dev
npx sapper dev --port 4567
```


qwabra:
maybe you should add this to the documentation?

PR #187, opened 2018-03-10
Replace Chalk with Clorox

lukeed:
Returning the favor~ 😇 

By the way, was quickly looking at Polka inside export (🎉)  -- it doesn't look like it's doing anything? It's [defined](https://github.com/sveltejs/sapper/blob/master/src/cli/export.ts#L11) once but then never interacted with or passed anywhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,9 @@
     "test": "test"
   },
   "dependencies": {
-    "chalk": "^2.3.0",
     "cheerio": "^1.0.0-rc.2",
     "chokidar": "^1.7.0",
+    "clorox": "^1.0.3",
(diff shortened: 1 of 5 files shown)


Rich-Harris:
thank you!

Huh, you're right. An earlier implementation of `export` *did* use Express, but I guess that's no longer applicable. I'll move it to devDependencies (since it's used in tests)

lukeed:
Sounds good. I'll look at it closer later. Also, not a TS guy, but shouldn't `tslib` be a devDep too?

Rich-Harris:
yes, yes it should! fixed

PR #188, opened 2018-03-11
use devalue instead of serialize-javascript

### LINKED ISSUE #112: [question] use of modules/classes in preload
I think this issue has to do with the serialization/deserialization needed for the preload mechanism. 

What I'm trying to do is use imported classes in my svelte component. I'd like to get some data and construct my objects in `preload` - this seems to work when preload is called in the browser, but of course fails when the server preloads, and sends the data to the browser. The browser just gets the json representation devoid of functions.

Is there a good place to intercept the server-preloaded content to ensure that I can `Object.assign` that data back into my real objects? Or is it enough to trust the server-rendered content for the first load, and just try/catch to gracefully detect when the objects are not what I expect?

Hope this makes sense - just in case i have a demonstration here, based on the template project.
https://github.com/samhatchett/sapper-module-preload

Rich-Harris:
This fixes #112 by handling non-POJOs correctly. [devalue](https://github.com/Rich-Harris/devalue) handles objects like `Set` and `Map` that cause serialize-javascript to fail badly. If it encounters a custom object...

```html
<script>
  export default {
    preload() {
      class Foo {}
      return {
        foo: new Foo()
      };
    }
  };
</script>
```

...it will bail altogether, leaving the client to run `preload` again, rather than mistakenly serializing it as `{}`.

One thing serialize-javascript does that devalue (by design) *doesn't* is stringify functions. So this will need to be in a 0.9 release, since it could theoretically be a breaking change for some people.


EmilTholin:
Awesome stuff! Haven't really dug into the `devalue` code yet, but will it sanitize the data like `serialize-javascript` does (which I think was the main reason it was added in the first, if I recall correctly)?

Rich-Harris:
yep! I actually totally forgot that we were using serialize-javascript, and thought we were just using `JSON.stringify`... but yeah, devalue 'borrows' the XSS mitigation code from serialize-javascript, which basically consists of escaping <, > and /

PR #190, opened 2018-03-11
allow server routes to be .ts files (or anything else)

### LINKED ISSUE #57: Support TypeScript for non-component source code
I guess we just need to recognise file extensions other than .js, and update the webpack config in sapper-template

Rich-Harris:
fixes #57. It still leaves a lot of stuff for the user to do, in terms of configuring webpack, but that shouldn't be a barrier for the sorts of people who want to use TypeScript in their apps.

### DIFF
--- a/src/core/create_routes.ts
+++ b/src/core/create_routes.ts
@@ -3,7 +3,7 @@ import glob from 'glob';
 import { src } from '../config';
 import { Route } from '../interfaces';
 
-export default function create_routes({ files } = { files: glob.sync('**/*.+(html|js|mjs)', { cwd: src() }) }) {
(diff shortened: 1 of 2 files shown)


src/core/create_routes.ts
```diff
@@ -3,7 +3,7 @@ import glob from 'glob';
 import { src } from '../config';
 import { Route } from '../interfaces';
 
-export default function create_routes({ files } = { files: glob.sync('**/*.+(html|js|mjs)', { cwd: src() }) }) {
+export default function create_routes({ files } = { files: glob.sync('**/*.*', { cwd: src() }) }) {
```
EmilTholin: I might be mistaken, but will not `'**/*.*'` include the folders as well? That might be desired?

Issue #181, opened 2018-03-09
Sapper makes assumptions about where to find the 'app' folder.


naturalethic:
For example here:

https://github.com/sveltejs/sapper/blob/master/src/middleware/index.ts#L117

Sapper is assuming the current working directory is similar to development, however in some situations the cwd may be different.  For example in my deployment to google cloud functions, I have a deeper directory structure, placing my sapper app in a deep folder from the path where the function is run.

Something like a SAPPER_APP_DIR, SAPPER_BASE_DIR or something similar should be configured to allow for running the app without relying on cwd.

I experimented with using SAPPER_DEST, and assuming that `app` would be one level above this, however that is not the case for `export` and breaks that test.

Rich-Harris:
As of 0.94 you can configure `SAPPER_BASE`, `SAPPER_APP` and `SAPPER_ROUTES` — hopefully that meets your needs, let me know if not

naturalethic:
Perfect, thanks so much!

Issue #172, opened 2018-03-06
Static site optimisations


Rich-Harris:
When using `sapper export`, there are some interesting opportunities for optimisations that don't really make sense when pages are being generated on-demand. For example, I just learned about [subfont](https://github.com/Munter/subfont) by @Munter, which can generate an efficient subset of the fonts a given page uses.

A few ideas off the top of my head:

* Subfont
* Minify HTML
* Further optimise CSS
* Compress SVG assets
* Compress other images
* Probably some other Assetgraph-related optimisations

Munter:
Looks like https://github.com/sveltejs/sapper/blob/master/src/cli/export.ts#L64-L93 can be replaced by assetgraph crawling the pages and https://github.com/sveltejs/sapper/blob/master/src/cli/export.ts#L47-L62 can be removed, since an assetgraph-based optimisation would require to contain the entire graph in memory at the same time.

So in essensce assetgraph would just scrape the page in its entirety from the temporary server, then run a bunch of optimisations on the entire set before writing it to disc.

Ping @papandreou

Rich-Harris:
Now we're talking!

> Of course this all hinges in sapper outputting standard compliant enough code for assetgraph to pick up in its static analysis

What are some of the gotchas to be aware of?

Rich-Harris:
(Also: are there plans to integrate Subfont into Assetgraph? Or would that be a separate step?)

Munter:
> What are some of the gotchas to be aware of?

In most cases it's things like unbalanced html, syntax errors in css that browsers compensate for, missing files etc. All stuff that you'd want to be aware of.

We haven't built in support for the latest js features like ES modules and such. But since this is a compile output I assume your existing pipeline has already build browser compatible output. Things like custom runtime module loaders might not be picked up

Rich-Harris:
It's currently using webpack, with code-splitting — we plan to also support Rollup soon, which can output ESM (for browsers with native module support) or System.js modules.

> In most cases it's things like unbalanced html, syntax errors in css that browsers compensate for, missing files etc. All stuff that you'd want to be aware of.

So it bails if it encounters something invalid? I think that's fine (Svelte won't compile if your CSS is bad anyway, so the only way that would happen is if there was a broken `assets/styles.css` file or something — as long as it doesn't just fail silently!

Munter:
[...]

PR #193, opened 2018-03-11
minify HTML templates

### LINKED ISSUE #15: Minify HTML templates
Probably isn't worth doing this on each page render, but the initial templates could be minified to save a few bytes

Rich-Harris:
fixes #15. This is separate to #172 — it only minifies `app/template.html` itself (and only in prod), not the end result with rendered components. Still, it shaves off a few bytes which is nice.

I suppose we could improve on this by doing the minification at build time and writing the template to the build directory — something to consider at the same time as #181.


Rich-Harris:
I figured it was worth addressing #181 at the same time. This PR adds the following environment variables:

* `SAPPER_BASE` — defaults to cwd
* `SAPPER_APP` — defaults to `${SAPPER_BASE}/app`
* `SAPPER_ROUTES` — defaults to `${SAPPER_BASE}/routes`

`SAPPER_DEST` can still be overridden during dev, if necessary (for `build` and `export` it's part of the command itself).

Also, minification now happens at build time, which means a faster startup.

PR #204, opened 2018-03-18
Add support for sourcemap *.map files

nolanlawson:
For some apps I'd like to enable source maps in production. However this causes the following output for `%sapper.scripts%`:

```html
<script src='/client/<hash>/main.js'></script>
<script src='/client/<hash>/main.js.map'></script>
```

This PR fixes it so that Sapper outputs:

```html
<script src='/client/<hash>/main.js'></script>
```

### DIFF
--- a/src/cli/export.ts
+++ b/src/cli/export.ts
@@ -25,6 +25,10 @@ export async function exporter(export_dir: string, { basepath = '' }) {
 		sander.copyFileSync(build_dir, 'service-worker.js').to(export_dir, 'service-worker.js');
 	}
 
+	if (sander.existsSync(build_dir, 'service-worker.js.map')) {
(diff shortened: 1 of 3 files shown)


nolanlawson:
I realized my previous PR was incomplete. This new commit adds full support for `.map` files, for both `sapper start` and `sapper export`, for both the client and the service worker.

So now a user can, for instance, put the following configuration into `client.config.js` or `service-worker.config.js`:

```js
devtool: isDev ? 'inline-source-map' : 'source-map'
```

...and all the `*.map` files will be correctly served and exported.

nolanlawson:
Last change: there's no sense in adding `<link rel=preload>` for `.map` files.

Rich-Harris:
sweet, thanks! cutting a release shortly

PR #205, opened 2018-03-19
prefetch on mouse stop

Rich-Harris:
This allows `rel=prefetch` to be used more liberally on anchor tags — rather than prefetching on mouseover, it waits until the mouse stops over an element, which prevents the false positives that resulted from the mouse 'brushing' over a list of links.

Touch behaviour is unchanged.


Rich-Harris:
Can't seem to get the tests to pass — it's not registering the synthetic mousemove event

Rich-Harris:
No idea how to make this work in Nightmare, so I'll do the next best thing — `skip` the test

PR #215, opened 2018-03-25
Stable sort

Rich-Harris:
Found a situation in which route sorting could be unstable, resulting in incorrect navigation. This fixes it. Also includes an unrelated commit I'd forgotten about, which binds `store` to error pages

### DIFF
--- a/src/core/create_routes.ts
+++ b/src/core/create_routes.ts
@@ -102,7 +102,10 @@ export default function create_routes({ files } = { files: glob.sync('**/*.*', {
 					}
 
 					if (!a_sub_part.dynamic && a_sub_part.content !== b_sub_part.content) {
-						return b_sub_part.content.length - a_sub_part.content.length;
+						return (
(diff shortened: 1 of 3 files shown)


src/core/create_routes.ts
```diff
@@ -102,7 +102,10 @@ export default function create_routes({ files } = { files: glob.sync('**/*.*', {
 					}
 
 					if (!a_sub_part.dynamic && a_sub_part.content !== b_sub_part.content) {
-						return b_sub_part.content.length - a_sub_part.content.length;
+						return (
+							(b_sub_part.content.length - a_sub_part.content.length) ||
```
thgh: Why not just alphabetically?

Rich-Harris: Because `a/b/foo.html` needs to match before `a/b/f[xx].html`

thgh: OK I see now 😄  thank you!

PR #226, opened 2018-04-02
Downgrade chokidar to 1.7.0

naturalethic:


package.json
```diff
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "cheerio": "^1.0.0-rc.2",
-    "chokidar": "^2.0.2",
+    "chokidar": "^1.7.0",
```
TehShrike: Prolly otta remove the `^`?

Conduitry: Eh, I'd say that's unlikely to matter. There aren't any versions of chokidar between 1.7 and 2, and `^` won't pick 2+.

Rich-Harris:
Thanks, released 0.10.3

PR #234, opened 2018-04-13
Do not encode characters allowed in path when generating routes

akihikodaki:
In RFC 3986, some characters not allowed in query, which encodeURIComponent
is designed for, is allowed in path.
A notable example is "@", which is commonly included in paths of social
profile pages. Such characters should not be encoded.

The new encoding function is conforming to the RFC.


Rich-Harris:
Thank you, this is so helpful! I had no idea about the distinction.

PR #236, opened 2018-04-15
Allow to have middleware for the path same with a HTML page

akihikodaki:
HTTP allows to change the type of the content to serve by Accept field in
the request. The middleware for the path same with a HTML page will
be inserted before the HTML renderer, and can take advantage of this
feature, using expressjs's "accepts" method, for example.


Rich-Harris:
Thanks for this — I think I understand it, and it seems sensible, but just to help me make sure I've got this down, could you walk me through a scenario where you'd use this, and what the middleware would look like?

akihikodaki:
Here is a simple example: https://github.com/miniorg/miniverse/blob/6e8a82c8036d62df4f3d5b2303cd907e3b7c05fa/routes/%40%5Bacct%5D/index.js

If the user agent prefers `application/activity+json` or `application/ld+json`, it returns a representation of a person in format called JSON ActivityStreams. If the user agent prefers `text/html` or shows no preference, an HTML representation will be served.

Rich-Harris:
sounds good — thanks for the clarification

PR #243, opened 2018-04-19
Accept directory entries which starts with dot as routes

akihikodaki:
It allows to implement `.well-known` URIs.


Rich-Harris:
nice, thank you 👍

silentworks:
What is a `.well-known` URI? this change seem to have created an new issue where files like `.DS_Store` is not getting included as a route in the manifest files.

PR #245, opened 2018-04-23
Include process.env in exporter server options

johnmuhl:
I ran into a problem deploying with up where having access to an environment variable for flow control would make solving the problem easy. I think it would useful in other environments or for other reasons also; and not needing to `if (!process.env.MISSING) ...` to ensure the export won't fail when things still run fine in development.


Rich-Harris:
seems very reasonable, thank you!

PR #250, opened 2018-05-04
implement --launcher

Rich-Harris:
This adds a `--launcher` flag that creates a file, defaulting to `build/index.js`, that configures the necessary environment variables before running `build/server.js`. With this change, a Sapper app doesn't even need to have Sapper as a dependency; it can be in `devDependencies` instead (it will get bundled by webpack). This is good news for apps running on Lambda or whatever.

Running the app is as simple as `node build`.

It will default to setting `process.env.PORT` to 3000, but that can be changed with `--port`. Additionally, it will respect existing environment variables, so you can also do `PORT=4567 node build`.


Rich-Harris:
Wait. I just realised we can make this simpler. Instead of having the option to create a custom launcher file, we just *always* create an `index.js` launcher in the `dest` directory. Simples.

PR #253, opened 2018-05-04
Update clients manifests file to have id and type on each route

silentworks:
Remove unnecessary if statement

### DIFF
--- a/src/core/create_manifests.ts
+++ b/src/core/create_manifests.ts
@@ -59,7 +59,7 @@ function generate_client(routes: Route[], path_to_routes: string, dev_port?: num
 						? '{}'
 						: `{ ${route.params.map((part, i) => `${part}: match[${i + 1}]`).join(', ')} }`;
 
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Thanks — remind me why the ID is necessary?

silentworks:
For later when doing stuff with Route transitions, the more information I have as to which route I am going from and which I am going to, the better.

Rich-Harris:
Ah, I see. The thing about `id` is it's a little bit cryptic — it's basically for internal usage, and is often something garbagey like `'blog_$slug$_json'`. In #259 I've added the `path` property, which is possibly a bit more self-explanatory — it allows you to do [this sort of thing](https://github.com/sveltejs/sapper-template/blob/98813dd8fde5bf6f1351aca2efff70a035ceccc3/app/App.html#L1-L6):

```html
<Nav page={
	props.path === '/' ? 'home' :
	props.path === '/about' ? 'about' :
	props.path.startsWith('/blog') ? 'blog' :
	null
}/>
```

Maybe we don't need `id`? And if it eventually turned out we did need something like it, perhaps that purpose would be better served by e.g. `route.matches('/some-path')` or whatever. Thoughts?

silentworks:
That makes sense.

PR #254, opened 2018-05-04
Ignore files and directories with leading dots except .well-known

### LINKED ISSUE #252: Routes are not ignoring .DS_Store files and build breaks
Sneer: why, Apple. Why.

This change https://github.com/sveltejs/sapper/commit/93128a01561b25d57a0505f783a39e925a540beb introduced the problem. The build breaks when those sometimes autogenerated Apple files appear. I'm wondering what other `.` files will cause problems?

akihikodaki:
Resolves #252.


Rich-Harris:
thanks!

PR #256, opened 2018-05-04
replace discontinued @std/esm package with esm

johnmuhl:


Rich-Harris:
actually... I'm not sure we're using it at all, now that I think of it. Can probably just remove altogether — will do that. Thanks

Issue #157, opened 2018-03-04
Suggestion: a main app component


mrkishi:
So, I've been using pure Svelte and Roadtrip in order to build an app with an architecture similar to sapper. I wanted to use sapper but it was just too early... Thankfully this allowed me to test different things that _might_ benefit sapper.

One of them is having a main app component. The simplest (and implicitly used if one's not provided) main app component would be something like this:

```html
<:Component {routeComponent} :routeData/>

// when Svelte gets spread parameters

<:Component {routeComponent} {{...routeData}}/>
```

The framework would then define an API for the main app component in order to expose the routed components (although the minimal API would be simply "there's a `routeComponent` component and a `routeData` object").

This would allow us to have persistent components throughout the whole app lifecycle. It would also be a good place to expand sapper's routing API (which is something I think is currently lacking -- say, getting notified whenever routes change), with full Svelte access.

This would probably only be ergonomic once spread parameters are in, but hopefully it's worth tinkering with!

Thoughts?

Rich-Harris:
It might be a little tricky because of `preload`. Currently, each top-level component (aka page) gets to load asynchronous dependencies before it's rendered (either on server or client) — non-top-level components *don't* have that ability (similar to Next's `getInitialProps`).

I've wondered about having some way to preserve components between route transitions:

```html
<Nav key:nav/> <!-- because of key:nav, the component is 'recycled', never destroyed -->
```

Having a main app component would solve that problem in a slightly more elegant way, but I think we'd need to solve the async loading problem before we could really do that. (Perhaps we need to steal some of [these ideas](https://reactjs.org/blog/2018/03/01/sneak-peek-beyond-react-16.html) from React.)

I agree that there should be a way to listen for route changes.

mrkishi:
In my app, I also preload route data.

What I do is `app.set({ route: Component, data: preloaded })`, but I could see a fancier API for cases where the main app component would intercept the routing logic.

I guess what I proposed here is not changing the way `routes` work, but rather `init` -- making some of sapper's framework logic more... Svelty, for lack of better word.

(ps. thanks for the link! I'll check it out)

Rich-Harris:
Ah, I think I understand what you mean now. Yes, that could potentially work. Will think on that. Definitely needs spread in order to be ergonomic.

rob-balfre:
[...]

PR #259, opened 2018-05-05
switch to single App component model

Rich-Harris:
See #157.

TODO:

* [x] error if no `App` is provided to the middleware or the client runtime
* [x] add `preloading` property
* [ ] ~~add `selected: { params, query, path }` property~~ on second thoughts, maybe add this later if we need it
* [x] update docs


Rich-Harris:
Ok, think this is probably ready. Associated PRs:

* https://github.com/sveltejs/sapper.svelte.technology/pull/15
* https://github.com/sveltejs/sapper-template/pull/57

Issue #267, opened 2018-05-15
Use ansi-colors instead of clorox and and parseurl instead of url-parse


Rich-Harris:
A few dependencies that can be replaced for smaller/faster/etc alternatives:

* clorox --> ansi-colors
* url-parse --> parseurl
* glob --> tiny-glob

cc @lukeed

lukeed:
Want me to do this? I have a few minutes now

lukeed:
Can be closed~

PR #268, opened 2018-05-15
Update colors dep

lukeed:
Step 1 of 3 on #267. Didn't have time for the others, sorry. 

There's room to update Sapper to use the `symbols` provided by `ansi-colors`, but I didn't touch that. Symbols are kind of a personal touch 😆 

Also, might as well keep using `glob` since webpack relies on it. Would be adding a new player in this case, rather than replacing one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,9 +18,9 @@
     "test": "test"
   },
   "dependencies": {
+    "ansi-colors": "^2.0.1",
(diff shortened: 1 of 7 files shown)


lukeed:
Updated to `webpack-format-messages@2.x` because it's also using `ansi-colors` now.

Rich-Harris:
thank you!

PR #283, opened 2018-06-12
Feature: use regexp in routes

elcobvg:
Needed this feature for current project. Using regexp to distinguish routes that would otherwise be clashing. This makes it possible to have routes like so:

```
index.html
[page].html
[lang([a-z]{2})]/index.html
[lang([a-z]{2})]/[page].html
[category([a-z-]{3,})]/[page].html
```
Obvious caveat _(for now at least)_: it's not possible to use `/` or `\` in the regexp patterns as that would break the route path.

### DIFF
--- a/src/core/create_routes.ts
+++ b/src/core/create_routes.ts
@@ -4,171 +4,171 @@ import { locations } from '../config';
 import { Route } from '../interfaces';
 
 export default function create_routes({ files } = { files: glob.sync('**/*.*', { cwd: locations.routes(), dot: true, nodir: true }) }) {
-	const routes: Route[] = files
-		.filter((file: string) => !/(^|\/|\\)_/.test(file))
-		.map((file: string) => {
-			if (/(^|\/|\\)(_|\.(?!well-known))/.test(file)) return;
+    const routes: Route[] = files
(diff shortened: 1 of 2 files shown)


elcobvg:
Tests fail, clearly I've overlooked something. Back to the drawing board...

Rich-Harris:
Thanks! This is a neat idea. (Apologies for not commenting sooner, I didn't realise this was no longer WIP.)

Could you fix the whitespace so that it matches the original please (the codebase uses tabs, not spaces)? It's showing changes in lines that haven't actually changed. Also, this'll need some tests, in [here](https://github.com/sveltejs/sapper/blob/master/test/unit/create_routes.test.js) 😀

elcobvg:
Fixed the whitespace and added some unit tests. 
An additional issue: one _could_ use capturing groups in the regexp pattern, which would result in 'nested' params. But webpack chokes on it: throws an error that it can't find the module and I can't figure out why (my webpack knowledge is limited...) Since I have a hard time thinking of a real world use case for this anyway, I've taken the liberty to disallow `(` and `)` in the regexp pattern for now

Rich-Harris:
awesome, thank you!

Issue #262, opened 2018-05-08
Add support for nested routes


tonyfsullivan:
This one came up in Gitter today and seems worth investigating.  It could be really helpful for Sapper to support nested routing, potentially with a special `slot`, `<sapper:child />` or similar.

Sapper currently handles routes by walking the routes directory and finding the matching path (`/settings/profile.html` or `/settings/profile/index.html` for example).  In cases where you want a shared Settings layout that changes content, you currently need to create something like a `SettingsLayout.html` component and use that to wrap content in `/settings/profile.html`.  i.e. The child route needs to know all parent components and include them in it's layout.

This gets a little tricky with multiple levels of nested routes.  Ember, Angular, and React Router have slightly different solutions for nesting routes - worth investigating how they solved it and what makes the most sense for Svelte/Sapper

Rich-Harris:
Been reading the linked docs and, well, hoo boy. I can't believe it's nearly as complicated as Vue, Ember and Angular make it out to be, with special elements and extra configuration all over the place.

React Router has by far the most sensible solution to this problem, though the design of RR is fundamentally at odds with that of Sapper.

I think the solution is two-fold:

1. Add some way of naming a page that says 'match multiple path parts'
2. Use `#if` blocks in the template

By which I mean that we could have a page like `routes/settings/[+]/index.html`, it would match `/settings`, `/settings/notifications`, `/settings/notifications/emails` and so on.

That file could look like this:

```html
<h1>Settings</h1>

{#if menu[0] === 'notifications'}
  <Notifications {menu}/>
{:else}
  <!-- ... -->
{/if}

<script>
  export default {
    components: {
      Notifications: './_components/Notifications.html'
    },

    preload({ path }) {
      return {
        menu: path.split('/')
      };
    }
  };
</script>
```

(Note that if we change the default behaviour outlined in #295, whereby the page component persists between navigations if the constructor is unchanged, then we would need to implement something like the `persist` option discussed therein.)

Am I being naive? Or would that cover the cases we care about?

silentworks:
Your solution sounds like a lot of manual work, where-as Ember and Vue is not so manual.

Rich-Harris:
Huh. Genuinely surprised at that reaction — in Vue we have to add this kind of boilerplate...

```js
const router = new VueRouter({
  routes: [
    { path: '/user/:id', component: User,
      children: [
        {
          // UserProfile will be rendered inside User's <router-view>
          // when /user/:id/profile is matched
[...]

PR #308, opened 2018-07-16
[WIP] Nested routes

Rich-Harris:
This is a big PR with a lot of moving parts. So far all it does is change how routes are created — instead of constructing route objects from an array of filenames (generated from a globbing pattern), it walks over a directory building up the intermediate components it needs. As well as enabling #262, this is arguably much simpler and more robust.

Next task is to change the manifest generation from something like this...

```js
// This file is generated by Sapper — do not edit it!
export const routes = {
  ignore: [/^\/blog.json\/?$/, /^\/blog\/([^\/]+?).json\/?$/],

  pages: [
    { pattern: /^\/?$/, params: () => ({}), load: () => import(/* webpackChunkName: "_" */ '../../routes/index.html') },
    { pattern: /^\/about\/?$/, params: () => ({}), load: () => import(/* webpackChunkName: "about" */ '../../routes/about.html') },
    { pattern: /^\/blog\/?$/, params: () => ({}), load: () => import(/* webpackChunkName: "blog" */ '../../routes/blog/index.html') },
    { pattern: /^\/blog(?:\/([^\/]+?))?\/?$/, params: match => ({ slug: match[1] }), load: () => import(/* webpackChunkName: "blog_$slug$" */ '../../routes/blog/[slug].html') }
  ],

  error: () => import(/* webpackChunkName: '_error' */ '../../routes/_error.html')
};

if (module.hot) {
  import('/Users/208311/Development/SVELTE/sapper/sapper-dev-client.js').then(client => {
    client.connect(10000);
  });
}
```

...to something like this:

```js
// This file is generated by Sapper — do not edit it!
import root from '../../routes/index.html';

const page_index      = () => import('../../routes/_default.html');
const page_about      = () => import('../../routes/about.html');
const page_blog       = () => import('../../routes/blog.html');
const page_blog_index = () => import('../../routes/blog/_default.html');
const page_blog_$slug = () => import('../../routes/blog/[slug].html');

export const routes = {
  ignore: [/^\/blog.json\/?$/, /^\/blog\/([^\/]+?).json\/?$/],

  root,

  pages: [
    {
      pattern: /^\/?$/,
      parts: [
        { component: page_index }
      ]
    },

    {
      pattern: /^\/about\/?$/,
      parts: [
        { component: page_about }
      ]
    },

    {
      pattern: /^\/blog\/?$/,
      parts: [
        { component: page_blog },
        { component: page_blog_index }
      ]
    },

    {
      pattern: /^\/blog\/([^\/]+?)\/?$/,
      parts: [
        { component: page_blog },
        { component: page_blog_$slug, params: match => ({ slug: match[1] }) }
      ]
    }
  ],

  error: () => import(/* webpackChunkName: '_error' */ '../../routes/_error.html')
};

if (module.hot) {
  import('../../../../sapper/sapper-dev-client').then(client => {
    client.connect(10000);
  });
}
```

After that, the middleware needs updating to make use of the new server-side manifest, then the runtime needs updating for the client-side manifest.

There are still some details that are subject to bikeshedding, e.g. [what to call the property](https://github.com/sveltejs/sapper/issues/262#issuecomment-401887775) that contains child information for each branch. (I'm leaning towards `child` — it'd only be a problem if someone explicitly added a `child={...}` property on the `<svelte:component>` that introduced a child component, and the docs would tell you not to do that.)


Rich-Harris:
All tests now passing (locally at least — started getting failures on AppVeyor that appear to be connected to https://github.com/segmentio/nightmare/issues/1476), which makes me very happy. One major piece of missing functionality — if you have a route like

```
routes/[x]/[y]/[z].html
```

and you go from `/foo/bar/baz` to `/foo/bar/qux`, it re-preloads `x` and `y` even though they're unchanged.

Instead, it should keep `x` and `y` intact, preload `z` with the new parameter, set the `z` component to `null` then set it back to what it was previously.

Aside from that, there's probably a bit of tidying up to do (e.g. we can improve the manifest slightly, I think) and then it's just a question of documentation.

Rich-Harris:
Ok, think this is just about ready — it now preserves each unchanged 'level' of component. There's still some tidying up to do but this PR needn't wait on it.

One gotcha that I hadn't anticipated — if you preload data and then update those same keys after the component is rendered, they will get clobbered on route change:

```html
<!-- this will be 43 once oncreate runs, but if the
     user navigates to a new route that preserves
     this component, it will revert to 42 -->
{foo}

<script>
  export default {
    preload() {
      return { foo: 42 };
    },
    oncreate() {
      this.set({ foo: 43 });
    }
  };
</script>
```

This is potentially a bit annoying but is understandable, and easily worked around. I don't think it's a dealbreaker.

It's so nice being able to do `preload` at each level of the route hierarchy!

Rich-Harris:
We talked about this in Discord a bit. Nothing is set in stone, but the reason we landed where we did is that the default page is likely to include nested components, and it makes more sense to have a `_default` directory with an `index.html` and various `NestedComponent.html` files than it would to group them inside a directory named `index`.

There *is* something appealing about calling index pages `index.html` though! It'd probably be `_layout.html` rather than `Layout.html`, for consistency with other invisible-to-the-router files, especially with `_error.html`.

On the other hand it feels odd to have `preload` logic in a 'layout' file.

@tehshrike You convinced me before that index content should go in `_default.html` or `_default/index.html`, can you refresh my memory? You've been thinking about this problem for longer than most.

Rich-Harris:
Another change: it no longer makes sense to have this...

```js
import { routes } from './manifest/client.js';
```

...since that object contains more than just an array of routes. Instead it is now this (with the equivalent change on the server):

```js
import { manifest } from './manifest/client.js';
```

Rich-Harris:
@ansarizafar Thanks for that last-minute save. I've spent the day thinking about it and I'm very confident that we've made the right decision. Layout components are optional; if omitted, Sapper will use a default layout component:

```html
<svelte:component this={child.component} {...child.props}/>
```

PR #309, opened 2018-07-21
Don't create fallback routes

Rich-Harris:
Am in two minds about this one. In #308, if you have something like `routes/foo/bar/baz.html` but *don't* have (say) `routes/foo/index.html`, Sapper will import a fallback component that looks like this:

```html
<svelte:component this={child.component} {...child.props}/>
```

Since `routes/foo/index.html` would probably look like that *anyway*, we can a) save the developer some typing, b) avoid cluttering `routes`, and c) prevent the situation where you have multiple components that look like that, duplicated unnecessarily, which is good for file size. You'd then only need to create an intermediate `index.html` file if you needed specific markup or behaviour.

However, it's less explicit, and it's possible to create a webpack config that breaks it, by only compiling components inside `routes`.

So for now I'm just going to leave it here as a PR to the original PR.

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -105,8 +105,15 @@ class Watcher extends EventEmitter {
 
 		const dev_port = await ports.find(10000);
 
-		const routes = create_routes();
-		create_main_manifests({ routes, dev_port });
+		try {
+			const routes = create_routes();
+			create_main_manifests({ routes, dev_port });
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Came up with a solution I like much better — write the default layout component to the `manifest` directory. Explicit, and more or less guaranteed to be compiled correctly.

PR #326, opened 2018-07-29
Add `ignore` option

lukeed:
When Sapper runs as a top-level middleware, it (strongly) assumes it's the final handler for any/all requests on that domain. While this is 100% desired in most cases (and especially when relegated to a `basepath`), it does make it near-impossible to prevent this behavior in the 1-2% of cases where an escape hatch is needed.

More specifically, Polka users are forced to run _everything_ as a root-level middleware, even if a sub-application is meant to control its own basepath. This is currently the only work around in this case:

```js
polka()
  .use(
    compression(),
    // ...
    (req, res, next) => {
      if (req.path.startsWith('/users') {
        req.path = req.path.substring(6); // remove "/users"
        return users_sub_application(req, res, next);
      }
      next();
    },
    // ...
    sapper()
  )
```

With this feature, Polka (or any) users can bypass Sapper's routing and allow their legacy/sub/third-party/whatever apps to run on the same server, without worrying about middleware sequencing.

The `options.ignore` can be an instance of `RegExp`, `Function`, `String`, or an `Array` of any combination.

```js
sapper({
  store,
  manifest,
  ignore: [
    /foobar/i,
   'foobar', // leading slash will be added
    '/foobar', // identical
    uri => uri.startsWith('/foobar')
  ]
});
```

In the middleware composition, a new middleware is added only if/when the `options.ignore` is defined. The `req.__SAPPER__IGNORE__` boolean will be set & is used in the following middlewares to exit early, avoiding any sapper-related work that isn't desired for the current, incoming request.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -73,9 +73,18 @@ interface Component {
 	preload: (data: any) => any | Promise<any>
 }
 
+const IGNORE = '__SAPPER__IGNORE__';
+function toIgnore(uri: string, val: any) {
+	if (Array.isArray(val)) return val.some(x => toIgnore(uri, x));
+	if (val instanceof RegExp) return val.test(uri);
+	if (typeof val === 'function') return val(uri);
+	return uri.startsWith(val.charCodeAt(0) === 47 ? val : `/${val}`);
+}
+
 export default function middleware(opts: {
 	manifest: Manifest,
 	store: (req: Req) => Store,
+	ignore?: any,
 	routes?: any // legacy
 }) {
 	if (opts.routes) {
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Thanks — I still strongly feel that Polka should adopt Express's behaviour here (partly because it's just much easier to understand, but also because it would make it simpler to migrate existing Express apps), but in lieu of that...!

Before I merge, any idea what's going on with the test failures?

Rich-Harris:
super weird — when I checkout this branch locally, I get the same errors as Travis and Appveyor. Can't figure out why it'd be related to these changes (though i haven't looked deeply yet). Will have to investigate further when i'm not at work

lukeed:
I'll try to do the same tonight or tomorrow night, including fresh installs. 

Sorry for the trouble

lukeed:
Ah, found it 😆 The failing test navigates to `/foo/bar/baz` which I've marked as a route to ignore. My "custom" routes are just sending down plaintext, so it's competing with one another.

Fix incoming~

Rich-Harris:
thanks! released 0.15.4. Opened an issue on the docs repo — not exactly sure where to put it https://github.com/sveltejs/sapper.svelte.technology/issues/24

PR #330, opened 2018-08-01
[WIP] detect unused data on initial render

Rich-Harris:
This is somewhat experimental. The idea is that if we can wrap preloaded data in proxies, we can detect which values are actually used in the initial render (including in computed properties etc), and print a warning if any data is *not* used, ideally prompting the developer to remove those properties from `preload` return values.

It also logs a message if the preloaded data is too large, regardless of whether that's because of unused properties.

Todo:

* [ ] add tests
* [ ] maybe support `Map` objects as well?
* [x] emit an event, don't just print to the console
* [ ] figure out how much preloaded data is 'too much'
* [ ] decide whether this is really a good idea

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -224,11 +224,8 @@ class Watcher extends EventEmitter {
 					});
 
 					this.proc.on('message', message => {
-						if (message.__sapper__ && message.event === 'basepath') {
-							this.emit('basepath', {
-								basepath: message.basepath
-							});
-						}
+						if (!message.__sapper__) return;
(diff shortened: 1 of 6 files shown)


Rich-Harris:
I guess that at the very least, we only want to do this in dev mode — I don't know what the performance implications of doing this would be, but even if they're small, it's overhead that we shouldn't have in prod mode.

Rich-Harris:
There's now a package designed for this very purpose: https://github.com/reconbot/js-off. It currently uses accessors rather than proxies, but the result should be the same

PR #335, opened 2018-08-02
Better/faster exporting

### LINKED ISSUE #325: export without rebuilding
This is perhaps related to #323: in some situations, you might want to re-export an app without rebuilding it (because rebuilding is generally the bottleneck). For example, your pages might be based on frequently-updating data.

In this situation it'd be useful to be able to use a previous build of the app — something like this...

```
sapper export --no-build
```

...or maybe this:

```
sapper export --build-dir=my-custom-build-dir
```

If you wanted to eke out even more performance you could do so by running the server in the background and curling it yourself. But that would be a nuisance because you'd have to reimplement all the crawling logic (or otherwise be careful to hit all the pages that need to be regenerated), so the CLI flag could be a useful halfway house.

cc @1wheel

Rich-Harris:
This improves the `sapper export` experience in a number of ways:

* It updates to the latest port-authority, which fixes a `setTimeout` bug that caused export to always take at least 5 seconds
* It adds a `--build` flag (that defaults to true) which allows you to reuse an existing build by running `sapper export --no-build`. This fixes #325 
* It adds a `--build-dir` flag that allows you to specify the export directory, which defaults to `.sapper/prod` (dev now defaults to `.sapper/dev`, so that you can see the outputs side-by-side)
* The terminal output is more informative. Instead of having duplicate entries for failed requests (one  for the saved error file, one for the error itself), they are grouped together, and large files are highlighted. In this screenshot, I've added a large page, a very large page, and a few pages that error out:

screen shot 2018-08-02 at 8 23 31 am

This is a lot more informative than what we had before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # sapper changelog
 
+## 0.15.4
+
+* Add `ignore` option ([#326](https://github.com/sveltejs/sapper/pull/326))
+
 ## 0.15.3
 
 * Crawl pages in parallel when exporting ([#329](https://github.com/sveltejs/sapper/pull/329))
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,8 @@
 {
   "name": "sapper",
-  "version": "0.15.2",
+  "version": "0.15.4",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@arr/every": {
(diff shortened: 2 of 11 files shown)


Rich-Harris:
One more additional feature: it now handles redirects, by creating a file like this:

```html
<!-- redirect-from/index.html -->
<script>window.location.href = "/redirect-to"</script>
```

PR #336, opened 2018-08-02
Use npm ci instead of npm install

Rich-Harris:


Rich-Harris:
Ugh, this doesn't work with Node 6, because it is bundled with an older npm.

Conduitry:
Is it possible/practical to have an extra step in the Node 6 CI script that first installs a more modern version of npm?

Rich-Harris:
Probably. A cursory Google didn't yield anything earlier so I gave up, but it seems like it should be possible. `npm i -g npm@latest` maybe? i'm just not sure if that'll actually speed things up or throw away the gains from doing `npm ci` in the first place

Conduitry:
Yeah true. Well failing some sort of thing where you can just install the newer version of npm once and then set that as some sort of snapshot to run future builds on (which sounds like the sort of thing that wouldn't be offered in the free tier of travis/appveyor) - then I guess just doing `npm ci || npm i` seems reasonable enough.

Rich-Harris:
ah yeah, good idea

PR #342, opened 2018-08-06
Remove basepath from deferred urls and add trailing slash to root 

aubergene:
The `pathname` that comes back from the request is missing the `basepath`. I thought maybe it's easier to always strip them off. Also the root request is missing the trailing slash, my fix is kinda ugly, I'm sure you have a more elegant solution.

Fixes #342


Rich-Harris:
d'oh! thank you — released 0.15.6. I added a test so that I can't accidentally break this again

PR #345, opened 2018-08-07
Fix Preload's Redirect

lukeed:
Strips the leading slash within `this.redirect(code, path)` calls. Because it's done whenever the `redirect` object is set, the equality check still passes.

Also chose to strip because of this line: https://github.com/sveltejs/sapper/blob/master/src/middleware.ts#L405

Lastly, the idea of relative redirects was mentioned in the issue. I personally don't think that should happen, but am open to implementing it if majority rules.

_Closes #291_


lukeed:
Failing AppVeyor is from previous commit (84a58f3); log here: https://ci.appveyor.com/project/Rich-Harris/sapper/build/837

Rich-Harris:
nice 👊

PR #354, opened 2018-08-11
root preloading value should only be set on navigations (not prefetching), closes #352

### LINKED ISSUE #352: Prefetching links sets root preloading value (and doesn't unset value when prefetching finishes either)
Since the `preloading` value is primarily used for showing progress indicators, it should not be set when user-transparent prefetching is happening.

s3ththompson:
This change moves `root.set({preloading: true})` out of `prepare_page` (which is run during navigations and prefetching) and into `navigate` (which is only run during navigation).

I didn't add a regression test as it looks like it's difficult to get nightmare to properly handle the `mousemove` event.

### DIFF
--- a/src/runtime/index.ts
+++ b/src/runtime/index.ts
@@ -226,8 +226,8 @@ function prepare_page(target: Target): Promise<{
 
 		const props = { path, query };
 		const data = {
-			preloading: false,
 			path,
+			preloading: false,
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Thank you! I removed the second `root.set({ preloading: false })` as that effectively happens in `render` anyway.

Will cut a release shortly, just having some trouble publishing to npm...

PR #363, opened 2018-08-19
[WIP] Slim down the bundle

Rich-Harris:
[This](https://beta.observablehq.com/@mbostock/package-dependencies) was a wake-up call:

sapper-dependencies

We can make Sapper smaller, and faster to install/launch, by being more careful about dependencies, and bundling what makes sense to.

So far I've removed `glob` in favour of `tiny-glob`, and `cheerio` in favour of a handful of regular expressions (we don't need to fully parse crawled HTML to find `<a href="...">`, we can just strip out comments and CDATA and look for `<a [\s\S]+?>` matches, which is also obviously much faster).

I'd love to get rid of micromatch since, astonishingly, [it's the reason for most of Sapper's 220 dependencies](https://npm.anvaka.com/#/view/2d/micromatch). Unfortunately that would mean either removing chokidar (which is certainly on the table), or persuading chokidar to ditch micromatch for an alternative.


Rich-Harris:
@terkelg I was wondering the exact same thing! Am just taking a look at the micromatch test/benchmark suite to see how feasible that would be — if we could create a drop-in micromatch replacement that passed all the tests and was as fast, it seems like it would be a no-brainer. We'd still need to convince the chokidar maintainers of course.

terkelg:
Sounds like a good plan!

Another alternative is to use `cheap-watch` by @Conduitry. We can integrate `tiny-glob` if glob support is needed for the watcher.

Rich-Harris:
The trouble is that any app that depends on Sapper will also depend on webpack, which depends on chokidar. We're basically stuck with it. Am trying to get my head round micromatch to see if replacing its innards is plausible, but it sprawls across dozens of repos, and I'm pretty much at the point of giving up.

lukeed:
Doesn't chokidar rely on micromatch?

Rich-Harris:
yes — the entire right-hand-side of the chart above is micromatch, and it's only there because chokidar depends on anymatch which depends on micromatch.

Rich-Harris:
Well, I think we have to shelve the chokidar/micromatch question for now. The changes in this PR reduce the size of a fresh installation of Sapper from 22.2Mb to 15.1Mb, so we're moving in the right direction even if there's still a ways to go.

The resulting graph looks all the more ridiculous:

sapper-0 16-dependencies

lukeed:
So good 😂 "beware the composable dependencies" I guess

PR #369, opened 2018-08-22
Don't try to crawl hundreds of pages simultaneously

Rich-Harris:
We could maybe tweak this to make it a global queue, but this is probably fine for any realistic website


src/api/export.ts
```diff
@@ -140,11 +144,13 @@ async function execute(emitter: EventEmitter, {
 
 					if (href) {
 						const url = new URL(href, base.href);
-						if (url.origin === origin) urls.push(url);
```
ikarigeorge: Hi, I think the `urls` declaration is not needed anymore in line 127

Rich-Harris: good catch, thanks — removed

Issue #305, opened 2018-07-02
`console.log` no longer emits from app/server.js compile


arxpoetica:
On latest sapper upgrade, I can no longer `console.log(...)` from the `app/server.js` file.

The breaking change appears to be between sapper `v0.13.3` and `v0.13.4` (works in `.3`, but not `.4`).

arxpoetica:
Small stab in the dark, but this might be the breaking change: https://github.com/sveltejs/sapper/commit/0800fa016b315b2dc9e7d48c872a2bfd66446faa

arxpoetica:
A little more info, it appears to be a timing issue. If I wrap console.log in a setTimout call, stuff starts showing up...

```js
setTimeout(() => console.log('yolo!'), 100)
```

shijiezhou1:
I'm having this issue these day my sapper just no longer console.log on any js file

PR #416, opened 2018-09-03
Blessed logs

Rich-Harris:
This improves (I think) dev mode logging by separating build logs from server logs, instead of interleaving them as is currently the case. It looks like this:

screen shot 2018-09-03 at 5 29 35 pm

The terminal is on the right hand side, split into a top half and a bottom half. The top half contains build logs (including warnings, such as the unused style notice), while the bottom half contains whatever your server logs out. Invalidating the build clears the top half; if the server successfully restarts, the bottom half is cleared.

For terminals that are wider than they are tall, you can toggle between horizontal and vertical splits by clicking on the divider, which is also draggable.

Also fixed in this PR: #305.

To get the old behaviour back, do `sapper dev --stream`.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -15,13 +15,15 @@ prog.command('dev')
 	.option('--hot', 'Use hot module replacement (requires webpack)', true)
 	.option('--live', 'Reload on changes if not using --hot', true)
 	.option('--bundler', 'Specify a bundler (rollup or webpack)')
+	.option('--stream', 'Stream logs, inside of boxing them', false)
 	.action(async (opts: {
 		port: number,
 		open: boolean,
 		'dev-port': number,
 		live: boolean,
 		hot: boolean,
-		bundler?: string
+		bundler?: string,
+		stream: boolean
 	}) => {
 		const { dev } = await import('./cli/dev');
 		dev(opts);
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -15,7 +15,7 @@ prog.command('dev')
 	.option('--hot', 'Use hot module replacement (requires webpack)', true)
 	.option('--live', 'Reload on changes if not using --hot', true)
 	.option('--bundler', 'Specify a bundler (rollup or webpack)')
-	.option('--stream', 'Stream logs, inside of boxing them', false)
+	.option('--stream', 'Stream logs, instead of boxing them', false)
 	.action(async (opts: {
 		port: number,
 		open: boolean,
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sapper",
-  "version": "0.19.0",
+  "version": "0.19.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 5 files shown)


src/cli.ts
```diff
@@ -15,13 +15,15 @@ prog.command('dev')
 	.option('--hot', 'Use hot module replacement (requires webpack)', true)
 	.option('--live', 'Reload on changes if not using --hot', true)
 	.option('--bundler', 'Specify a bundler (rollup or webpack)')
+	.option('--stream', 'Stream logs, inside of boxing them', false)
```
Conduitry: Should be 'INSTEAD of boxing them'?

Conduitry:
From Blessed's readme:

> Blessed is over 16,000 lines of code and terminal goodness.

is there no other reasonable way to achieve this result? It looks like this library does a lot more than we need here, and that sounds like a sizeable `node_modules` price for everyone (including, I'm assuming, people running production apps).

src/cli.ts
Rich-Harris: Yes. yes it should

Rich-Harris:
> is there no other reasonable way to achieve this result?

You'd think. I spent some time looking into this today, and blessed felt like the most realistic option, even though it's unmaintained. (I had a go at implementing the functionality myself — ouch.)

I was thinking that perhaps we could fork blessed and remove all the stuff we don't need. But I didn't want to do that until we could see if we like how it looks and behaves. It adds just under two megabytes to an installation, which probably isn't catastrophic

Conduitry:
To reiterate what I said on Discord: I accept that this functionality is nicer than what there is now. I also begrudgingly accept that this is probably the best way to go about this for now, and we can think about a stripped-down fork of Blessed or something later on.

kylecordes:
Okay, here's my pitch for a lighter weight way to do this:

Make the command line log output be roughly as it was before. Just some straightforward marking of whether each line is build log or server log.

Then develop a simple application... using Sapper. It's job will be to (in the node server) watch the logs. Make them available to a connected UI using a SSE (fit for purpose) or web socket (more popular). In the browser, show the logs. Use a layout somewhat like what is shown in the Blessed screen shot above. But because it would be a Sapper application, it would have the same dependencies Sapper already has. And it would mean there is a Sapper application of nontrivial complexity, in the Sapper repo, which of course could be used to test Sapper further during CI. The user is of course already in the browser (for their application); and for some kinds of testing it might be helpful to be able to bring it up remotely (trivial while running in the browser). It could also grow into an increasingly complete development status/tooling screen.

This also has a very charming "meta-" appeal.

Rich-Harris:
@kylecordes that is brilliant, and also breath-takingly obvious. The only real downside is you don't have as nice an experience using an editor's integrated terminal, but the upsides far outweigh it as far as I can see.

arxpoetica:
I just came across this w/ the fix of 305 (thanks Rich!). I love @kylecordes idea so much! Tooling around Sapper apps (not just logs!) is just so brilliant. I'm imagining such a thing being useful in so many ways. Love it.

shijiezhou1:
Screen Shot 2021-08-15 at 6 50 06 PM
Why the `sapper dev --stream` is not the option

PR #423, opened 2018-09-07
Add SAPPER_TIMESTAMP environment variable

nolanlawson:
See https://github.com/nolanlawson/pinafore/issues/80 for why this is needed. I'll also open up a documentation pull request to explain the feature.


nolanlawson:
Documentation PR: https://github.com/sveltejs/sapper.svelte.technology/pull/33

Rich-Harris:
Thanks, this makes sense. I wonder if this is something that is best solved in userland though? The `timestamp` import is there as a convenience but it could equally be this:

```diff
-import { timestamp, assets, shell, routes } from './manifest/service-worker.js';
+import { assets, shell, routes } from './manifest/service-worker.js';

+const timestamp = process.env.SAPPER_TIMESTAMP;
const ASSETS = `cache${timestamp}`;
```

```diff
import resolve from 'rollup-plugin-node-resolve';
import replace from 'rollup-plugin-replace';
import commonjs from 'rollup-plugin-commonjs';
import { terser } from 'rollup-plugin-terser';
import config from 'sapper/config/rollup.js';

const mode = process.env.NODE_ENV;
const dev = mode === 'development';

export default {
	input: config.serviceworker.input(),
	output: config.serviceworker.output(),
	plugins: [
		resolve(),
		replace({
+			'process.env.SAPPER_TIMESTAMP': process.env.SAPPER_TIMESTAMP || Date.now()
			'process.browser': true,
			'process.env.NODE_ENV': JSON.stringify(mode)
		}),
		commonjs(),
		!dev && terser()
	],

	// temporary, pending https://github.com/rollup/rollup/pull/2432
	watch: {
		chokidar: { disableGlobbing: true }
	}
};
```

From a 'teach a man to fish' perspective (not to mention keeping the core as simple as possible) this feels better, but maybe there's something I've overlooked? Or maybe this is a common enough requirement that it deserves special treatment?

nolanlawson:
This sounds like a great suggestion to me. I totally overlooked the fact that this can be solved in userland. I'll just update the documentation PR and then let's call that good!

PR #425, opened 2018-09-07
Add cache-control:immutable for immutable assets

nolanlawson:
Since the `client/` assets are immutable and uniq'ed by a checksum, we can add [immutable caching](https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/) headers to allow browsers that support it (Firefox, Edge currently) to avoid unnecessary revalidation requests.

The way Chrome currently works is that it uses heuristics based on the `max-age` rather than `immutable`, so this will also work for Chrome. Not sure about Safari.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -151,7 +151,7 @@ export default function middleware(opts: {
 
 		serve({
 			prefix: '/client/',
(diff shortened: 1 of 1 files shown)


Rich-Harris:
TIL! Fantastic, thank you

PR #428, opened 2018-09-07
Set service-worker max-age to 0

nolanlawson:
As of Chrome 68 and Firefox 57 (and Edge and Safari), service workers are always revalidated regardless of the `cache-control` directive. So it's recommended to use `cache-control: max-age=0`, since it aligns older browsers with the newer behavior and also makes it clear to developers what's going on.

For the service worker sourcemap file I don't think it really matters what we do, since it's a request that is only made when DevTools are open, but I'd say align it with `service-worker.js` just in case the two get out of sync somehow.

Sources:

- https://developers.google.com/web/updates/2018/06/fresher-sw
- https://stackoverflow.com/a/38854905

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -141,12 +141,12 @@ export default function middleware(opts: {
 
 		fs.existsSync(path.join(output, 'service-worker.js')) && serve({
 			pathname: '/service-worker.js',
-			cache_control: 'max-age=600'
+			cache_control: 'max-age=0'
 		}),
 
 		fs.existsSync(path.join(output, 'service-worker.js.map')) && serve({
 			pathname: '/service-worker.js.map',
-			cache_control: 'max-age=600'
+			cache_control: 'max-age=0'
 		}),
 
 		serve({


lukeed:
They should be `private, no-cache`, no?

nolanlawson:
I mean, `max-age=0` should be the same as `no-cache`, right? Not sure about `private`; my intuition is actually to set `public` for everything, but I don't know if that's a sensible default since I'm not very familiar with how proxies handle that stuff.

lukeed:
Well, `service-worker`s should completely bypass any cache so that it doesn't ever risk being stale.

While it's true that `max-age=0` is effectively the same thing, you're still adding an entry into the browser _and_ CDN cache, only to have them immediately expire. It's like doing this:

```js
let obj = {};

function onRequest()
  obj.foo = 'bar';
  if (obj.foo) {
    delete obj.foo;
  }
}

onRequest();
onRequest();
// etc
```

The combination of [`private` and `no-cache`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Directives) completely skips the cache assignment(s), avoiding useless work. It's as equally supported as the `public` directive.

Rich-Harris:
From that MDN page:

> ## Preventing caching
> To turn off caching, you can send the following response header. In addition, see also the `Expires` and `Pragma` headers.

>     Cache-Control: no-cache, no-store, must-revalidate

Issue #415, opened 2018-09-03
this.cache(...) in preload


Rich-Harris:
At present there's no way to set a cache header for a page. It might be useful to be able to do this:

```js
export default {
  async preload({ params }) {
    this.cache('max-age', 30 * 60 * 1000); // cache for 30 minutes
    await this.fetch(...).then(...);
  }
};
```

(An interesting question that arises is whether that should change behaviour on the client as well. Should the browser cache the preloaded data for the specified duration and reuse it if the same URL is visited later? This would make some navigations faster but would increase memory usage and introduce a danger of mutation bugs, so perhaps it's better to rely on the service worker.)

Or would it be better to have a more generic `this.headers({...})` function?

nsivertsen:
I'm in favour of having something like this in Sapper. I feel that your suggestion, especially in the general form of `this.headers({...})`, is related to #362, especially to [my comment](https://github.com/sveltejs/sapper/issues/362#issuecomment-418701202) about route-specific middleware for things like caching API responses server-side, but which could also be used for setting headers. Another solution to both problems would be to add a `middleware` function to routes that has access to `req` and `res`:

```js
export default {
  async middleware({ req, res }) {
    res.set('Cache-Control', `max-age=${30 * 60 * 1000}`); // cache for 30 minutes
    // do other stuff with `req` and `res`
  }
};
```

The advantage of this approach is that you can do anything that Express or Polka allow you to do, and are not limited to the functions that Sapper exposes on the content object.

The disadvantage is that you forfeit the ability of also doing something like caching things client-side as you propose above with `this.cache(...)`.

mrkishi:
I would prefer a server-specific solution like @nsivertsen's myself.

I'd instinctively associate a `this.cache()` inside `preload` with the preloaded _data_ only -- meaning no matter what crazy thing I did inside `preload`, the _returned_ data would be cached locally, where local means either the client or server.

dogada:
> Or would it be better to have a more generic this.headers({...}) function?
This is IMO a better solution because if I will set a cache header on server, browser will cache the page also (in addition to the intermediate caching proxies). I will probably also need to set a `Vary` header. I think Sapper should provide access to the request object also like Next.js do in `getInitialProps`. It may look like:
```
preload({req, res}) {
  if (req) this.headers(...)
}

antony:
[...]

PR #429, opened 2018-09-07
Use consistent cache-control:max-age=600 for HTML pages

nolanlawson:
Currently there is a `cache-control:max-age=600` set for `index.html`, but no `cache-control` header set for the HTML pages served from routes. This PR makes both of them consistently use `max-age:600` in production and `no-cache` in development.

In a future PR, I would like to make all of these cache-control headers configurable:

- immutable assets (`'max-age=31536000, immutable'` by default)
- routes, `index.html` (`'max-age=600'` by default)
- service worker (`'max-age=0'` by default, see #428)

But for now I would settle for making them consistent. :)

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -136,7 +136,7 @@ export default function middleware(opts: {
 
 		fs.existsSync(path.join(output, 'index.html')) && serve({
 			pathname: '/index.html',
-			cache_control: 'max-age=600'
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Makes sense — thanks. More configurability would definitely be good in the longer term — I opened https://github.com/sveltejs/sapper/issues/415 the other day

PR #436, opened 2018-09-12
Enable debugging in Chrome and VS Code - fixes #435

### LINKED ISSUE #435: Debugging Sapper app server in dev mode in Chrome or VS Code not possible
One can debug the server portion of a Sapper app quite well using ndb (because it handles child processes intelligently out of the box), but currently it's not possible to do so with Chrome's DevTools or VS Code. If one tries to run:

```bash
$ node --inspect node_modules/sapper/dist/cli.js dev
```

from inside a Sapper app, one gets:

```
Debugger listening on ws://127.0.0.1:9229/afd0539e-40c6-474e-bf06-4caa40c57e41
For help, see: https://nodejs.org/en/docs/inspector
✔ server (765ms)
✔ client (778ms)
> Server crashed
Starting inspector on 127.0.0.1:9229 failed: address already in use

✔ service worker (13ms)
```

...because `sapper dev` spawns a new process for the server, and by default this new process inherits all arguments from the parent process, including the port on which the DevTools server should run, and this port is obviously already occupied.

The fix is quite simple and I'm already working on a PR, just wanted to leave this here in the meantime.

nsivertsen:
This PR makes it possible to start and connect to a debugging session inside the dev server of a Sapper app from Chrome and VS Code. Previously this only worked from ndb.

See #435

**Chrome**:

1. `$ cd my_sapper_project`
2. `$ node --inspect node_modules/sapper/dist/cli.js dev`
3. visit `chrome://inspect` in Chrome
4. click on "Open dedicated DevTools for Node"
5. (you might have to add the folder of your Sapper project to the workspace)

**VS Code**:

1. `Debug > Add Configuration...`
2. Use the following `launch.json`:
  ```
  {
    "version": "0.2.0",
    "configurations": [
      {
        "type": "node",
        "request": "launch",
        "name": "dev",
        "program": "${workspaceFolder}/node_modules/sapper/dist/cli.js",
        "args": [
          "dev"
        ],
        "autoAttachChildProcesses": true
      }
    ]
  }
  ```
3. `Debug > Start Debugging`

I've found the experience of coding in VS Code with a dev server + debug session running in the background quite pleasurable: Make some changes, hit save, the app is rebuilt, all while your breakpoints and watch expressions stay in place.

I've tried this in Webstorm, and it works there as well, until you change a file and hit save. Webstorm tries to hold onto the previous server process (which has been killed) and gets confused when you try to carry on stepping through your code. I assume this is a bug in Webstorm, and not a Sapper problem.

Fixes #435

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -36,6 +36,8 @@ class Watcher extends EventEmitter {
 	live: boolean;
 	hot: boolean;
 
+	devtools_port: number;
+
 	dev_server: DevServer;
 	proc: child_process.ChildProcess;
 	filewatchers: Array<{ close: () => void }>;
@@ -57,6 +59,7 @@ class Watcher extends EventEmitter {
 		'dev-port': dev_port,
 		live,
 		hot,
+		'devtools-port': devtools_port,
 		bundler,
 		webpack = 'webpack',
 		rollup = 'rollup',
@@ -68,6 +71,7 @@ class Watcher extends EventEmitter {
 		'dev-port': number,
 		live: boolean,
 		hot: boolean,
+		'devtools-port': number,
 		bundler?: string,
 		webpack: string,
 		rollup: string,
@@ -83,6 +87,8 @@ class Watcher extends EventEmitter {
 		this.dev_port = dev_port;
 		this.live = live;
 		this.hot = hot;
+
+		this.devtools_port = devtools_port;
 
 		this.filewatchers = [];
 
@@ -128,6 +134,9 @@ class Watcher extends EventEmitter {
 		if (this.bundler === 'rollup') copy_shimport(dest);
 
 		if (!this.dev_port) this.dev_port = await ports.find(10000);
+
+		// Chrome looks for debugging targets on ports 9222 and 9229 by default
+		if (!this.devtools_port) this.devtools_port = await ports.find(9222);
 
 		let manifest_data: ManifestData;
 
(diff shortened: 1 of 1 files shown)


src/api/dev.ts
```diff
@@ -57,6 +59,7 @@ class Watcher extends EventEmitter {
 		'dev-port': dev_port,
 		live,
 		hot,
+		'devtools-port': devtools_port,
```
nsivertsen: I've added this here with an eye towards making the DevTools port a command-line switch in the future.

Rich-Harris:
Oh wow, this is awesome! Thank you

QuantumInformation:
@nsivertsen can you share your exact settings and steps to debug this in webstorm please?

I tried 
Screenshot 2019-08-14 at 12 28 40

nsivertsen:
Hi @QuantumInformation, I unfortunately don't current have access to the computer I tried this on with Webstorm, but your config seems right, apart from the working directory (I would set that one level higher up), and you need to pass the `dev` flag to `cli.js`. I would turn on source maps and debug the source.

QuantumInformation:
Great, that seems to have worked now, will play around.

Screenshot 2019-08-15 at 10 45 53

QuantumInformation:
I can't debug scripts in svelte files though in webstorm.

QuantumInformation:
I tried setting breakpoints in chrome but they are never hit, did you have any luck?

Screenshot 2019-08-16 at 17 52 25

QuantumInformation:
For posterity: ndb is https://github.com/GoogleChromeLabs/ndb

QuantumInformation:
I made a video to show this debugging in action:

https://www.youtube.com/watch?v=x155wGkbKm8

Issue #439, opened 2018-09-13
Legacy client code not parsing __SAPPER_CSS_PLACEHOLDER


silentworks:
It seems the legacy client code is not parsing the PLACEHOLDER text, this can be seen from the resulting output in the build directory.

Here is part of the output from `build/client/client.******.js`:

```js
{
    pattern:/^\/about\/?$/,
    parts:[{
        component: {
            js:()=>import("./about.2dbe86cf.js"),
            css:["about.2dbe86cf.css","chunk.953bc4b7.css"]
        }
    }]
}
```

Here is part of the output from `build/client/legacy/client.******.js`:

```js
{
    pattern:/^\/about\/?$/,
    parts:[{
        component:{
            js:function(){return import("./about.627078a3.js")},
            css:"__SAPPER_CSS_PLACEHOLDER:about.html__"
        }
    }]
}
```

This seem to be an issue at build time, it seems that `build/client/client.******.js` gets parsed by the `extract_css` function and written to file https://github.com/sveltejs/sapper/blob/91f2c6e49c2b35e485a56b157c9f6c9ee0054f80/src/core/create_compilers/extract_css.ts#L210, while `build/client/legacy/client.******.js` doesn't.

silentworks:
I have a solution but it feels very much like a nasty hack.

In `build.ts` before line 83 I added `client_result.to_json(manifest_data, dirs);` and then inside of `extract_css.ts` after line 172 I added: 

```js
if (process.env.SAPPER_LEGACY_BUILD)
        main = 'legacy/' + client_result.assets.main;
```

Rich-Harris:
Released 0.20.4 with the fix — thank you

PR #441, opened 2018-09-13
Fix for legacy manifest file

silentworks:
Related to https://github.com/sveltejs/sapper/issues/439


Rich-Harris:
Nice fix, thank you. The side-effecty `to_json` method looks even more awkward now — we'll have to refactor it eventually. This tides us over till then!

silentworks:
Yeah I was trying to think of a better way to do this, but unfortunately time wouldn't allow me.

PR #446, opened 2018-09-21
Fix filenames

mrkishi:
In the same vein as rollup/rollup-plugin-svelte#38, this just escapes filenames that'll be used in source strings.

Additionally, adds a missing `posixify` to that one filename in order to fix Windows support.

Possibly enough for sveltejs/sapper-template#77.


Rich-Harris:
Thank you! Released 0.21

Issue #444, opened 2018-09-19
Optimise apps with only one route


Rich-Harris:
[This](https://twitter.com/rauchg/status/1041439324091252736) has been a long term goal...

<img width="633" alt="screen shot 2018-09-19 at 7 29 13 pm" src="https://user-images.githubusercontent.com/1162160/45787268-5e414780-bc42-11e8-984e-4f6eb515227b.png">

...but Guillermo's tweet has got me thinking about what concrete steps we can take in that direction.

One is to avoid hydrating static routes, and in fact that's one of the oldest issues on this repo: https://github.com/sveltejs/sapper/issues/8

But another, simpler step we can take is to optimise for the (not that uncommon!) case where you only have a single route. In that case, we don't need a router! Which means we don't need `sapper/runtime.js`.

Ideally the user wouldn't need to think about this. That means that we'd need to change this...

```js
import { init } from 'sapper/runtime.js';
import { manifest } from './manifest/client.js';

init({
  target: document.querySelector('#sapper'),
  manifest
});
```

...to something like this:

```js
import * as app from './manifest/client.js';

app.init({
  target: document.querySelector('#sapper')
});
```

In the case where the app is completely static (i.e. no interactivity in the `src/routes/index.html` component), the manifest file could be as simple as this:

```js
// src/manifest/client.js
export function init(){} // noop
```

I'm glossing over a few things here. What becomes of `goto` and `prefetchRoutes` etc — do they also become part of `src/manifest/client.js` instead of `sapper/runtime.js` (and if so, does it still make sense to call it a 'manifest')?

Details aside, this feels like a promising direction that would make it a complete no-brainer to use Sapper for projects of all sizes.

Rich-Harris:
Was just thinking about https://github.com/sveltejs/sapper/issues/445 in the context of this issue, and it occurred to me that there's a neat approach to all of this: we get rid of the Sapper middleware *and* `sapper/runtime.js`.

Instead, that logic goes into the generated files, not unlike the way Svelte dissolves itself into a compiled component.

Here's what I'm thinking:

```diff
// app/client.js
-import { init } from 'sapper/runtime.js';
-import { manifest } from './manifest/client.js';
+import * as sapper from './__sapper__/client.js';

sapper.init({
- manifest,
  target: document.querySelector('#sapper')
});
```

```diff
// app/server.js
import sirv from 'sirv';
import polka from 'polka';
-import sapper from 'sapper';
import compression from 'compression';
[...]

PR #453, opened 2018-09-30
move app logic into templates (#444)

Rich-Harris:
This lays the foundation for #444, by moving app logic out of the Sapper package (`sapper/dist/middleware.js` and `sapper/runtime.js`) and into the generated files.

Those generated files, which used to be `project/src/manifest/{client,server,service-worker}.js` are now written to the `project/src/__sapper__` directory. That is still up for debate. One possibility is to put *all* generated files in `project/__sapper__` (i.e. outside `src`), so that it would look like this:

```
├ __sapper__
│ ├ .build
│ │ ├ dev
│ │ │ ├ # dev files here (during `sapper dev`)
│ │ ├ prod
│ │ │ ├ # prod files here (during `sapper export`)
│ ├ client.js
│ ├ server.js
│ ├ service-worker.js
├ build (output of `sapper build` — should this be `dist` instead?)
├ export (output of `sapper export` via `__sapper__/.build/prod`)
├ src
│ ├ # ...
├ static
│ ├ # ...
```

TODO

* [x] settle the folder structure question
* [x] probably need to provide some more informative errors
* [x] update docs


lukeed:
I think anything generated should live outside the working files entirely. I can totally go either way on this, but that'd still be my vote.

The latest structure has `src/manifest/*`, which is all generated, as well as `.sapper/*` files. It's trivial for Rollup/Webpack to maintain correct file aliases across directories so that shouldn't be involved in the decision IMO.

This is especially true if we rename the manifest imports to something like this:

```js
import { manifest } from 'sapper/server';
//
import { manifest } from 'sapper/client';
//
import { timestamp, assets, shell, routes } from 'sapper/sw';

// or, of course, use 1 file and export manifests/partials from it

import { server as manifest } from 'sapper/manifest';
```

lukeed:
Ah, I forgot this issue drops the manifests entirely 😇 

Either way, I'd still vote for having all generated files in a single directory, preferably away from my user-source files.

lukeed:
Sorta. I think this makes the most sense:

```
├ __sapper__
│ ├ dev
│ │ ├ # dev files here (during `sapper dev`)
│ ├ build
│ │ ├ # prod files here (during `sapper build`)
│ ├ export
│ │ ├ # prod files here (during `sapper export`)
│ ├ client.js
│ ├ server.js
│ ├ service-worker.js
│
├ src
│ ├ # ...
├ static
│ ├ # ...
```

Rich-Harris:
Oh, interesting. I can definitely see the appeal of that. The only downside is that `node build` becomes `node __sapper__/build` (to run the built app) and instead of `npx serve export` to serve an exported app we have `npx serve __sapper__/export`. But that doesn't feel like a dealbreaker.

lukeed:
I would go further and rename the directory to `sapper` and rename `server.js` to `index.js`, so that running a server becomes `node sapper` and running the exports becomes `npx sirv sapper/export`

Rich-Harris:
I think it needs to be something like `__sapper__` rather than `sapper` for the same reasons `src/manifest` doesn't quite work — it's not totally obvious that it contains generated files, and it's likely to sit between the user's directories (e.g. the svelte.technology project has a `content` directory and a `src` directory, and it's weird if `sapper` goes between them).

Also, `__sapper__/server.js` *isn't* executable, it's just something that `src/server.js` imports. The file written to `build` is already called `index.js` — that's how `node build` works.

(Btw the reason I say `npx serve export` instead of `npx sirv export` is because I can never be bothered to look up how to invoke the binary of a package that doesn't share its name. I don't suppose you've ever considered merging `sirv` and `sirv-cli`...?)

lukeed:
RE `__sapper__`: totally understood, I agree which is why I didn't include the suggestion originally. 

RE `server.js`: gotcha – made bad assumption that that had changed as well. Sounds good.

RE `sirv`: Lol thew that in there to see if you'd notice 😆 It doesn't matter to me that you use it! I get the frustration, but it's to avoid downloading the CLI counterpart when you're just after the server module. While you and this circle are keen on tree-shaking all the way down the line, the 99% aren't or won't; so I've solved/avoided that problem thru separation of concerns rather than changing their toolchains & habits.

Rich-Harris:
Ok, I've put everything in Sapper. It *is* nice and neat. Will try this branch out with a few different projects and see if I've missed anything.

No-one's apps would be adversely affected by having `sirv-cli` inside `sirv`, tree-shaking or not — it'd just mean having to install a tiny handful of additional dependencies that don't end up getting used. Personally I'd say it's well worth it for the ease of doing `npx sirv .` wherever you are. But that's just me 😀

PR #456, opened 2018-10-01
Fix rollup input paths on Windows

mrkishi:
Normalizes input paths so they match rollup's `chunk.modules`.


Rich-Harris:
nice, thanks

PR #465, opened 2018-10-05
fix path for component css when using a baseUrl

brittharr:
Fixes a `500` from the missing "/" before "client" when loading a component's css
> Refused to apply style from 'http://localhost:3000/custom/pathclient/chunk.8d05bbb2.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.

Let me know if I should push my sapper-template fork that reproduces the original error! I think this only happens when using express and a basepath.

### DIFF
--- a/templates/src/client/app.ts
+++ b/templates/src/client/app.ts
@@ -340,7 +340,7 @@ export function prepare_page(target: Target): Promise<{
 }
 
 function load_css(chunk: string) {
-	const href = `${initial_data.baseUrl}client/${chunk}`;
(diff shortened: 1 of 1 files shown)


Rich-Harris:
hey 👋 Thanks for this. Turns out the solution is simpler still — I messed up when I included the `baseUrl` in the first place, as the link is resolved against the base path anyway. Using an absolute path that includes the base path works, but it means it's loading the CSS twice because it doesn't notice that the server-rendered HTML already included it. So I'll close this in favour of #466

PR #467, opened 2018-10-06
[WIP] overhaul tests

Rich-Harris:
The test suite is a mess. Everything is bunged together in a confusing, ad hoc order, and it's missing essential things like Rollup tests, dev mode tests, and more.

My current goals:

* Replace Nightmare with Puppeteer, which appears to be significantly faster, and which I'd expect to stay more current (it does mean our tests are still Chromium-based, but we can worry about cross-browser stuff another time)
* Rewrite tests in TypeScript. Among other things this means we can use async await which is much nicer for this kind of thing, without waiting for Node 6 to die out
* Instead of having a single app to test every aspect of Sapper, we can have multiple apps (e.g. one for testing basepath-specific behaviour, one for testing export-specific behaviour, one for testing dev-specific behaviour, etc). By default they should use Rollup since it's faster than webpack for small codebases

I'm also using Sapper's JavaScript API rather than shelling out to the CLI, and in the process learning that the API is really quite awful. That's not a huge problem since it's not really the intended way of using it, but it presents a good opportunity to improve the architecture in a couple of places.


Rich-Harris:
The tests now run in ~20 seconds rather than ~50 (on this machine), and the output is much less garbled. Still a bit of work to do to DRY out the code etc

PR #468, opened 2018-10-08
[WIP] Better api

Rich-Harris:
This improves the JavaScript API somewhat — it makes it promise-based, and less reliant on magic env vars. No real external benefit, just better code organisation.

TODO:

* [ ] I'm inclined to remove `sapper start` and `sapper upgrade` from the CLI, since they're unmaintained, untested and currently do anything, and no-one has noticed.
* [ ] Less indirection in cli.ts — it's now practical to use the API directly from there, rather than having wrapper objects
* [ ] Remove the `process.chdir()` and `process.env.NODE_ENV = 'production'` stuff from the tests, since this should no longer be necessary


src/api/dev.ts
```diff
@@ -54,34 +69,23 @@ class Watcher extends EventEmitter {
 	}
 
 	constructor({
-		src = locations.src(),
-		dest = locations.dest(),
-		routes = locations.routes(),
+		cwd = process.cwd(),
```
lukeed: I find it better to do `path.resolve(opts.cwd || '.')` then `path.join` directories from there. It'll arrive at the same value most of the time, but allows dev to specify a `--cwd foobar` flag, which is useful for testing or some power-user shiz

src/cli.ts
```diff
@@ -21,10 +32,93 @@ prog.command('dev')
...
+				if (first) {
+					console.log(colors.bold.cyan(`> Listening on http://localhost:${event.port}`));
+					if (opts.open) {
+						const { exec } = await import('child_process');
+						exec(`open http://localhost:${event.port}`);
+					}
+					first = false;
+				}
+
+				// TODO clear screen?
```
lukeed: Yeah, but I would [soft-clear](https://github.com/lukeed/console-clear#clearkeephistory) here. I've found that whenever I do a hard clear I get some push back.

src/api/dev.ts
Rich-Harris: Doesn't it amount to the same thing currently? If `cwd` is specified (as an option, not a CLI flag, which isn't supported just yet... TK) then it uses that, otherwise it uses `process.cwd()`. It's how the tests in this PR work (`await build({ cwd: __dirname })`)

src/cli.ts
Rich-Harris: Oh I didn't know about console-clear — handy, thanks. This is an old TODO though (I just copied it from a now-deleted file), should probably be dealt with in a separate PR

src/api/dev.ts
lukeed: Right right, I was thinking towards a flag.. I left my comment in wrong place.

The flag would be nice as it means one can use something other than the `SAPPER_*` global to set the cwd.

Rich-Harris: Should we discard `SAPPER_BASE` etc altogether? With the new API they're not really pulling their weight, and they're not documented anywhere

Rich-Harris: (I guess if we wanted to be rigorous we'd have flags for all the various path options — `--src`, `--routes`, one for the `__sapper__` directory (`--output`?) that's separate from the `--build-dir` option... maybe it should be a follow-up PR now that I think about it

lukeed: Mmmm, I'll defer to you on that. I think as long as the `--cwd` flag makes it in (or similar), it'll be okay. 

I may also be biased since I prefer relying on flags or options rather than globals. As long as they're all removed (or all kept), it won't be confusing.

I've only seen the BASE/ASSETS globals used in deploys to Lambda or GCP Functions. Every other is internal-facing only iirc.

PR #470, opened 2018-10-09
Allow SSI comments

aubergene:
Dear Rich, 
Pleeeeeeeease allow [SSI](https://en.wikipedia.org/wiki/Server_Side_Includes) directives.
I will owe you 🍻 but not as a bribe this is a valid enhancement of the Sapper stack. 
Thank you kindly
JB

### DIFF
--- a/src/api/utils/minify_html.ts
+++ b/src/api/utils/minify_html.ts
@@ -7,6 +7,7 @@ export default function minify_html(html: string) {
 		conservativeCollapse: true,
 		decodeEntities: true,
 		html5: true,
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Since you asked nicely! You're probably right that we should eventually find a nice way to expose the minification config, but this'll do nicely for the time being. Released 0.23

lukeed:
What is this magic

aubergene:
Thanks! 🍻 will be dispatched by my NYC operative agents or when I'm next in town

PR #471, opened 2018-10-12
include css depended upon by entry point, even if also depended on by a lazily-loaded component

Rich-Harris:
This fixes a bug whereby CSS depended upon by the entry point (which statically depends on `src/routes/_layout.html` if it exists, along with `src/routes/index.html` and `src/routes/_error.html`) would be omitted if it is *also* depended upon by a lazily-loaded chunk.

The code is a bit woolly, and it could use some tests, so I'm just going to leave it here for the moment

### DIFF
--- a/src/core/create_compilers/extract_css.ts
+++ b/src/core/create_compilers/extract_css.ts
@@ -195,7 +195,7 @@ export default function extract_css(client_result: CompileResult, components: Pa
 			});
 
 			if (chunks_with_css.has(chunk)) {
-				css_dependencies.push(chunk.file);
+				css_dependencies.push(chunk.file.replace(/\.js$/, '.css'));
 
 				chunk.modules.forEach(file => {
 					unclaimed.delete(file);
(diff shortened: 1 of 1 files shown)


arggh:
I tried running Sapper from `missing-css` branch, which caused almost all CSS styles to disappear, but not all. At a glance it seems like styles used on multiple routes, such as the nav, are shipped.

Rich-Harris:
oh, wait, I forgot I ran into a similar thing earlier today — [d'oh](https://github.com/sveltejs/sapper/pull/471/commits/44736754adfeec016572d207aedbed4a30fe48fc). Can you try again please?

arggh:
@Rich-Harris Works now!

Rich-Harris:
Oddly enough I couldn't quite figure out how to make a failing test for this, even though I managed to reproduce the bug in a separate project. Rather than wrestle with it all day I've released 0.23.2

PR #472, opened 2018-10-13
Fix hash link reliability

DayBr3ak:
It avoids preventing an event basically. Test added. Let me know if the test is ok.

I'm waiting to see if all tests pass because 10sec of timeouts on my machine is too short for the runner to launch.


Rich-Harris:
Wonderful, thank you! Released 0.23.1

PR #473, opened 2018-10-13
Handle tag click from another page

### LINKED ISSUE #341: Anchor tags cannot be accessed from other pages
Navigation to an anchor link from another page doesn't work, and only takes you to the top of the page. However, navigation to an anchor link does work if you are already on the page. 

I encountered this issue when trying to create drop-down top navigation items that take you to an anchor within another page.

You can reproduce the error here: https://github.com/alizauf/another-anchor-test 

This may be related to https://github.com/sveltejs/sapper/issues/331 (and thanks @mhkeller) for a convenient sample repo to easily demonstrate.

DayBr3ak:
Clicking on a link directing to a new page with a tag works.

Fix #341

### DIFF
--- a/templates/src/client/app.ts
+++ b/templates/src/client/app.ts
@@ -106,18 +106,17 @@ export function scroll_state() {
 	};
 }
 
-export function navigate(target: Target, id: number, noscroll = false): Promise<any> {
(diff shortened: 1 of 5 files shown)


DayBr3ak:
Not sure why travis has connection errors on tests.

Rich-Harris:
Thanks so much! Yeah, Travis flakes out sometimes

Issue #481, opened 2018-10-18
Custom build directory: server `build_dir` relative to project root


laurentpayot:
I'm generating a firebase-hosted sapper app in a custom directory within my project with
`sapper build firebase/functions/app`

When I serve my build with Firebase I get:
```
Error: ENOENT: no such file or directory, open 'firebase/functions/app/build.json'
    at Object.fs.openSync (fs.js:646:18)
    at fs.readFileSync (fs.js:551:33)
    at Ie (/home/laurent/projects/my-project/firebase/functions/app/server/server.js:70:20865)
```
This is because the `build_dir` constant in the server template is set to `firebase/functions/app` but the server is not run from the project root directory by Firebase. `build_dir` should be able to be set to a correct value by Firebase (using `__dirname`).

This issue appeared when I upgraded sapper to use the new folder structure.
Before that, `process.env.SAPPER_DEST` (manually set) was taken into account and the server was working, as described in #368.

mikenikles:
Hi @laurentpayot, have you figured out a way to host on Firebase? I'm in the same situation as you describe in this issue and wonder if there's a solution before I start digging into it.

PR #482, opened 2018-10-18
Custom build directory server fix

laurentpayot:
This is a fix for #481


benmccann:
It looks like this file doesn't exist anymore and the output directory is now set here: https://github.com/sveltejs/sapper/blob/9e5a140c9658c3c01d00f2fd8a20d25f6f752f42/src/core/create_app.ts#L182

antony:
Closing for now as it's stale. We can rebase+re-open with tests if this is still required

Issue #483, opened 2018-10-18
Better prefetch cache?


mrkishi:
Currently, sapper will only cache one prefetched page at a time. That works fine and means we don't have to deal with memory growth, but consider the following scenario:

- User hovers link to `/a`, `/a`'s preloaded data gets saved
- User hovers link to `/b`, `/a`s preloaded data gets discarded and `/b`'s replaces it
- User navigates to `/a`, `/a` preloads _again_
- User navigates to `/b`, `/b` preloads _again_

It's wasteful to prefetch a link just to throw it away and prefetch it again. I think we should consider a caching mechanism that allowed for multiple pages to be saved and for the invalidation to be configured per project.

A simple expiration in seconds with cleanup on navigation would already be better than what we currently have, imo. Thoughts?

ps. the current prefetching code will also use the latest preloaded page to figure out if it should preload a new one or not, making `goto(currentPage)` a noop if you haven't preloaded anything, but not a noop if you have.

Issue #349, opened 2018-08-07
Preload should be triggered on `goto(<self>)`


mhkeller:
I'll make a proper reproduction repo this evening but for now scaffolding out the issue.

If you make a simple search form like this, `preload` should fire when you navigate using `goto`. It currently does not.

```js
<svelte:head>
  <title>Search</title>
</svelte:head>

<h1>Search</h1>

<form class="search" method="GET">
  <input name="search" type="search" placeholder="" bind:value/>
  <input type="submit" />
</form>

<script>
import { goto } from 'sapper/runtime.js';

export default {
  data () {
    return {
      value: ''
    };
  },
  methods: {
    query (e) {
      e.preventDefault();
      const { value } = this.get();
      goto(`/search?search=${value}`);
    }
  },
  preload ({ query }) {
    console.log('this should appear on form submission. But it does not', query.search);
    // const results = await this.fetch(`search.json`);
  }
};
</script>
```

Rich-Harris:
The reason for this is that it's ignoring everything after `?` — `goto(currentUrl)` *should* be a noop (I would argue), but changing the query string would have to cause a full refresh (including tearing down the root component, in case it uses `query` in its `oncreate` hook...) in order to guarantee the correct behaviour.

Though I suppose we could preserve components that didn't have `oncreate` hooks, and just feed them new data... will have a think on this

mrkishi:
I actually don't believe `goto(currentUrl)` should be a noop -- just like clicking a link to `currentUrl` isn't!

If it makes sense being a noop for a particular use-case, we can always manually implement it on top of `goto`. On the other hand, unless `goto` changes, we can't force it to trigger a page load.

PR #484, opened 2018-10-18
Ensure deepest layout is always refreshed on goto

mrkishi:
See: #349, #483's ps.

This PR just forces the deepest page component to refresh on `goto`, even if that's `goto(currentPage)`. This ensures that a change to a query param triggers a navigation.

One might argue that `goto(currentPage)` should be a noop, but I'd argue otherwise for two reasons:

1. Plain old regular links to the current page aren't noop, and thus users' expectation is that the page will refresh. If you're on hackernews' homepage and click its logo, would you expect it not to refresh revealing the freshest pile of nonsense?

2. `navigate(currentPage)` could be implemented as user code to be noop or not on top of a `goto` that always navigates. The opposite isn't true: if `goto(currentPage)` is a noop, we can't have a userland `navigate` that works differently.

This should probably still be polished (especially if #483 is something we'd like to do), but I think it should work for a temporary fix.

### DIFF
--- a/templates/src/client/app.ts
+++ b/templates/src/client/app.ts
@@ -217,6 +217,10 @@ export function prepare_page(target: Target): Promise<{
 		segments[changed_from] === new_segments[changed_from]
 	) changed_from += 1;
 
+	if (changed_from === new_segments.length) {
+		changed_from -= 1;
+	}
+
 	let redirect: Redirect = null;
 	let error: { statusCode: number, message: Error | string } = null;
 


Rich-Harris:
If the assumption is that `goto(currentPage)` should cause a refresh, don't we need to destroy and re-render intermediate layout components too?

Also, I'm probably being dumb but I'm not sure what you mean in 2) about being able to implement the existing noop behaviour in userland?

mrkishi:
I don't know what the answer to the first question is. The behavior expected by users is for everything to refresh. But that's also true for navigations.

Imagine you have dynamic data on your top-most layout. Say, a number on the root menu. If you are on the homepage and click on the site's logo, would you expect that number to update?

Old behavior says yes, yes you would. But SPAs and PWAs are all about optimizing these away on a case by case. Sapper needs a way for devs to express how deeply it should cache.

---

Re. 2- Given a `goto(page)` implementation that always navigates, one can do:

```javascript
const my_goto = (page) => page === currentPage ? goto(page) : 👎
```

Given a `goto(page)` that does not navigate to `currentPage`, you can't create a `my_goto` that does navigate regardless (at least without editing sapper's goto).

ak5:
Ran into this when trying to force a redraw. Can't we do something like `goto(page, {force: true})`

Rich-Harris:
I think we have the opportunity to fix this more robustly in the Svelte 3 version (see #552), but this makes sense until then — thanks

Rich-Harris:
Released 0.25

Issue #487, opened 2018-10-20
Error thrown from async routes not being handled


nikku:
Starting from the [sapper-template](https://github.com/sveltejs/sapper-template), suppose I change [`blog/index.json.js`](https://github.com/sveltejs/sapper-template/blob/master/src/routes/blog/index.json.js) to the following:

```js
export async function get(req, res) {
  throw new Error('OOPPPS');
}
```

#### Expected Behavior

1. the error is caught, logged properly and indicated via a `500`
2. when rendering the route `/blog` server side a status code `>= 400` triggers the appropriate error message to be rendered
3. when prefetching or rendering the blog client-side the error page indicates a problem

#### Actual Behavior

An unhandled promise rejection error is being logged and the error is silently discarted:

```
(node:14150) UnhandledPromiseRejectionWarning: Unhandled promise rejection
```

Serving the request hangs forever (server-side rendering simply hangs).

#### Environment

`sapper@0.23.3`

nikku:
As mentioned [here](https://github.com/sveltejs/sapper/pull/488#issuecomment-431616703) the issue is not 100% resolved. Should I open another issue suggesting follow up improvements?

Rich-Harris:
Ah, my bad. Yes, if you could that'd be very helpful, thanks

InstanceOfMichael:
for me, this:
```js
{#await promise}
  <Loader />
{:then users }
   [...]
{:catch error}
  <ErrorAlert {error} />
{/await}
```
is throwing:
```shell
(node:27201) UnhandledPromiseRejectionWarning: Error: Request failed with status code 401
    at createError (/home/forge/code/node_modules/axios/lib/core/createError.js:16:15)
    at settle (/home/forge/code/node_modules/axios/lib/core/settle.js:18:12)
    at IncomingMessage.handleStreamEnd (/home/forge/code/node_modules/axios/lib/adapters/http.js:202:11)
    at IncomingMessage.emit (events.js:203:15)
    at endReadableNT (_stream_readable.js:1145:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
(node:27201) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
[...]

PR #488, opened 2018-10-20
handle async route errors

nikku:
Related to #487

### DIFF
--- a/templates/src/server/middleware/get_server_route_handler.ts
+++ b/templates/src/server/middleware/get_server_route_handler.ts
@@ -53,7 +53,12 @@ export function get_server_route_handler(routes: ServerRoute[]) {
 			};
 
 			try {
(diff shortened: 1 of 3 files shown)


nikku:
This improves the situation described in https://github.com/sveltejs/sapper/pull/487, as it makes sure that errors in async routes are actually caught. It only really improves direct access to the `.json` file though.

If the file is consumed via `this.fetch` during preloading it will now complain about fetch returning invalid JSON. It returns the actual error text.

```
FetchError: invalid json response body at http://127.0.0.1:3000/index.json reason: Unexpected token O in JSON at position 0
    at eval (webpack:///./__sapper__/server.js?:810:32)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

Rich-Harris:
nice, thank you!

PR #521, opened 2018-11-27
Update package.lock file with latest npm-run-all with event-stream fix

silentworks:
In relation to this fix.

https://github.com/mysticatea/npm-run-all/issues/149


Conduitry:
Thanks. This changes a bunch of stuff however, and so I just committed that only updates npm-run-all.

PR #531, opened 2018-12-09
Fixed tests for node 6

artemjackson:
Well, after hours of investigation the "prefetches programmatically" test failure I noticed several things:

- the test passes when run in isolation or run in head of sequence
- the test passes when page's `requestInterception` is disabled
- sapper does prefetching (I've checked promise resolution result in [AppRunner.capture](https://github.com/sveltejs/sapper/blob/master/test/apps/AppRunner.ts#L117) and it's ok)

I guess that the real issue is in puppeteer.

So, my solution is to isolate `setRequestInterception` and deal with it only where needed more exactly in "redirects" tests.


Rich-Harris:
thank you!

PR #535, opened 2018-12-16
Do not include sourcemap files in Service Worker shell

### LINKED ISSUE #534: Sourcemap files should not be cached in Service Worker
Using Webpack, sourcemap files are included in the Service Worker `shell` files. This is wasteful because there's no need for a ServiceWorker to cache these; they're only useful for debugging. It increases the size of the Service Worker (doubles the number of JS strings in the list) and also adds to the page weight and disk usage.

nolanlawson:
Fixes #534 

Technically this bug only affects Webpack, not Rollup, but I added tests for Rollup as well. I probably should be able to modify this in the Webpack-specific code rather than `build.ts`, but I wasn't sure how.

In any case, the tests cover both Rollup and Webpack, and I confirmed that the Webpack test fails before this PR but succeeds after.


Rich-Harris:
thank you!

PR #538, opened 2018-12-21
[WIP] Support Svelte 3

Rich-Harris:
This doesn't represent much work — basically the minimum required to get an app working. A particular annoyance is that we can no longer check for the existence of `Component.preload`, because in Svelte 3 things like `preload` are separate named imports:

```html
import Component, { preload } from './Component.html';
```

Rollup doesn't like that — it chokes when you try to import a name that the module doesn't export. We have to jump through some unpleasant hoops to get it to work, and even then it gives us a bunch of warnings. The solution, I think, is to precompile page/layout components so that we can see which ones have `preload`, and add the imports more selectively.

Anyway, I'll release an alpha version so that I can deploy the new (also very WIP) site.


Rich-Harris:
Bringing this back on track... all tests now passing except the ones relating to `Store`

src/core/create_manifest_data.ts
```diff
@@ -9,14 +10,28 @@ export default function create_manifest_data(cwd: string): ManifestData {
 		throw new Error(`As of Sapper 0.21, the routes/ directory should become src/routes/`);
 	}
 
+	function has_preload(file: string) {
+		const source = fs.readFileSync(path.join(cwd, file), 'utf-8');
+
+		if (/preload/.test(source)) {
+			try {
+				const { stats } = svelte.compile(source, { generate: false });
```
Conduitry: I think we should use an empty warning handler here to suppress the display of any warnings while we're determining whether there's a preload. They'll be getting displayed during the main compilation anyway.

Rich-Harris: good catch — added locally

Issue #532, opened 2018-12-11
Sub-pages like [id]/foo.html cannot navigate back to [id].html


nolanlawson:
If you have a page structure like:

    /accounts/[accountId].html
    /accounts/[accountId]/foo.html

Then if you navigate from the first to the second, then press the back button in the browser, Sapper will not actually render the first page. It will just stay stuck on the second page.

I wrote a minimal repro using `sapper-template`. Steps to repro:

```bash
git clone https://github.com/nolanlawson/sapper-template.git \
  --branch demo-sapper-sub-page-bug \
  --depth 1 \
  --single-branch
cd sapper-template
npm i
npm run dev
```

Then in a browser, open `localhost:3000`, click on "Account # 1", then click on "my sub-page", then click the back button.

**Expected result:** it navigates back.

**Actual result:** the URL changes, but it does not navigate back.

nolanlawson:
It seems you can fix this by using `[accountId]/index.html` instead of `[accountId].html`, but it's a bit counter-intuitive that they would work differently.

Conduitry:
The branch is using Sapper 0.22.10, but this looks to also be present in the latest 0.24.1.

cudr:
Have same problem on Edge browser in production mode

Rich-Harris:
This is fixed in 0.25 thanks to @cudr

PR #539, opened 2018-12-25
Fix nested routes navigates

cudr:
This PR repair error with transition between routes, in the case when we use prefetching for nested paths.

The problem occurs when you have several nested links on a page with rel=prefetch. If you move mouse between them and then click on any of these links, there is no transition between the pages.

This is due to the nullable_depth is overwritten for prefetched route

Now we are set new segments at the moment, when navigation takes place between the routes and not at the time of prefetching page.

Reference to this issue #532

### DIFF
--- a/templates/src/client/app.ts
+++ b/templates/src/client/app.ts
@@ -36,7 +36,7 @@ const root_props: RootProps = {
 
 export let prefetching: {
 	href: string;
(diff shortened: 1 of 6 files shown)


Rich-Harris:
Great fix, thank you — released 0.25

PR #541, opened 2018-12-31
support Rollup 1.0

Conduitry:
Support the `.generate` response from both Rollup 0.x and 1.0.

The `inlineDynamicImports` thing is something I'm really not sure about, I _think_ that makes sense, but it's a very big 'think'.


Rich-Harris:
I think you're probably right about `inlineDynamicImports`. I can't imagine it coming up very often anyway

Issue #528, opened 2018-12-05
this.fetch mutates headers passed to it


antony:
Just spent a while hunting this down, but if you keep a reference to a set of `request defaults`, the mutation of `opts.headers` at the end of preload context's `fetch` method can cause a security issue whereby non-authenticated clients end up getting the cookies of an authenticated user's request (And seeing privileged data).

Essentially I think it's best not to mutate things that are passed into a method. There has been some effort not to do this, but headers is always nested in opts and therefore becomes a reference.

I also note that this is a potential security issue when dealing with headers which can store credentials, so I think it's worth fixing.

Line causing the issue is here:

https://github.com/sveltejs/sapper/blob/e69cb3639adc97d61fa878f454f22693093c98f6/templates/src/server/middleware/get_page_handler.ts#L122

Conduitry:
From working through this with @antony, my suggested but untested fix is to replace

```javascript
if (!opts.headers) opts.headers = {};
```

with

```javascript
opts.headers = Object.assign({}, opts.headers);
```

Rich-Harris:
released 0.25 with the fix

PR #542, opened 2018-12-31
avoid mutating `opts.headers` in `this.fetch` (#528)

Conduitry:
Noticed that I never actually submitted a PR for this fix for #528. No tests, because that sounds like more of a pain than I'd like to deal with at the moment.


Conduitry:
I'm not sure what that test failure is. I think it's irrelevant?

Rich-Harris:
yep, kicked Travis and it's all green

PR #559, opened 2019-02-08
Use .svelte extension

Rich-Harris:
See https://github.com/sveltejs/svelte/issues/2067


Rich-Harris:
Components can now have either .html or .svelte extensions. The only breaking change resulting from this is that webpack users must ensure that the `.svelte` extension is enabled both in `resolve.extensions` and the svelte-loader `test` option

PR #561, opened 2019-02-13
Use getElementById to avoid querySelector error

### LINKED ISSUE #516: Initial load error when location hash isn't a selector
I'm using the location hash to store state, in order to make it shareable. 

I get the following error when loading as Sapper is trying to find an element on the page with the id of the hash so that it can scroll to it.

Demo: https://sapper.svelte.technology/guide#asdasd&&

```
client.354c98f5.js:1 Uncaught (in promise) DOMException: Failed to execute 'querySelector' on 'Document': '#asdasd&&' is not a valid selector.
    at X (https://sapper.svelte.technology/client/client.354c98f5.js:1:7630)
    at J.n.then (https://sapper.svelte.technology/client/client.354c98f5.js:1:7116)
```

thgh:
Fix #516

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -158,7 +158,7 @@ export async function navigate(target: Target, id: number, noscroll?: boolean, h
 
 		if (hash) {
 			// scroll is an element id (from a hash), we need to compute y.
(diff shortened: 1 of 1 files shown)


Rich-Harris:
interesting edge case, thanks!

PR #525, opened 2018-12-02
add service-worker-index.html

nolanlawson:
This fixes #422 by outputting a new `/service-worker-index.html` file which can be cached by the service worker and served in offline use cases.

The reason I went with `service-worker-index.html` rather than the old `index.html` is that the latter breaks `sapper export`, since the service worker-targeted `index.html` conflicts with the base route. Since `index.html` is kind of an imprecise name for it anyway, I figured renaming to `service-worker-index.html` made the most sense.

The only change that users will have to make in order to get this to work is to use the commented code from `sapper-template` defined [here](https://github.com/sveltejs/sapper-template/blob/1234971da512d37a6453cfebbf8e47a023f547be/src/service-worker.js#L51-L59). We'll probably want to update that commented code anyway, since it hasn't worked for about a year.

I have tested this with `sapper-template` and it works.

To test, you just `npm link` this PR to `sapper-template`, then in `sapper-template`, do `npm run build && npm start` and then navigate to the base URL, then put the Dev Tools in offline mode, uncomment the commented-out service worker code, then refresh. At this point, the new page should be served by the service worker using `service-worker-index.html`.

PR #563, opened 2019-02-15
Remove unnecessary index.html file serve

nolanlawson:
As far as I can tell, `/index.html` should no longer exist (outside of `sapper export`, which doesn't apply here). After #525 was merged, we're now using `/service-worker-index.html`, but that's handled by `get_page_handler.ts`, not here. So this code is doing nothing.

### DIFF
--- a/runtime/src/server/middleware/index.ts
+++ b/runtime/src/server/middleware/index.ts
@@ -52,11 +52,6 @@ export default function middleware(opts: {
 			next();
 		},
 
-		fs.existsSync(path.join(build_dir, 'index.html')) && serve({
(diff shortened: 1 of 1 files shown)


nolanlawson:
The build failures don't seem related to this PR?

SkaterDad:
> The build failures don't seem related to this PR?

Looking at the CI log, those are the same errors I get when running the tests on a fresh clone.  Makes it a challenge to write a failing test for a PR I have in mind.

Rich-Harris:
thanks! yep, the build failures were related to a Rollup bug that's now fixed

Issue #402, opened 2018-08-31
Add <link rel="preload"> for needed scripts when exporting


Rich-Harris:
Sapper adds preload headers when the server is running, but that doesn't work with `sapper export`

Rich-Harris:
fixed by #568

khrome83:
@Rich-Harris - this appears to be broken now. I am not see this work in the most recent version.

MaxMilton:
@khrome83, how about openning a new issue and adding some info about the regression?

khrome83:
@MaxMilton - I can do that. No real steps to reproduce though. It just does not exist...

PR #568, opened 2019-02-16
fix: add link rel=preload for exported sites

nolanlawson:
I think `sapper export` should try to preserve the `link rel=preload` header as much as possible, which in this case means embedding it into the HTML.

Also I've found that preload makes a big enough difference in load time that it's worth it. (Just tested on Pinafore using simulated slow 3G, and it's a difference of 8.7s for the load event versus 59.9s. :grimacing:)

I added a test using Webpack because it looks like none of the Rollup tests generate preload link headers.


src/api/export.ts
```diff
@@ -139,38 +146,49 @@ async function _export({
 		clearTimeout(the_timeout); // prevent it hanging at the end
 
 		let type = r.headers.get('Content-Type');
+
 		let body = await r.text();
 
 		const range = ~~(r.status / 100);
 
 		if (range === 2) {
```
nolanlawson: Hiding whitespace will make the diff below a lot easier to read.

test/apps/export-webpack/src/client.js
```diff
@@ -0,0 +1,9 @@
+import * as sapper from '@sapper/app';
```
nolanlawson: I copy-pasta'd the `test/apps/export` project into `test/apps/export-webpack`, changing only `rollup.config.js` by replacing it with `webpack.config.js`.

test/apps/export-webpack/test.ts
```diff
@@ -0,0 +1,19 @@
...
+
+	// hooks
+	before(async () => {
+		await api.build({ cwd: __dirname, bundler: 'webpack' });
+		await api.export({ cwd: __dirname, bundler: 'webpack' });
+	});
+
+	it('injects <link rel=preload> tags', () => {
+		const index = fs.readFileSync(`${__dirname}/__sapper__/export/index.html`, 'utf8');
+		assert.ok(/rel=preload/, index);
```
nolanlawson: There is probably more I could do to test this, but I wasn't sure if you wanted me to include an entire HTML parsing library or something. (Cheerio?)

test/apps/export-webpack/test.ts
```diff
@@ -0,0 +1,19 @@
...
+
+	// hooks
+	before(async () => {
+		await api.build({ cwd: __dirname, bundler: 'webpack' });
+		await api.export({ cwd: __dirname, bundler: 'webpack' });
+	});
+
+	it('injects <link rel=preload> tags', () => {
+		const index = fs.readFileSync(`${__dirname}/__sapper__/export/index.html`, 'utf8');
+		assert.ok(/rel=preload/.test(index));
```
nolanlawson: There is probably more I could do to test this, but I wasn't sure if you wanted me to include an entire HTML parsing library or something. (Cheerio?)

test/apps/export-webpack/webpack.config.js
```diff
@@ -0,0 +1,73 @@
+const webpack = require('webpack');
```
nolanlawson: I copy-pasted this config file over from the only other webpack text project, `with-sourcemaps-webpack`.

test/apps/export-webpack/test.ts
Rich-Harris: nah, that'd be overkill — this is what I'd have done as well 👍

Rich-Harris:
Fixes #402. Thanks!

khrome83:
@nolanlawson - this appears to not be working anymore... Opened #919 around this.

Issue #553, opened 2019-02-01
Drop Node 6 support?


Rich-Harris:
Node 6 is still in LTS, but the main reason Sapper still supports Node 6 — that Lambda and Cloud Functions are stuck on that version — no longer holds (both officially support Node 8). Could make things rather simpler. Would that get in anyone's way?

lukeed:
Depends if wanna jump the gun on LTS. Version 6.x expires this April 🎉 I'll be updating my stuff then.

As mentioned, we _only_ added Node 6 because of lambda/function platforms and they're both solved now.

Rich-Harris:
Switched over from TypeScript to Sucrase and realised that it won't build in Node 6 any more anyway... it's annoying that it's awkward to test that something *runs* in an old version of Node without also *building* it in that version, but it is, and I'm lazy, so bye-bye Node 6

PR #569, opened 2019-02-17
update Rollup, remove some superfluous deps

Rich-Harris:
Latest version of Rollup includes a fix that applies to Sapper. While I was at it, removed `rimraf`, `mkdirp` and `sander` dependencies which are total overkill


Rich-Harris:
Sucrase doesn't run in Node 6, so I guess this closes #553 as well

Issue #573, opened 2019-02-20
Use slots for routing


Rich-Harris:
This is a bit of an existential mega-issue for Sapper. Will try and articulate what's in my head as succinctly as possible. (It mostly concerns internal implementation details, and would only entail minor changes to Sapper apps themselves.)

At present, nested routing is handled with `_layout.svelte` components that have to contain something like the following:

```html
<svelte:component this={child.component} {...child.props}/>
```

Upon navigation, Sapper constructs a new tree for the current route, which looks something like this (assuming the user was visiting `/foo/bar`):

```js
const props = {
  child: {
    segment: 'foo',
    component: SomeLayoutComponent,
    props: Object.assign({}, somePreloadedData, {
      child: {
        segment: 'bar',
        component: SomeLeafComponent,
        props: Object.assign({}, someMorePreloadedData, {
          child: {}
        })
      }
    })
  }
};
```

That data is passed to the root component, and the app redraws. If you were to navigate from `/foo/bar` to `/foo/baz`, `props.child.component` and `somePreloadedData` would stay as they are, which is nice, and if `SomeLeafComponent` is used for both `bar` and `baz`, we only need to re-run that component's `preload` function to get new data for the leaf. So in terms of preserving the component tree and minimising the amount of `preload` data-fetching etc, it's quite efficient.

But. Because we're applying that data at the top level, we're still setting `props.child.props` on the layout component, which can cause (avoidable, but surprising) side-effects, due to the component thinking it has received new state.

Then there's the `child` value itself, which is a Sapperism that components have to accommodate (by declaring `export let child`, etc). And it's a bit *weird* that we're using `<svelte:component>` for something that `<slot>` is conceptually better suited to. 

What if we used `<slot>` instead?

## Component-based routing

There are basically three different approaches to routing:

* traditional Express-style routing (things like page.js and Vue router fall into this category)
* component-based routing (React Router)
* filesystem-based routing (Next, Nuxt, Sapper)

The advantage of fs-based routing is that you get code-splitting for free (because the framework does the complicated part for you — it's able to do so because your routes are essentially statically analysable), along with a pretty nice development experience. But if we had component-based routing, we could avoid the inefficiency outlined above:

```html
<!-- the `foo` part -->
<SomeLayoutComponent {...somePreloadedData}>
  <!-- the `bar` or `baz` part -->
  <SomeLeafComponent {...someMorePreloadedData}/>
</SomeLayoutComponent>
```

[...]

PR #574, opened 2019-02-20
Slot-based routing

Rich-Harris:
ref #573. Mostly done, but a bunch of tests that aren't passing yet


Rich-Harris:
Been dogfooding this at work all day and I do prefer it — much less paranoid about state change false positives causing unwanted side-effects. That said, I kind of miss being able to pass props from a layout component to its children, so that `/user/alice/friends` and `/user/alice/enemies` can both share an object representing Alice.

It's definitely solvable with the context API, but not *quite* as elegant. One possible solution would be to have a way to expose all slot props:

```diff
-<svelte:component this={level1.component} segment={segments[1]} {...level1.props}>
+<svelte:component this={level1.component} segment={segments[1]} {...level1.props} let:*={props}>
  {#if level2}
-    <svelte:component this={level2.component} {...level2.props}/>
+    <svelte:component this={level2.component} {...props} {...level2.props}/>
  {/if}
</svelte:component>
```

Can cross that bridge another time though.

Issue #558, opened 2019-02-08
querystring change


Ti-webdev:
Hi
I have simple route:
```html
<a href="?category=filament&weight=300g">filament 300g</a>
<a href="?category=filament&weight=1kg">filament 1kg</a>
<pre>{ JSON.stringify(items, 0, 2) }</pre>

<script context="module">
export async function preload ({ query }) {
  console.log('preload')
  const res = await this.fetch(`/items.json?${new URLSearchParams(query)}`)

  if (res.status === 200) {
    return { items: await res.json() }
  } else {
    this.error(res.status)
  }
}
</script>

<script>
	export let items
</script>
```
For first open everything is fine.
When I click to any of link preload is not called but querystring in URL is changed.
How I should do that?

SkaterDad:
I'm seeing this as well.  Here's a Codesandbox repro (see the blog section): https://codesandbox.io/s/6y330krxmz

Stepping through the code, I think this line is the problem:
https://github.com/sveltejs/sapper/blob/82e637ea7c7295cc4f84a3c4cc33c00ce281efb0/runtime/src/app/app.ts#L250
```js
// This evaluates True when the query param was the only change.
if (!session_dirty && current_branch[i] && current_branch[i].segment === segment) return current_branch[i];
```

antonheryanto:
hit similar bug. it happen after #538 merge

SkaterDad:
Working on a  PR for this.

Should query param changes trigger preload on all nested segments, or only the last one?

My own use case only needs the query string for the final segment.

thgh:
Sort of related: I wish a store could be preloaded. Deriving from $page would then allow to implement something similar to the preload behavior.

```js
  export async function preload() {
    await $items
  }
  const items = derive(stores.page, page =>
    fetch(`/items.json?${new URLSearchParams(page.query)}`).then(r =>
      res.status === 200 ? r.json() : []
    )
  )
```

Or fancier:
```js
preload: items = derive(stores.page, page =>
  fetch(`/items.json?${new URLSearchParams(page.query)}`).then(r =>
    res.status === 200 ? r.json() : []
  )
)
```

Conduitry:
Huh #701 was apparently a duplicate of this and no one noticed. It was fixed in #704 and released in 0.27.1.

PR #577, opened 2019-02-23
Failing test for #558 - preload on query param change

SkaterDad:
Here is a failing test for Issue #558.

As stated in that issue, there is a line of code in `runtime/App.ts` which skips the preload stuff when it detects the route segments haven't changed.  This prevents a query parameter change from triggering the route's preload function.

```js
function hydrate_target(...) {
 //.....
    if (!session_dirty && current_branch[i] && current_branch[i].segment === segment) return current_branch[i];
```

The query param issue can be fixed by removing this line, but then you get a different test failing about un-necessarily recreating components:
```
 1) layout
       only recreates components when necessary:
```

Another option could be detecting when only the query string changed in the link click handler, and calling the preload function(s) directly, since the dynamic imports are already done?

### DIFF
--- a/test/apps/preloading/src/routes/query.svelte
+++ b/test/apps/preloading/src/routes/query.svelte
@@ -1,0 +1,13 @@
+<script context="module">
+	export function preload({query}) {
+		return { text: query.text };
+	}
+</script>
+
+<script>
+export let text
+</script>
+
+<h1>{text}</h1>
+
+<a href="query?text=Second">Second</a>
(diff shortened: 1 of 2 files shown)


SkaterDad:
I've spent a bit more time debugging this problem, and am curious about the router design decisions.

That layout test I referenced in the previous post appears to require that page `preload` functions shouldn't be called if the URL path didn't change.

My expected behavior when clicking a link, even if it's to the same URL I'm on, is that the page will "refresh".  In SSR sites, the browser will load that URL again.  In a SPA, I would want the data fetching to be performed again.

I won't be able to spend much more time on this, but I hope the test is helpful!

Sapper is a project I'm aspiring to be able to use on my projects going forward ⭐️

Conduitry:
This has been fixed in #704.

Issue #339, opened 2018-08-03
Preload function in no nested routes after error page/layout is not called


mkv27:
If you go from "/byevdom" (error 404) to "/about" route preload is not called.

https://glitch.com/edit/#!/sveltejs-sapper-template-1
Running: https://sveltejs-sapper-template-1.glitch.me/byevdom

s3ththompson:
i'm running into the same issue as well. I wonder if it has anything to do with #304 ?

mkv27:
Probably because even in the new route(after the error page) bindings does not work.

Panya:
I guess that this code:
https://github.com/sveltejs/sapper/blob/master/templates/src/server/middleware/get_page_handler.ts#L151-L157

Is incorrect if we navigate to 404 page (that doesn't have `preload` method) than preloaded fo the page will be `{}` (empty object) and then in the client-side code we check:
https://github.com/sveltejs/sapper/blob/master/templates/src/client/app.ts#L261-L268
`initial_data.preloaded[i + 1]` will be the empty object. As a temporary workaround we can define the `preload` method in `routes/_error.html` component and return `undefined`.

nsivertsen:
Can confirm that this bug exists in Sapper 0.25.0, and that @Panya's workaround fixes it for the time being.

Panya:
The bug exists in 0.26.0-alpha.9 (with Svelte v3) too. Can be fixed if we change https://github.com/sveltejs/sapper/blob/master/runtime/src/server/middleware/get_page_handler.ts#L168 `{}` to `undefined`. Or rewrite the condition to:
```js
if (part.preload) {
    return part.preload.call(preload_context, {
        path: req.path,
	query: req.query,
	params
    }, session);
}
```

Conduitry:
Should be fixed by #620.

PR #578, opened 2019-02-25
Add failing test for #339 and possible fix

Panya:
Not sure about the fix, but all tests are green.


Panya:
Already fixed by #620

PR #585, opened 2019-03-01
Fix export race condition (bonus windows support for dev)

halfnelson:
Export was firing off up to 8 requests, but only waiting for the last one to complete before killing the server.js process. This uses the q.close() method that seems to be designed to wait for all queued promises to complete.


Conduitry:
Rich actually recently removed rmraf and added a simple standalone version of it - https://github.com/sveltejs/sapper/blob/master/src/api/utils/fs_utils.ts - The build script should probably be changed to use this, rather than bringing rimraf back in again.

PR #591, opened 2019-03-03
Fix redirect with basepath

### LINKED ISSUE #589: redirect() is not correct working with base path
I can reproduce this bug in Opera 58, Chrome 33, Chrome 74 Dev. And only if I go the link from address bar. Click by link is worked correctly.

I set base path as `/sapper/page` - site worked correctly. 
Then added this code to the `routes/index.html`
```
export default {
    preload() {
        return this.redirect(302, 'login')
    }
}
```
Actual redirect behavior:
`http://localhost:3000/sapper/page ==> http://localhost:3000/sapper/login`
but I expected this:
`http://localhost:3000/sapper/page ==> http://localhost:3000/sapper/page/login`

---

I can fix this bug if I change follow line in the file `__sapper__/server.js`:
```
if (redirect) {
    console.log(req.baseUrl);
    // const location = url.resolve(req.baseUrl || '/', redirect.location); // Original
    const location = url.resolve((req.baseUrl || '') + '/', redirect.location); // Changed
    res.statusCode = redirect.statusCode;
    res.setHeader('Location', location);
    res.end();
    return;
}
```

thgh:
Fixes #589

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -177,7 +177,7 @@ export function get_page_handler(
 
 		try {
 			if (redirect) {
-				const location = URL.resolve(req.baseUrl || '/', redirect.location);
+				const location = URL.resolve((req.baseUrl || '') + '/', redirect.location);
 
 				res.statusCode = redirect.statusCode;
 				res.setHeader('Location', location);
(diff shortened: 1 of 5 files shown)


Rich-Harris:
d'oh! thank you

Issue #595, opened 2019-03-07
Rediscover the lost arts of 0.26.0-alpha.10


Conduitry:
`sapper@0.26.0-alpha.10` seems to have some code changes that were never checked into git, and which were lost in alpha 11. It's motherfucking archaeology time.

Conduitry:
Running a recursive diff on the contents of the two versions of the package (and renaming a couple `chunk` files that correspond with each other so that diff will compare them) results in:

```diff
diff -r alpha9/dist/api.js alpha10/dist/api.js
21c21
< var dev_ts = require('./chunk-f292f4ed.js');
---
> var dev_ts = require('./chunk-85bf4152.js');
24c24
< var build_ts = require('./chunk-fce48595.js');
---
> var build_ts = require('./chunk-4f956fda.js');
diff -r alpha9/dist/chunk-11k.js alpha10/dist/chunk-11k.js
191c191
< 					} catch (err) {
---
> 					} catch (error) {
193c193,194
< 							message: err.message
---
> 							type: 'manifest',
> 							error
400,401c401,402
< 		compiler.watch((err, result) => {
< 			if (err) {
---
> 		compiler.watch((error, result) => {
> 			if (error) {
404c405
< 					message: err.message
---
> 					error
504c505
< //# sourceMappingURL=chunk-f292f4ed.js.map
---
> //# sourceMappingURL=chunk-85bf4152.js.map
diff -r alpha9/dist/chunk-4k.js alpha10/dist/chunk-4k.js
139c139
< //# sourceMappingURL=chunk-fce48595.js.map
---
> //# sourceMappingURL=chunk-4f956fda.js.map
diff -r alpha9/dist/cli.js alpha10/dist/cli.js
379c379
< var version = "0.26.0-alpha.9";
---
> var version = "0.26.0-alpha.10";
420c420
< 		const { dev } = await Promise.resolve(require('./chunk-f292f4ed.js'));
---
> 		const { dev } = await Promise.resolve(require('./chunk-85bf4152.js'));
464,465c464,473
< 				console.log(__chunk_2.colors.red(`✗ ${event.type}`));
< 				console.log(__chunk_2.colors.red(event.message));
---
> 				const { type, error } = event;
> 
> 				console.log(__chunk_2.colors.bold().red(`✗ ${type}`));
> 
> 				if (error.loc && error.loc.file) {
[...]

PR #596, opened 2019-03-07
various lost changes and fixes from 0.26.0-alpha.10

Conduitry:
Ref #595.

None of these changes look particularly shocking, but it would be a good idea to go over these with an extra discerning eye,

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -199,9 +199,10 @@ class Watcher extends EventEmitter {
 						});
 
 						manifest_data = new_manifest_data;
-					} catch (err) {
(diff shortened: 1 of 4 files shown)


Rich-Harris:
this all looks familiar and correct. still no idea where this stuff went — perhaps a byproduct of me working on two computers and not remembering to sync changes between them

Conduitry:
Glblah well I just found some of the changes, they were still on the `better-errors` branch. I'm glad they appear to be identical to the ones I reconstructed. I still don't know where the `app.mjs` -> `client.js` fix came from or went though.

PR #597, opened 2019-03-08
Fix preloads in SSR

arxpoetica:


Conduitry:
This currently breaks because the changes here are incompatible with the bundled beta3 of Svelte. (#598)

Conduitry:
Closing this in favor of #599 which includes the necessary upgrade to svelte (and also stops bundling it).

Issue #598, opened 2019-03-08
Don't bundle Svelte compiler


Conduitry:
At least the latest alpha of Sapper is bundling the Svelte compiler (version 3.0.0-beta.3). This avoids issues with changes between compiler versions, but I don't think this is a good idea overall.

Conduitry:
During the pre-release phase of Svelte and Sapper, I think it's reasonable to not be particularly careful about breaking changes or compatibility between different versions of the two projects. I don't foresee more changes in this area of Svelte, but if there are, people will just have to upgrade together.

Conduitry:
It looks like the intention is to not bundle Svelte, but in `rollup.config.js`, we are constructing the `external` array from (in part) `Object.keys(pkg.dependencies)`, which will contain `'svelte'` but not the `'svelte/compiler'` that Svelte 3 now uses.

Rich-Harris:
blurgh. yeah, sounds like we need to update that Rollup config — definitely don't want to bundle the compiler

Issue #594, opened 2019-03-05
(Rollup v3) `preload` not getting called consistently on server, but always runs on client


arxpoetica:
**Update** it's consistent, and it has to do w/ style preprocessing.

Unfortunately, I've had a very difficult time pinning this down, but it appears that preload sometimes doesn't get called *on the server*. It's totally inconsistent, and I can't reproduce it on anything reliable, but it seems to be happening when I switch to different pages with different preloads (and refreshing).

Navigating to new pages that are not server hits (but client only) is just fine and works great. It's the server that sometimes doesn't call `preload`.

Sorry this is less helpful. I feel pretty blind as to what the problem is. It'll start working and I'll have no clue, and then the next moment it won't.

Potentially related bugs: #554, #415, #589, #586, #588, #587, and #555.

neoel:
Is this with svelte.v3 & sapper alpha? I seem to be having the same issues there at least. although it is very consistently not being called on the server. 

Put a [sample video online](https://youtu.be/26t0jFx4iBw).
Here's a [REPL link](https://www.youtube.com/redirect?q=https%3A%2F%2Fv3.svelte.technology%2Frepl%3Fversion%3D3.0.0-beta.10%26gist%3Da99cdc5a270f3f5acac20e083d079dae&v=26t0jFx4iBw&event=video_description&redir_token=op9l4sMg-tFbxwBpHxN6Cj-ip9x8MTU1MjA1NDIzMUAxNTUxOTY3ODMx)

arxpoetica:
@neoel are you using preprocessors anywhere?

neoel:
@arxpoetica Yeah, I am using svelte-preprocess. For loading scss styles. I've tried only passing the styles prop in the preprocess object . 
```js
// ...
preprocess: { style: preprocess().style }
}
```

But that didn't seem to influence it.

neoel:
Ah, I've now completely removed it. And it does not trigger the bug. So svelte-preprocess has something to do with it then, any clue what changes cause this. (I'm not familiar with the innards here)

arxpoetica:
I created a repro: https://github.com/arxpoetica/gah

Conduitry:
[...]

PR #599, opened 2019-03-08
Server-side preload check fixes

### LINKED ISSUE #575: Update for Svelte beta 4
There may be other things that need to happen, but at least when compiling to check on the presence of `preload`, we need to not pass `onwarn` and we need to check for `vars` at the top level

Conduitry:
Should fix #575 and #594 and #598.


Conduitry:
It might be good to have a unit test for the preload/preprocessing thing, but my brain is done for the evening.

Rich-Harris:
the hero gotham needs, just not the one it deserves

arxpoetica:
A favorite film.

PR #600, opened 2019-03-08
Construct absolute base href using request data.

superafroman:
Some browsers (Internet Explorer - sorry :( ) have issues when the base href isn't an absolute URI. It's also defined as being absolute by W3C - https://www.w3.org/TR/html4/struct/links.html#edef-BASE

This PR is a suggestion for how it could be updated - keen to discuss.


superafroman:
A little more info on this, Mozilla specify it can be relative (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) so debatable if it's worth resolving here. You will also introduce an issue if the site is being served from different domains, where the first request will set the domain of the base URI for all requests. (template processing being done only on first load in production)

In addition, having an absolute base URI doesn't fix all issues in IE. Sapper uses `document.baseURI` for `goto` and `prefetch` which doesn't get populated in IE < 11. 

So a polyfill is still needed. I'll leave this PR here for anyone else to feed in to if they like.

arve0:
> Sapper uses document.baseURI for goto and prefetch which doesn't get populated in IE <= 11.

Adding: `<base>` also affects URLs for `<a>` anchor links and `` images in browsers not supporting "semi-absolute" `href`, not only the javascript APIs. Using absolute `href` overcomes this, which makes it possible to create apps that works on older browsers (assuming no dependency to running JS in the client).

Stumbled upon this when doing a funny side project, taking *progressive enhancement* to the extreme; create an app that works in both IE5.01 and modern browsers.

I'll admit the silliness of using browsers released in 1999 and do not expect sapper to support it, though it works! I find that intriguing, considering the 20-year age gap!

benmccann:

benmccann:

PR #605, opened 2019-03-12
Fixes #604 - using single queue for export.

### LINKED ISSUE #604: Export uses a queue for every handle(url), so scaling in exponential concurrent promises
Problem is that the `queue` is per `handle(url)`:
https://github.com/sveltejs/sapper/blob/7be7e1eb9f8c17e8d8b6b6966cb67fe7b54bc9ab/src/api/export.ts#L165-L187

See https://github.com/bwbroersma/exponential-export-queue for the issue.

I changed the yootils queue behavior https://github.com/Rich-Harris/yootils/issues/3 and have a simple sapper patch.

bwbroersma:


Rich-Harris:
Thanks — I made some further changes in 1707fe8e9d3c964b292d22f6adb9d15625c3b810. Basically I think we should keep the semantics of `q.close` as they are, and just ensure that `handle` doesn't resolve until all its pages have been fetched

Rich-Harris:
...though I just realised that fails when you have a very large number of pages, since it calls itself recursively 🙈. Working on a fix

bwbroersma:
Yes, that's the problem why we probably need a change or add functionality to yootils.
See https://github.com/bwbroersma/exponential-export-queue for a nice export test (although this example will never end, it will show the exponential concurrency ;)

Rich-Harris:
Figured it out: we just need to make sure that the call to `handle` isn't the thing that's queued. Instead, it's the call to `fetch` that can only happen 8 at a time — #641

bwbroersma:
Yes, I already found it a bit weird and fixed it on the other/wrong end, but this is the neat approach. Thanks!

Issue #545, opened 2019-01-06
[...rest] route params


Rich-Harris:
This has come up a few times but I just realised we don't have an issue for it (that I can find).

It'd be cool if there were a way to match paths of arbitrary depth. Happily, there's a really nice idiomatic way we can achieve that, by using rest syntax:

```
src/routes/foo/[...bar]/baz.html
```

This would match `/foo/a/b/c/baz`, and `params.bar` would equal `['a', 'b', 'c']`.

Rich-Harris:
added in 0.26

PR #607, opened 2019-03-13
[...spread] routes

cudr:
implement #545


cudr:
Important notice: changes in the comparator function only work on node 11 and above, see https://github.com/nodejs/node/issues/24294

Rich-Harris:
Thanks @cudr, this is great news — though I'm not sure I understand the comment about the comparator function. Are you saying Sapper's behaviour would differ between Node 10 and 11 with this PR?

cudr:
@Rich-Harris Yes, there was some uncertainty due to which the comparator function behaved differently between 10 and 11. Now more tests are added and it works the same way on both 10 and 11 versions of the nodejs

Rich-Harris:
awesome, thanks

Issue #587, opened 2019-03-02
`preload` without `return {}` causes error


imbolc:
I'm using `preload` to only fill session without any return object and after the last changes it starts to cause an error: `TypeError: Cannot convert undefined or null to object`

njbotkin:
Got this too.  Thanks for figuring out `return {}`.  I only ran into this on a component that was 2 levels deep, if that means anything for the debugger.

artemjackson:
@imbolc which side did you catch the error? From server or client?

Rich-Harris:
Fixed in 0.26 — closing

PR #608, opened 2019-03-13
Don't crash if preload return empty

cudr:
fix for #587

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -225,7 +225,7 @@ export function get_page_handler(
 
 					props[`level${l++}`] = {
 						component: part.component,
-						props: preloaded[i + 1],
+						props: preloaded[i + 1] || {},
 						segment: segments[i]
 					};
 				}
--- a/test/apps/preloading/src/routes/preload-nothing/index.svelte
+++ b/test/apps/preloading/src/routes/preload-nothing/index.svelte
@@ -1,0 +1,5 @@
+<script context="module">
+	export function preload() {}
+</script>
+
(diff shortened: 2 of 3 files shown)


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -225,7 +225,7 @@ export function get_page_handler(
 
 					props[`level${l++}`] = {
 						component: part.component,
-						props: preloaded[i + 1],
+						props: preloaded[i + 1] || {},
```
artemjackson: I'm not sure but it seems like `null`, `0` and `""` are normal values for return from `preload` which are cut-off by this line

artemjackson: May be comparison with `undefined` may be more suitable

thgh: `preload` must always return an Object.

@cudr This might also error:
https://github.com/sveltejs/sapper/blob/6bf3dd04dd011db328e746bd3b7a3793f149ff20/runtime/src/server/middleware/get_page_handler.ts#L212

artemjackson: Still can't understand why preload **must** return an object. From what I see the props are [just spread to components](https://github.com/sveltejs/sapper/blob/6bf3dd04dd011db328e746bd3b7a3793f149ff20/src/core/create_app.ts#L278). It shouldn't be error-prone 🤔

From where is [the error](https://github.com/sveltejs/sapper/issues/587) originated from?

Rich-Harris:
Awesome, I ran into this the other day. Thanks

Issue #450, opened 2018-09-25
No async data on root index.html after getting error


antoineneff:
If the root index.html file (`routes/index.html`) has some async data in its preload method (returning a promise or using async/await), the page won't reload after getting an error from `this.error`, the error layout is still showing up.

I'm getting 3 errors in the browser console when trying going back to `/` route :
```
TypeError: "each_value is undefined"
TypeError: this.proxyTarget is undefined (x2)
```

To reproduce, you can copy the `routes/blog/index.html` from sapper template to `routes/index.html`, get an error navigating to an invalid route and trying to go back on home from navigation. I did that on that repo : https://github.com/antoineneff/sapper-error-repro

Feel free to ask if you need further information.

sgametrio:
Same bug here.
I'll link [here](https://github.com/sveltejs/sapper-template/issues/94) issue I opened on sapper-template too

Conduitry:
Should be fixed by #620.

Issue #304, opened 2018-07-02
window bindings undefined on error pages


johnmuhl:
repo with `sapper-template` https://github.com/johnmuhl/glowing-barnacle. Visit e.g. `http://localhost:3000/wrong` to see the `undefined` values.

Conduitry:
Should be fixed by #620.

PR #620, opened 2019-04-06
Error page lifecycle

cudr:
In current alpha, no JS seems to be loaded on error page. Fixed.

Pick #304, looks like #339 works too.


Conduitry:
Probably also fixes #450 and https://github.com/sveltejs/sapper-template/issues/94

Rich-Harris:
Brilliant, thank you!

PR #622, opened 2019-04-08
Fixed %sapper.styles% injection for webpack apps

artemjackson:
Hi again!

Several month ago I've [added](https://github.com/sveltejs/sapper/pull/529) `css` field to `build.json` info file for webpack apps.
The `css` field had a shape like:
```typescript
{
  main: string,
  chunks: Record<WebpackChunkName, CSSFileName>
}
```

That time I didn't notice two things:

1. Page part styles [are retrieved](https://github.com/sveltejs/sapper/blob/master/runtime/src/server/middleware/get_page_handler.ts#L283) via `build_info.css[part.file]`. Because of the implementation described above used the `wepackChunkName` to form `ccs.chunks` keys the styles were never added to `%sapper.styles%`.
2. The `css.chunks` values should be arrays not strings

The PR fixes both of the issues described.

### DIFF
--- a/src/core/create_compilers/WebpackResult.ts
+++ b/src/core/create_compilers/WebpackResult.ts
@@ -1,6 +1,6 @@
 import format_messages from 'webpack-format-messages';
 import { CompileResult, BuildInfo, CompileError, Chunk, CssFile } from './interfaces';
-import { ManifestData, Dirs } from '../../interfaces';
+import { ManifestData, Dirs, PageComponent } from '../../interfaces';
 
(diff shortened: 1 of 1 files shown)


artemjackson:
Btw, I tested the PR with sapper@0.24.3.

Will it be possible to use the fix with sapper version lower the 0.26-alpha ?

Rich-Harris:
Sorry for the delay, Sapper has been a big neglected during the Svelte 3 release. I also published a 0.25.1 version with just this fix. Thanks!

PR #628, opened 2019-04-13
Allow to have middleware for the path same with a HTML page

akihikodaki:
The feature was introduced with 9e2d0a7fbca10034d9ed83df03200fd549174f73, but regressed with commit 8dc52a04e4bb3e358834442c6125a1989fa2ea4e.


Rich-Harris:
Thank you — added a test in 0862d0e2c8380ded9493244a9fbf249ef41b1294 to guard against further regressions

PR #632, opened 2019-04-25
Add support for custom route file extensions.

### LINKED ISSUE #219: Sapper shell --ext command for alternate file types
For example, including pug files.

Example:

```
sapper dev --ext html,pug
sapper build --ext html,pug
```

pngwn:
Adds an `--ext` option to the sapper cli to allow users to define custom file extensions for route components.

I don't know if this is how you would like to achieve this, the Sapper codebase is pretty alien to me, but this seems to work. It still defaults to the old `.svelte` and `.html` extensions when no argument is passed.

I added a new folder `test/app/custom-extensions` with some tests because I needed a custom config to check that arbitrary file extensions worked. I also added some unit tests (or tried) but they all seem to be broken (and switched off) and I don't know enough about what the manifest is supposed to look like to be confident of fixing them. But there is something resembling a unit test for custom extensions in there too.

I've also tested this locally with a sapper-template and it doesn't seem to break any of the builds, with or without an amusing extension (I don't really know how to test dev/export with the AppRunner thing).

I think this would close #219. Not sure if there are any other issues relating to this.


pngwn:
I have updated the tests to match the new test API, fixed the broken unit tests and added an extra unit tests for server routes. Not sure if it was necessary but I did it anyway.

thgh:
It might have been easier to create a rollup/webpack plugin that resolves .pug or other extensions to .svelte

Conduitry:
A bundler plugin wouldn't help - When Sapper is examining the `src/routes` directory to create a manifest of routes, this is before the bundler is even running. I don't know about webpack, but I don't think renaming a file as part of a plugin is something that can happen anyway - https://github.com/rollup/rollup/issues/530

Rich-Harris:
I think we should probably support Svelte markdown files by default, if we can just settle on the right extension. Releasing this will allow us to experiment at least. Thanks @pngwn

PR #636, opened 2019-04-27
update to node stable

Rich-Harris:
For some reason, the `export-webpack` test fails in Node 12. I get the same result running `sapper export` on a webpack project; it exports fine, but then doesn't exit. Works fine with Rollup, works fine with Node 11. Not sure what's happening — for now I've just pinned the Travis config to use Node 11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 language: node_js
 
 node_js:
-  - "11"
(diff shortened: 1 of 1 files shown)


Rich-Harris:
fixed by b624edf47c23d8b525475d9d2dbc7ce2c7b42e4c

PR #638, opened 2019-04-28
start adding site to main repo

Rich-Harris:
Just the homepage for now, about to start on `/guide`. Copy/tagline etc all open to bikeshedding

Screen Shot 2019-04-28 at 14 37 57


Rich-Harris:
Ok I've deployed this in its current form to http://sapper-staging.surge.sh/, if anyone is interested

Conduitry:
Visiting http://sapper-staging.surge.sh/migrating directly gives me the page's json. Do we also want this to have its own nav link at the top right of the page?

Rich-Harris:
what the... that must be some Surge-specific behaviour. Going to http://sapper-staging.surge.sh/migrating/ takes you to the proper page. It won't ultimately be deployed on Surge, it's just the quickest way to throw it up for now.

I wondered about the nav. It feels like clutter for something that relatively few people will need though

site/content/migrating/01-migrating.md
```diff
@@ -0,0 +1,367 @@
...
+Until we reach version 1.0, there may be occasional changes to the project structure Sapper expects.
+
+
+### 0.25 to 0.26
+
+The most significant change yet: Sapper is now built on Svelte 3.
+
+#### Importing Sapper
+
+Your app is now built to `src/node_modules/@sapper` — this allows you to easily import it from anywhere in your source code. Update your `server.js`...
```
Conduitry: I dunno if I'd say that your _app_ is built to there - your app is still built to `__sapper__/dev` and `__sapper__/build` - It's stuff that you want to import that's built to `src/node_modules/@sapper` before your app's bundled. I don't know how important of a thing to be super precise about this is.

Rich-Harris: Gah, you're right. Honestly I was thinking about turning this whole page into something more like the docs page on svelte.dev and introducing some kind of a tutorial, since this page is kind of trying to do what the old `/guide` did.

site/content/migrating/01-migrating.md
```diff
@@ -0,0 +1,367 @@
...
+It is also available, along with `page` and `preloading`, as a store inside components:
+
+```html
+<script>
+	import * as sapper from '@sapper/app';
+	const { page, preloading, session } = sapper.stores();
+</script>
+```
+
+`page` and `preloading` are [readable stores](https://svelte.dev/tutorial/readable-stores), while `session` is [writable](https://svelte.dev/tutorial/writable-stores).
```
Conduitry: Are there caveats around where or when you can write to the `session` store and have it be saved and propagated to other places? That might be nice to mention here.

site/content/migrating/01-migrating.md
Conduitry: This _is_ the migration page so I don't think a more digestible or lengthy version of this is really necessary - but it might be nice to have separate tutorial and docs pages sometime in the future. Although I have no idea what implementing a Sapper REPL would involve, and we shouldn't hold up other things for that.

Conduitry:
"The next small thing in web development" still doesn't gel with me the way "Cybernetically enhanced web apps" does. I have no productive contributions to make in this area though :(

site/content/docs/01-structure.md
```diff
@@ -0,0 +1,112 @@
...
+
+#### src/client.js
+
+This *must* import, and call, the `start` function from the generated `@sapper/app` module:
+
+```js
+import * as sapper from '@sapper/app';
+
+sapper.start({
+	target: document.querySelector('#sapper')
```
Conduitry: Are you planning on getting #645 into 0.26.0?

Rich-Harris: nah, i think it needs some testing first. can be done relatively unobtrusively later

site/content/docs/01-structure.md
```diff
@@ -0,0 +1,112 @@
...
+* it should serve the contents of the `static` folder, using for example [sirv](https://github.com/lukeed/sirv)
+* it should call `app.use(sapper.middleware())` at the end, where `sapper` is imported from `@sapper/server`
+* it must listen on `process.env.PORT`
+
+Beyond that, you can write the server however you like.
+
+
+#### src/service-worker.js
+
+Service workers act as proxy servers that give you fine-grained control over how to respond to network requests. For example, when the browser requests `/goats.jpg`, the service worker can respond with a file it previously cached, or it can pass the request on to the server, or it could even respond with something completely different, such as a picture of llamas.
```
Conduitry: I don't know whether I was supposed to think about goatse when I read goats.jpg but I did

site/content/docs/02-routing.md
```diff
@@ -0,0 +1,127 @@
...
+	if (article !== null) {
+		res.setHeader('Content-Type', 'application/json');
+		res.end(JSON.stringify(article));
+	} else {
+		next();
+	}
+}
+```
+
+> `delete` is a reserved word in JavaScript. To handle DELETE requests, export a function called `del` instead.
```
Conduitry: The docs for this is absolutely the wrong place to be posting this, but: In line with Svelte 'export-as' thing, might we want to do the same thing here rather than treating a 'del' export specially?

site/content/migrating/01-migrating.md
Rich-Harris: Good point, have added a line

site/content/migrating/01-migrating.md
Rich-Harris: Tweaked it slightly — still not perfect, but happy enough with it for now

site/content/docs/01-structure.md
Rich-Harris: The young 'uns will never know the horrors we internet old-timers have seen

site/content/docs/02-routing.md
Rich-Harris: I think it's probably easier just to export `del` — either way, you need to know about `delete` being reserved, so export-as only adds extra code:

```js
export function del() {...}

// vs
function del() {...}
export { del as delete }
```

Rich-Harris:
Deployed to https://sapper.svelte.dev, using Netlify (first time using it, was impressed). Not 100% finished — a couple of broken links, and copy/tagline are still bikesheddable, but I think we're very nearly ready to release 0.26

lukeed:
👍 Netlify is definitely a strong contender for best static-site hosting.

Conduitry:
Interesting. Someone (presumably Netlify?) is redirecting stuff like https://sapper.svelte.dev/docs to https://sapper.svelte.dev/docs/ - Probably not a problem?

lukeed:
Netlify does this with "Pretty Urls" option: https://www.netlify.com/docs/redirects/#trailing-slash

PR #651, opened 2019-05-01
always use latest shimport

Rich-Harris:
I think this is what we want?


Conduitry:
Shimport is completely under your control, so you can say now that there will never be any breaking API changes. But even so, it does seem better to plan for them. Perhaps the current Shimport should be released as 1.0.0 at this point? And then you can just point to ^1.0.0.

Rich-Harris:
You're right — have made that change

PR #652, opened 2019-05-02
switch license to MIT

Rich-Harris:
I was hoping the LIL license would catch on, but it clearly hasn't... I'm pretty sure switching to MIT is legally sound since they're essentially identical (by design), but if any contributors have objections, please make them known


Conduitry:
Oh no Travis doesn't like the new license

PR #662, opened 2019-05-05
Fix mobile viewport KO

thollander:
- Changed `;` to `,` which causes the bug
- Current view on mobile is : 
image
- Changes to : 
image


mindrones:
related to https://github.com/sveltejs/svelte/issues/2551#issuecomment-487278633

Rich-Harris:
Thank you — redeploying. I've updated sapper-template to have the same attribute as well.

PR #664, opened 2019-05-05
Tidy a few things up

mrkishi:
Note:
- Removes the `req[IGNORE]` trick
- Speeds up `compose_handlers` by avoiding an extra closure and moving `ignore` test outside request handlers


Rich-Harris:
my kind of diff

PR #665, opened 2019-05-06
Switch to node stable

mrkishi:
Switches travis back to node stable by updating Svelte to non-alpha v3 along with all other dependencies.


Rich-Harris:
I have no idea how this fixes the weird hanging issue but... I'll take it

mrkishi:
At first I thought updating webpack alone would fix it since it was having some issues with v12, but it turns out it was the svelte alpha itself that wouldn't exit when exporting.

Issue #601, opened 2019-03-08
sapper export - can export update only one route, not all the contents of site?


NadiaRom:
In case of blog or news website we need to **update** collections of dynamic routes, but not to render each item on update as `sapper export` command does .

Is there a way to **export just one route**, not the whole site? similar to https://nuxtjs.org/api/nuxt-render-route.

The ideal workflow is:
1) server receives a webhook - for example, a new article is added
2) and triggers rendering of `routes/articles/fresh-article-slug`, then adds it to thousands of previously generated articles as `__sapper__/export/articles/fresh-article-slug` without affecting the rest.

Can I do it with Sapper?

As of now, the only solution I see is to have a temporary export destination, where to put only the updated article. Next dully copy the exported updated one to its fellows in a real export folder. Definetly this solution is terrible. How to do a proper update of static website (if possible)?

thgh:
I think Sapper currently uses timestamps in filenames. That makes incremental builds challenging.

NadiaRom:
Thank you! I see... After looking at Gatsby and Gridsome it is surprising that so few static site generators support incremental builds. 

I would appreciate any comments on the idea to copy&past one exported item from "temporary" export folder to a "general" export folder. May it work around the problem?

s0kil:
This is a great idea to consider, especially for larger projects that take a while to export.

khrome83:
Where do timestamps come in during the build?

I feel like this would be as easy as `sapper export --entry /url/to/updated/route --no-crawl` with the idea only that entry point would be crawled, but the code to go to multiple links would not run. 

Maybe I am over simplifying this?

It appears that during export hashes are no longer added based on #670 - is that right @Conduitry .

khrome83:
@NadiaRom @s0kil - we added #856 to allow building a single page. We will be opening a PR soon for filtering routes as well. This PR will only do the entry point. The next one will do routes that match a regex condition allow you to split your builds.

jthegedus:
Related:
- [Next.js 9.3 SSG](https://nextjs.org/blog/next-9-3#next-gen-static-site-generation-ssg-support)
- [Next.js incremental SSG discussion](https://github.com/zeit/next.js/discussions/11552)
- [Gatsby incremental builds](https://www.gatsbyjs.org/blog/2020-04-22-announcing-incremental-builds/)

davidwparker:
@Conduitry - I'm wondering if there's any updates on this?

[...]

PR #670, opened 2019-05-06
Build rollup config: Don't add hashes to files in dist

cristianl:
I tried this on a lark. No difference in passing tests.

If there's a technical reason for delivering the Sapper dist hashed, feel free to close this immediately.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -52,7 +52,8 @@ export default [
 		output: {
 			dir: 'dist',
 			format: 'cjs',
-			sourcemap: true
+			sourcemap: true,
+			chunkFileNames: '[name].js'
 		},
 		external,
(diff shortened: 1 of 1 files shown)


Conduitry:
The hashes are for cachebusting, especially when serving through Cloudflare or something else that caches JS and CSS files aggressively. Maybe we could make this configurable though, as part of facilitating something like incremental builds (#601)?

Conduitry:
Oh, this is for removing hashes from files in the built version of _Sapper_, not in the built version of a site made with Sapper. I don't have strong opinions about this then.

Rich-Harris:
Yeah, this is mildly nicer to look at and could make it slightly easier to look at stack traces. Thanks

PR #676, opened 2019-05-08
Consolidate index handling for page and server routes

### LINKED ISSUE #624: `index.js` in root of `routes` creates incorrect server route mapping
I'm trying to redirect the `/` route in a sapper project, by creating a server route as `index.js` in the root of `/routes`.

The structure of this would look like:

```
src/
├── routes/
|       └── index.js
...
```

Upon doing so, `/` returns a 404 and the generated route in `server.js` is:
```js
    server_routes: [
        {
            // index.js
            pattern: /^\/index$/,
            handlers: __route_index,
            params: () => ({})
        }
    ]
```

I would expect the generated route pattern to be `/^\/$/` instead of `/^\/index$/`.

mrkishi:
Fixes #624 (sorry for the spam there! I didn't realize force-pushing a commit did that).

Treats both page and server routes similarly in regards to index resolution.

It's a (somewhat) breaking change: server routes would never accept trailing slashes before, and pages always would. Since server indexes are functional with this, I also made them accept trailing slashes — as long as there's no extension in the route (since that seems to be a common use case for server routes).

We still need a stronger support for configuring clean URLs rules. Blindly responding to both `/page` and `/page/` has negative SEO impacts, and right now sapper forces us to handle redirects on an earlier middleware.

Note: this PR is on top of my earlier refactor PR because it'd have conflicts, but I can change this if there's interest in this and not the other.

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -101,7 +101,11 @@ export function select_target(url: URL): Target {
 	if (url.origin !== location.origin) return null;
 	if (!url.pathname.startsWith(initial_data.baseUrl)) return null;
 
-	const path = url.pathname.slice(initial_data.baseUrl.length);
+	let path = url.pathname.slice(initial_data.baseUrl.length);
+
+	if (path === '') {
+		path = '/';
+	}
 
 	// avoid accidental clashes between server routes and page routes
 	if (ignore.some(pattern => pattern.test(path))) return;
--- a/runtime/src/server/constants.js
+++ b/runtime/src/server/constants.js
@@ -1,1 +1,0 @@
-export const IGNORE = '__SAPPER__IGNORE__';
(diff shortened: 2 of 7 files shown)


Rich-Harris:
> We still need a stronger support for configuring clean URLs rules

Do you have strong opinions about what that should look like? i.e. should we redirect `/page` to `/page/` or vice versa?

mrkishi:
No strong opinions on the right redirect, but a big one that whatever we choose should be configurable.

The app might be using some `/basepath` in an environment that does one or another, or it might sit behind a forward proxy (and we wouldn't want to require it to be especially configured to avoid redirect loops!).

I'm also unsure if it's worth supporting a way to make specific routes go the other direction. Someone might want a specific route to have the slash so relative URLs work, even though the rest of the app uses no trailing slash (or the other way around).

Rich-Harris:
Man, sounds complicated. No cheap trick we can do with `<link rel="canonical">` instead? 😬

PR #677, opened 2019-05-08
Abort infinite loop for preload errors

mrkishi:
Really not ideal, but just something to avoid hanging the server on errors for now.

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -26,6 +26,15 @@ export function get_page_handler(
 	const { server_routes, pages } = manifest;
 	const error_route = manifest.error;
 
+	function bail(req: Req, res: Res, err: Error) {
+		console.error(err);
+
+		const message = dev ? escape_html(err.message) : 'Internal server error';
+
+		res.statusCode = 500;
+		res.end(`<pre>${message}</pre>`);
+	}
+
 	function handle_error(req: Req, res: Res, statusCode: number, error: Error | string) {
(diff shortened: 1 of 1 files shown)


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -171,6 +171,13 @@ export function get_page_handler(
 
 			preloaded = await Promise.all(toPreload);
 		} catch (err) {
+			if (error) {
+				// early bail for preload error
+				res.statusCode = 500;
+				res.end(`<pre>${escape_html(err.message)}</pre>`);
```
Conduitry: I agree we don't want to hang in an infinite loop here, but we need to be careful about exposing any details about what the exception was if we're in production.

mrkishi: True!

To be honest I just copied the other loop bail without thinking. Should I just make them both `Internal server error` for now?

Conduitry: I think that's be good, yeah. I didn't realize there was another error that would already be printing out exception detail in the response.

It might also be good to have something like `console.log(err.stack || err.message);` so that more detailed information will turn up in logs. I'm not sure what the standard behavior is around this type of logging though, and this is starting to get a bit away from the original purpose of the PR.

mrkishi: Alright, updated to do that. We can probably get the main logic a bit of love and bail at a single point, but for now let's just not hang node :P

Issue #660, opened 2019-05-05
Incorporate sirv


Rich-Harris:
Given that the `static` folder has special meaning (e.g. it's used to determine what should go in the service worker manifest) it probably makes sense just to include `sirv` inside Sapper, rather than having it in the project template separately.

That way, the simplest possible Sapper server could look roughly like this:

```js
import * as http from 'http';
import * as sapper from '@sapper/server';

http.createServer(sapper.middleware()).listen(process.env.PORT);
```

ansarizafar:
Its a great idea but we should be able to use other middlewares.
````
import * as http from 'http';
import * as sapper from '@sapper/server';
import bodyParser  from 'body-parser';

sapper.use(bodyParser.json())

http.createServer(sapper.middleware()).listen(process.env.PORT);
````

ansarizafar:
We can also try to replace http with https://github.com/uNetworking/uWebSockets.js which is much faster than node http.

Issue #645, opened 2019-04-30
Make `target` automatic


Rich-Harris:
The `<div id="sapper">` here is slightly cumbersome...

```html
<body>
  <div id="sapper">%sapper.html%</div>
  %sapper.scripts%
</body>
```

...especially paired with this in `client.js`:

```js
sapper.start({
  target: document.querySelector('#sapper')
});
```

It'd be nicer if the HTML sent from the server included something like `<sapper-anchor>`, so that we could just do this...

```html
<body>
  %sapper.html%
  %sapper.scripts%
</body>
```

...and this:

```js
sapper.start();
```

Internally Sapper would do something like

```js
const anchor = document.querySelector('sapper-anchor');
app = new App({
  target: anchor.parentNode,
  anchor,
  // ...
});

anchor.parentNode.removeChild(anchor);
```

Rich-Harris:
Just realised that for hydration, the target element can't have anything other than the app HTML, so it would need to be this...

```html
<body>
  <sapper-app>
    <!-- SSR'd HTML goes here -->
    <div id="app">...</div>
  </sapper-app>
  <script>...</script>
</body>
```

...rather than this:

```html
<body>
  <!-- SSR'd HTML goes here -->
  <div id="app">...</div>
  <sapper-anchor/>
  <script>...</script>
</body>
```

That's a sensible thing to do (since at the moment there are no guarantees that the target element doesn't contain additional stuff), but it could potentially affect layout, so it needs some testing. So I won't try and sneak this in to 0.26.

antony:
Solved in Kit

PR #682, opened 2019-05-10
Make Sapper server more self-contained

Rich-Harris:
This PR may not be totally uncontroversial, so I'd welcome feedback on it.

In #660 I proposed incorporating `sirv` directly into Sapper — we're already replicating some of its functionality (to serve client assets), and the `static` folder is 'special' to a Sapper project, so it kind of makes sense. And because it's a @lukeed project, it's small enough that it's basically free.

Right now, the template `src/server.js` looks like this:

```js
import sirv from 'sirv';
import polka from 'polka';
import compression from 'compression';

const { PORT, NODE_ENV } = process.env;
const dev = NODE_ENV === 'development';

polka() // You can also use Express
	.use(
		sirv('static', { dev }),
		sapper.middleware()
	)
	.listen(PORT, err => {
	});
```

Arguably, `compression` doesn't belong at this layer (it should be handled by your CDN), so we could simplify that to this..

```diff
-import sirv from 'sirv';
import polka from 'polka';
-import compression from 'compression';
import * as sapper from '@sapper/server';

-const { PORT, NODE_ENV } = process.env;
-const dev = NODE_ENV === 'development';
+const { PORT } = process.env;

polka() // You can also use Express
	.use(
-		compression({ threshold: 0 }),
-		sirv('static', { dev }),
		sapper.middleware()
	)
	.listen(PORT, err => {
		if (err) console.log('error', err);
	});
```

But then there's really no need for Polka — we can just do this:

```diff
-import polka from 'polka';
+import * as http from 'http';
import * as sapper from '@sapper/server';

const { PORT } = process.env;

-polka() // You can also use Express
-	.use(
-		sapper.middleware()
-	)
+http.createServer(sapper.middleware())
	.listen(PORT, err => {
		if (err) console.log('error', err);
	});
```

We can make that even nicer by exporting a `start` function:

```js

sapper.start();
```

(This PR also adds a `sapper.server(opts)` function, which just creates the server but doesn't listen on `process.env.PORT`.)

This also gives us a nice symmetry with `src/client.js` (especially with #645). **The trade-off is** that the barrier to understanding how to add your own custom non-Sapper routes is marginally higher, which does run counter to Sapper's 'show the guts' philosophy.

Not shown in this PR — a `sapper.render(...)` or `sapper.handle(...)` function. I think we should add that, probably in a separate PR, I'm just not 100% sure what the API should be yet.

Thoughts?


runtime/src/server/middleware/index.ts
```diff
@@ -1,23 +1,42 @@
...
+
+		sirv(`${build_dir}/client`, {
+			dev,
+			maxAge: 31536000,
+			immutable: true
+		}),
+
+		sirv(`${build_dir}/service-worker`, {
+			dev,
+			maxAge: 300
```
lukeed: Did you mean to change the SW settings? Was no-cache, must-revalidate, etc

ansarizafar:
I would suggest sapper.use() for using middlewares like helmet, bodyparser etc.

runtime/src/server/middleware/index.ts
Rich-Harris: huh, I guess not. Not sure what I was thinking. How would I add those with sirv?

lukeed: It's custom settings, so has to be thru `setHeaders` option.

lukeed:
Perhaps I missed it, but I think sirv needs to be moved to dependencies list

mrkishi:
I like it!

Do you mean `sapper.handle()` would expose sapper as a middleware? Because while `sapper.start()` is awesome for the main api, removing that option seems too closed-warden. If you meant it as a way to add custom routes, I think at this point `sapper.start()` could just return a sapper server with an usual routing api (`get`, `post`, etc), since it's basically acting as a server framework.

I'd really love to go for the direction where sapper's server provides a better internal api, and just have the main interface (`sapper.start`) use said internal api. We could then expose the internal api for advanced use-cases. I haven't had the chance to give it proper thought, but the holy grail would be granular enough to be able to expose treeshakeable rendering at the route level from `@sapper/server` (think lambda per route, routing info available to be converted to whatever lambda provider you use).

runtime/src/server/middleware/index.ts
```diff
@@ -1,23 +1,42 @@
...
+		}),
+
+		function condition_request(req: Req, res: Res, next: () => void) {
+			const qi = req.url.indexOf('?');
+			req.query = ~qi ? querystring.parse(req.url.slice(qi + 1)) : {}
+
 			if (req.baseUrl === undefined) {
-				let { originalUrl } = req;
+				let originalUrl = req.originalUrl || req.url;
```
benmccann, line 39: This change is also being requested in https://github.com/sveltejs/sapper/pull/942

PR #683, opened 2019-05-10
Allow disabling live reload

cristianl:
The sapper CLI has a `--live` flag for live reload that is enabled by default and currently cannot be disabled.

With this PR, `sapper dev --live=false` will not send reload events to the browser.


Rich-Harris:
whoops! bit of an oversight — thanks

Issue #689, opened 2019-05-13
sapper export: "UnhandledPromiseRejectionWarning: Error: Cannot add to a closed queue"


btakita:
This is due to not waiting for all requests to finish in a promise.

See https://github.com/sveltejs/sapper/blob/master/src/api/export.ts#L176

halfnelson:
This was clobbered by https://github.com/sveltejs/sapper/pull/641
the older code use `q.add` which allowed q.close to do the promise.all.

kmelve:
We're hit by this as well, making it impossible to deploy [our starter] on Netlify as a static site. It can be reproduced by deploying it [through sanity.io/create](https://www.sanity.io/create/?template=sanity-io%2Fsanity-template-sapper-blog).

It seems that async stuff in `prefetch` is the culprit? E.g https://github.com/sanity-io/sanity-template-sapper-blog/blob/master/template/web/src/routes/blog/%5Bslug%5D.svelte

btakita:
@kmelve I manually go into the sapper source & [modify](https://github.com/btakita/sapper/commit/a1648345c84b8bccdc92f534b07cff317a3564cd) the `_export` function to make the export work for now. Hopefully this will be fixed soon.

btakita:
I created a temporary version that fixes this issue & https://github.com/sveltejs/sapper/issues/688

Use the following in your `package.json`

`"sapper": "btakita/sapper#0.27.0-fixes"`

btakita:
I created a temporary version that fixes this issue & #688

Use the following in your package.json

"sapper": "btakita/sapper#0.27.0-hotfix"

kmelve:
Excellent @btakita – tnx 🙌

antony:
@kmelve I also believe there is an issue in the default configuration for the deployed app in your sapper starter - even with this hotfix the deploy will fail.

To deploy correctly you need to change the deploy directory from `__sapper__/export` to `web/__sapper__/export`.

PR #690, opened 2019-05-13
Waiting for all requests to finish before closing q.

btakita:
Fixes issues/689

`await Promise.all(promises)`

### DIFF
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -173,7 +173,7 @@ async function _export({
 					let match;
 					let pattern = /<a ([\s\S]+?)>/gm;
 
-					let promise;
+					const promises = [];
 
 					while (match = pattern.exec(cleaned)) {
 						const attrs = match[1];
@@ -183,12 +183,12 @@ async function _export({
 							const url = resolve(base.href, href);
 
 							if (url.protocol === protocol && url.host === host) {
-								promise = handle(url);
+								promises.push(handle(url));
 							}
 						}
 					}
 
-					await promise;
+					await Promise.all(promises);
 				}
 			}
 		}
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -173,7 +173,7 @@ async function _export({
 					let match;
 					let pattern = /<a ([\s\S]+?)>/gm;
 
-					const promises = [];
+					const urls = [];
 
 					while (match = pattern.exec(cleaned)) {
 						const attrs = match[1];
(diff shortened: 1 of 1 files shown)


mrkishi:
How does this relate to #604 and #641?

It looks good at first glance, but does it reintroduce #604's problem?

btakita:
Should we introduce rate limiting?

btakita:
https://github.com/sveltejs/sapper/pull/690/commits/d3fc0c1eae312a6a05464a175b9126ff8bdc9e61 adds rate limiting based on the `concurrent` option. The default is 8.

btakita:
I took another look & realized I was missing a big part of the picture.

Here's how I see the situation. This PR fixes https://github.com/sveltejs/sapper/issues/689 which is caused by `handle` not waiting for all requests to finish.

The `await Promise.all`simply waits for all of the requests to finish. There is some rate limiting by chunking calls to `handle`, however this does not guarantee the limit to be `concurrent` due to this solution not being a proper queue.

`q.add` should be the solution, however removing `q.add` was the fix for https://github.com/sveltejs/sapper/issues/604 & https://github.com/sveltejs/sapper/issues/641.

Does this mean there's a bug with the queue? If so, then the bug with queue should be fixed, since the queue is needed to guarantee a concurrent request limit of `concurrent`.

btakita:
@mrkishi To answer your question, this PR has no effect on https://github.com/sveltejs/sapper/issues/604 or https://github.com/sveltejs/sapper/issues/641.

The only way this would reintroduce https://github.com/sveltejs/sapper/issues/604 is if the `UnhandledPromiseRejectionWarning: Error: Cannot add to a closed queue` "fixes" the issue by limiting the number of requests.

mrkishi:
Hey, @btakita. It's not that I disagree that there's a problem, it's just that the root of the issue theoretically _shouldn't_ be awaiting those promises, since `handle` is supposed to use `queue` and thus guaranteed to finish sequentially.

I tried another fix that shouldn't ignore promises in case of timeouts at #698. Could you test that version to verify if it solves your problem? I don't have a codebase where #689 happens and I'm not sure what's required to create a test (if you do, we should add a test!).

mrkishi:
Err, I seem to have misunderstood everything. Indeed we should be waiting for all `handle`s since the `q` is only happening for `fetch`. In any case, the `q` should enclose the whole request or else timeouts would break.

btakita:
I tested your branch with the `dist` & `runtime` files `"sapper": "btakita/sapper#gh-689"` & the issue is fixed.

btakita:
I agree that the queue should be used to manage all of the tasks. Well done!

Rich-Harris:
Closing in favour of #698 — will cut a release soon

Issue #688, opened 2019-05-13
Navigating within the same spread route changes url but page stays the same


btakita:
I will follow up with a test repo.

For now, I isolated the issue to https://github.com/sveltejs/sapper/blob/master/runtime/src/app/app.ts#L306

btakita:
Here is the repro branch https://github.com/btakita/business/tree/sapper/issues/688

To repro the issue:

```shell
git clone https://github.com/btakita/business.git
cd business
git checkout sapper/issues/688
yarn
cd packages/web
yarn run dev
```

* Open localhost:3000
* Click on one of the top navigation links
* Click on one of the *.md links in the body
* The url will change but the page will not load

btakita:
In the branch above, the issue occurs with a spread route in a directory `/g/[...a1__segment]`

If the spread route is in the root directory, the issue does not occur.

btakita:
I have another repro case at https://github.com/ctx-core/sapper-template/tree/sapper/issues/688

This one is a bit easier.

```shell
npx degit ctx-core/sapper-template#sapper/issues/688 my-app
cd my-app
yarn
yarn run dev
```

* open localhost:3000
* click about
* click 001-about.md link

PR #691, opened 2019-05-13
compare sliced route match to determine if segment_dirty is true

btakita:
Fixes https://github.com/sveltejs/sapper/issues/688

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -292,11 +292,22 @@ export async function hydrate_target(target: Target): Promise<{
 	let l = 1;
 
 	try {
+		const { pattern } = route;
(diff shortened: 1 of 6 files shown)


btakita:
The PR seems to be working with additional tests which failed with the old implementation.

The problem was the `hydrate_target` was comparing the split path to the route segment, which resulted in the cached `current_branch[i]` erroneously being returned for spread routes.

The bug manifested as the `preload` on the spread route not being called due to the cached `current_branch[i]` being returned.

This fix checks the sliced route match array against the sliced route match array in `current_branch[i]`.

btakita:
I force pushed some formatting fixes.

Rich-Harris:
nice fix, thank you

PR #692, opened 2019-05-13
leave anchor scrolling on initial load to browser

### LINKED ISSUE #331: Sapper treats anchor tag clicks as route changes
Here's a repo reproducing the issue, which I'm currently only seeing in Firefox: https://github.com/mhkeller/anchor-test

When I go to a page that has an anchor tag like `localhost:3000/#link-5`, it will initially load at the correct page height but then jump to the top. In Chrome and Safari, the page stays where it's supposed to.

screenshot

Conduitry:
Fixes #331 hopefully without breaking anything else. This just sets noscroll to true during the initialization of the client-side app, so that wherever the browser decided to scroll to because of anchor fragments will be left untouched.

I don't really know what to test for this.

### DIFF
--- a/runtime/src/app/start/index.ts
+++ b/runtime/src/app/start/index.ts
@@ -40,7 +40,7 @@ export default function start(opts: {
 		if (initial_data.error) return handle_error(url);
 
 		const target = select_target(url);
-		if (target) return navigate(target, uid, false, hash);
+		if (target) return navigate(target, uid, true, hash);
 	});
 }
 


Conduitry:
It might be nice to backport this to 0.25 for the benefit of people on Svelte 2. Rollupjs.org springs to mind.

antony:
@Conduitry it would be amazing if this got backported. Sadly I'm stuck on v2 for the time being, and my stakeholders can't understand why something as "simple" as anchor links don't work!

Conduitry:
For the past couple days this change was locally made for deployments to svelte.dev and I haven't heard about any adverse affects. I think this should be good to go?

Rich-Harris:
I think that's the right call, yeah. Fingers crossed this deals with it once and for all

PR #694, opened 2019-05-15
Fix restarting server in dev

jarrodldavis:
Wait for the server process to exit before starting a new one, so that the server debugger port is free when the new process begins.


Rich-Harris:
nice, thank you

Issue #707, opened 2019-05-21
Regexp URLs broken


IlyaSemenov:
Per <https://sapper.svelte.dev/docs#Regexes_in_routes>:

> For example, src/routes/items/[id([0-9]+)].svelte would only match numeric IDs — /items/123 would match, but /items/xyz would not.

This is not true, /items/xyz still matches.

## Steps to reproduce

1. `npx degit sveltejs/sapper-template#webpack my-app`
2. Create `src/routes/[id([0-9]+)].svelte` with `<h1>hello</h1>`
3. Run `yarn && yarn dev`
4. Open <http://localhost:3000/xyz>

## Expected result

Error 404

## Actual result

The page shows 'Hello'.

jcollum:
```
$ tree src/routes/
src/routes/
├── _error.svelte
├── _layout.svelte
├── about.svelte
├── data
│   ├── [year([0-9]+)].svelte
```

Same here, going to `http://localhost:3000/data/xyz` renders the page instead of a 404.

Conduitry:
Fixed in 0.27.2, thanks!

jcollum:
Confirmed, thank you!

PR #721, opened 2019-06-02
Repair regexp routes

cudr:
After implement spread routes, looks like regexp routes was broken. Fixed. Added more tests for that case. Related #707


Conduitry:
Released in 0.27.2, thank you!

PR #723, opened 2019-06-04
fix: cli ignores --output option

Ashot-KR:
CLI ignores `--output` option and always uses default `src/node_modules/@sapper`.
PR fixes that.


Conduitry:
Released in 0.27.2, thank you!

PR #732, opened 2019-06-09
Fix resolution

Rich-Harris:
Svelte 3.5.0 broke Sapper, because of the changed folder structure. This fix — which I'm not proud of — replaces the `svelte/store.mjs` import with `svelte/store/index.mjs` if the version of Svelte is new enough.

The tests are failing locally because webpack is resolving imports incorrectly, in a way that I can't seem to reproduce in a standalone (non-Sapper) project. We could always release this fix for Rollup and then come back to webpack, it it takes a while

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1544,15 +1544,6 @@
           "resolved": "https://registry.npmjs.org/ms/-/ms-2.0.0.tgz",
           "integrity": "sha1-VgiurfwAvmwpAd9fmGF4jeDVl8g=",
           "dev": true
-        },
-        "resolve": {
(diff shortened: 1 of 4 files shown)


Conduitry:
I'm going to close my eyes and click and if I hit the approve button that's not really my fault is it

Conduitry:
With the change in Svelte 3.5.1, which parts of this are actually needed now? Can we not rely on bundler resolution and just import from `svelte/foo` instead of `svelte/foo.mjs`/`svelte/foo/index.mjs`?

Rich-Harris:
good catch

Issue #735, opened 2019-06-10
Proposal: Provide hostname to preload and page store


Conduitry:
It has come up a few times that people want to have slightly different pages served depending on the domain that a Sapper server is accessed via. The best advice is probably [this](https://github.com/sveltejs/sapper/issues/533#issuecomment-447061184) which is definitely hacky.

Thinking about this again, this doesn't seem that hard to provide to people though. We could just add a `hostname` key to the `page` object that `preload` is called with (which would be `req.domains.host` or `window.location.hostname` for server and client respectively), and we could also add a `hostname` to the `page` store value that's computed in `select_target(url)`.

Is there some reason to _not_ provide these things? Some extra client-side bytes that won't be used the majority of the time?

Rich-Harris:
Honestly, I think the main reason this wasn't part of the original design was that I had no idea it was possible. But apparently `req.headers.host` will give you the hostname the site is being accessed at; the `Host` header is required by HTTP 1.1.

Adding extra bytes is always something to be watchful for. But it's not many bytes. If this has utility then it's probably worth adding (though I confess I'm not yet clear on what that utility is).

Maybe just `host` rather than `hostname`? It contains more information. `req.headers.host` should be equivalent to `window.location.host`, if I understand correctly.

Conduitry:
Yep, calling it `host` rather than `hostname` and getting it on the client from `location.host` makes more sense, as `req.header.host` and `location.host` will both include the port number (if it's something non-standard) and `location.hostname` will not.

Only lightly tested, but here is what I have so far:

```diff
diff --git a/runtime/src/app/app.ts b/runtime/src/app/app.ts
index 8e15c0b..091e9fe 100644
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -121,7 +121,7 @@ export function select_target(url: URL): Target {
 			const part = route.parts[route.parts.length - 1];
 			const params = part.params ? part.params(match) : {};

-			const page = { path, query, params };
+			const page = { host: location.host, path, query, params };

 			return { href: url.href, route, match, page };
 		}
@@ -129,7 +129,7 @@ export function select_target(url: URL): Target {
 }

 export function handle_error(url: URL) {
-	const { pathname, search } = location;
+	const { host, pathname, search } = location;
 	const { session, preloaded, status, error } = initial_data;

 	if (!root_preloaded) {
@@ -154,7 +154,7 @@ export function handle_error(url: URL) {

 	}
[...]

PR #741, opened 2019-06-11
Add support for accessing host in preload and in page store

Conduitry:
Implements #735.

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -121,7 +121,7 @@ export function select_target(url: URL): Target {
 			const part = route.parts[route.parts.length - 1];
 			const params = part.params ? part.params(match) : {};
 
-			const page = { path, query, params };
+			const page = { host: location.host, path, query, params };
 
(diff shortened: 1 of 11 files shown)


talltyler:
I'm hosting a Sapper app in a firebase function so the host name is returned as the name of the functions host not as the domain that is visible to users which is available from req.headers['x-forwarded-host']. I could patch this for my app but I wondering why Sapper doesn't give access to all of the request headers by default. Shouldn't the preload function have raw access to the request headers? Things like UserAgent are often useful but many other custom things are managed this way. Maybe I'm missing something but if not let me know what you all would prefer and I'll try to put together a PR.

antony:
@talltyler It could be a useful feature, but please open a new issue to discuss it rather than necro-ing a year old PR

Conduitry:
Preload functions should not have access to all request headers because this would, for example, give client side code access to http-only cookies. If you need access to specific headers, you can expose them via the session - https://sapper.svelte.dev/docs#Seeding_session_data

PR #743, opened 2019-06-12
Support async method in session middleware option

### LINKED ISSUE #740: session option in sapper.middleware does not support Promise
I want to return promise in session option of sapper.middleware. But if I got an error saying 
`Failed to serialize session data: Cannot stringify arbitrary non-POJOs`. It seems that the session option does not support promise. Is there any workaround?

```javascript
polka()
  .use(
    sapper.middleware({
      session: async (req, res) => {
        const obj = await get_some_obj();
        return obj;
      }
    })
  )
```

hitochan777:
Fixes #740

You can now run async operation in session option.

```javascript
const app = polka()
  .use(
    sapper.middleware({
      session: async (req, res) => {
        return await async_operation();
      }
    })
);
```

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -11,7 +11,7 @@ import App from '@sapper/internal/App.svelte';
 
 export function get_page_handler(
 	manifest: Manifest,
-	session_getter: (req: Req, res: Res) => any
+	session_getter: (req: Req, res: Res) => Promise<any>
 ) {
 	const get_build_info = dev
 		? () => JSON.parse(fs.readFileSync(path.join(build_dir, 'build.json'), 'utf-8'))
@@ -88,7 +88,7 @@ export function get_page_handler(
 			res.setHeader('Link', link);
 		}
 
-		const session = session_getter(req, res);
+		const session = await session_getter(req, res);
 
 		let redirect: { statusCode: number, location: string };
 		let preload_error: { statusCode: number, message: Error | string };
--- a/test/apps/session/src/server.js
+++ b/test/apps/session/src/server.js
@@ -11,7 +11,7 @@ const app = polka()
 	})
 	.use(
 		sapper.middleware({
-			session: (req, res) => ({
+			session: async (req, res) => await Promise.resolve({
 				title: `${req.hello} ${res.locals.name}`
 			})
 		})


hitochan777:
I have one question: how to type-check? It seems that during the build process it just does transpilation without any typecheck.

ilyaztsv:
Also need that feature. One of cases when I need that: send request to API in order to get data and save the data in session (the request params based on `req` info).

Conduitry:
Making an API call as part of populating the session data is the precise reason I think we need to be careful with adding this feature. I am pretty sure that this would result in the server making the API call for every single request. We don't want to make it easy to fall into this sort of trap.

If the server needs to make an API call as part of creating the session data, this should happen once, as part of the appropriate login/whatever endpoint. This gets saved, once, with whatever session middleware you're using. Then the Sapper session can be (synchronously) populated from that.

hitochan777:
In my case I wanted to use `async/await` because a method provided by firebase-auth to verify session cookie is an async operation. I don't think it is a good idea to verify the cookie only once for security robustness.
https://firebase.google.com/docs/auth/admin/manage-cookies#verify_session_cookie_and_check_permissions

hitochan777:
@Conduitry what's your thought on my comment above?

ramaisen:
Please add this feature!, everything in js is async why would this not be the case?

Conduitry:
If you're proxying the API requests over to another server, and you're handling that via some middleware, you could just put the middleware to set `req.currentUser` after that (but still before the Sapper middleware).

Using middleware to set `req.currentUser` only seems inconvenient if you are using Sapper server routes to proxy to some external server, and that isn't too desirable anyway.

fdidron:
@Conduitry @ximus I had pretty much the same need and ended up running my async stuff in an async middleware executed right before sapper's .
The only thing you need to watch out for is that this middleware also gets hit by requests for static assets (js, css, images), so make sure you filter those .

Rich-Harris:
Thank you, and sorry for the long wait! Released 0.27.15

Rich-Harris:
(To respond to the earlier conversation — this *does* potentially introduce a way to make page requests slow, but in practice I think it's more likely to alleviate the problems arising from the workarounds: 1) you can fix it with a prior middleware, but then you're left with the filtering problem or you'll end up doing that work for *all* requests, not just pages, and 2) you can do the work in `preload`, but then you create unnecessary extra work in client-side navigation — and if it's data that belongs in `session`, then you'd likely have to duplicate that work in multiple `preload` functions)

benmccann:
@hitochan777 @Rich-Harris this PR broke the build  when it was merged

PR #766, opened 2019-06-25
Fixes #751. Support setting root context in server.js and client.js

### LINKED ISSUE #751: Feature request: Preload argument for context
## Problem

I've been trying to adopt Sapper the past couple weeks, and it's been mostly good. One topic seems to keep biting me though, and that has to do with "initializing" each request in various ways so that I can use my helpers in the client or on the server. For example, here are a couple things I wanted to do:

- setting up a shared ApolloClient instance for use both during `preload` and normal component lifecycle.
    - I can't use `context` to pass it because it's not available during preload.
    - I can't use `session` to pass it because the server-side version would be serialized to the client.
    - On the client I _could_ just have a singleton in a JS module. But I can't do that on the server because then every request would share the same ApolloClient (bad news). I also need access to the `req` object on the server so I can forward cookies.
- setting up honeycomb.io integration to send events when there is an error
    - on the client, I serialize events and hit a server route which calls honeycomb-beeline for each one (honecomb-beeline is their official client)
    - on the server, the events are created directly against honeycomb-beeline

## Workaround

I have so far been abusing `_layout.svelte`'s `preload` to ensure something happens on the server, and all of my helpers have a `process.browser` check to change behavior (and change which things are `require()`'d) in browser and server. I have abused `session` to share a property between all the `preload`s and all the ssr renders - in `_layout.svelte`'s preload I do `session.apolloClient = ....` and in `_layout.svelte`'s render I do `setContext(key, $session.apolloClient); $session.apolloClient = null;`. This is pretty ugly, but it's the best I've got at the moment.

## Solution?

I'd like to help implement something in Sapper to deal with this, if you have an guidance on what you want to see or whether this interests you.

My thinking was that a third parameter could be passed to `preload` which essentially acts like Svelte context - it doesn't persist between client and server, but is available during preload. E.g.

```js
<script context="module">
    export async function preload(page, session, context) {
        // set up some context both for preloads and for components
        // not the best place to do this though, see below for where I would actually do it.
        context.set('apolloClient', createClient(this.fetch)); 

        // a preload lower down the tree could do
        return {
            data: await context.get('apolloClient').query(....)
        };
    }
}
</script>
<script>
    getContext('apolloClient') // would only work during the server-side render, unless `preload` was called client-side
</script>
```

This third parameter could be initialized similarly to `session` with server/browser-specific details, which avoids lots of `process.browser` switches throughout my code. Something like:
 (client.js)
```js
sapper.start({
    target: ...,
    context: () => [
        ['apolloClient', createClient()],
        ['mixpanel', window.mixpanel],
        ['honeycomb', honeycombProxy]
    ]
})
```
(server.js)
```js
sapper.middleware({
    session: (req, res) => { ... }
    context: (req, res) => [
        // use cookies from the request when talking to API server
        ['apolloClient', createClient(req) ]
        // associate any events with the user on this request
        ['mixpanel', {
            track: (event, props) => require('mixpanel').track(event, {
                distinct_id: req.user.id,
                ...props
            })
        ],
        // honeycomb does magic to hook into express at module load, so no explicit user link is necessary
        ['honeycomb', require('honeycomb-beeline')() ]
    ]
})
```

~Note I used an array-of-arrays to initialize because Context keys can be both objects and strings (so neither a Map nor an object literal are sufficient).~ EDIT: Just realized I was thinking of WeakMaps which can't accept string. A `Map` is what's actually used, so returning a `Map` from this function would be just fine! Perhaps some other API is better.

Something like the above would solve my problem of overusing `process.browser`, and would let me pull helper logic out of `_layout.svelte` that doesn't really belong there. Most importantly I could stop abusing session.

I'm keen to hear what you think and again - I'm happy to put in work for this if you think it has legs.

Cheers!

hitsthings:
This is a preliminary commit for supporting root-level context on server and client as mentioned in #751 . I'm still hoping this is something you'd be interested in. In Discord, request-specific stores were also floated which sounds like an interesting alternative to the problem I mentioned in #751 .

This change currently does _too_ much and I need feedback on how to proceed. Here's what it currently does.

In client.js, you can now do:

```js
sapper.start({
    target: ...,
    context: () => {
        setContext('key', new Blammo()); // set root-level context

        // can also return a Map to merge into the context.
        // I don't think both APIs should be supported, but they currently are.
        // Which one do you prefer?
        return new Map([['key2', new Foobar()]]);
    }
})
```

In `server.js` you can do:

```
middleware({
    context: (req, res) => {
        setContext('key', new Blammo()); // set root-level context

        // can also return a Map to merge into the context.
        // I don't think both APIs should be supported, but they currently are.
        // Which one do you prefer?
        return new Map([['key2', new Foobar()]]);
    }
})
```

In `preload` there is both a new argument _and_ support for `getContext` and `setContext`. I don't think we should support both.

```js
export async function preload(page, session, context) {
    const foobar = context.get('key2'); // can access the context initialized above.
    const blammo = getContext('key'); // again - we shouldn't support both of these.
}
```

The code for supporting `getContext` and `setContext` is also a bit dirty for my liking, and there are TODOs in there about it. Is there a better way than calling `set_current_component`?

Further works needs to be done before this is can be merged, but I really do need to talk to maintainers before I go too far in the wrong direction. Cheers!


justinmoon:
@Rich-Harris @Conduitry Could one of you take a look at this? Something along these lines would really improve Sapper developer experience.

I'm willing to help rebase / update this PR if @hitsthings isn't

hitsthings:
I've kinda stopped using Svelte, so you have my support if you want to take over this code. Ping me anytime.

benmccann:

PR #768, opened 2019-06-26
[FIX] fix css sourcemaps

### LINKED ISSUE #421: Extracted CSS sourcemaps don't work
They work perfectly when appended as inline data URLs but not when they're separate .map files. Probably just need to tweak the path

### LINKED ISSUE #537: Rollup config: Extracted CSS sourcemaps always generated
`config.client.output()` always return an object with a truthy `sourcemap` property, so I get sourcemap files for both js and css files.

Even when using 
```js
output: {...config.client.output(), sourcemap: dev}
```
when `dev` is falsy I still get sourcemap files for css files (but not for js files).

Is it related to #421 ?

Teoxoy:
fixes #421
closes #537

There were 3 issues:

- `#` was missing from `/* sourceMappingURL=<URL> */`
- `sources` array was relatively mapped twice
- `sources` array would contain windows paths on a windows system

I have tested the PR in my own project and CSS sourcemaps work now.

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -61,6 +61,7 @@ export default class RollupCompiler {
 
 	async compile(): Promise<CompileResult> {
 		const config = await this._;
+		const sourcemap = config.output.sourcemap;
 
 		const start = Date.now();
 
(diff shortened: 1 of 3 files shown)


Teoxoy:
Now closes #537 too.
The rollup compiler will now respect rollups sourcemap output option for css maps.
This means that sourcemaps will fully work as expected now and paired with #772 will be the best API imo.

Issue #137, opened 2018-02-22
POST to pages?


Rich-Harris:
At the moment only server routes can accept POST requests. This is a bit of a nuisance if you want to e.g. handle a form submission without redirecting (which creates new request/response objects, which means you need to persist the posted data).

I wonder if there should be some way for pages to handle POST. Alternatively, if it was possible to server routes and pages to shadow each other, we could do something like this:

```bash
routes/form-submit.js   # accepts the POST, calls next
routes/form-submit.html # populates the response
```

mweststrate:
Ideally, I would be able to end an api request conceptually like:

`res.write(severSideRender(require("../thanks.html"), { someData: 3 }))`

Rich-Harris:
Yeah, it'd make sense to expose that. Shouldn't even be that difficult — we're almost there with

```js
import Page from '../thanks.html';

export function post(req, res) {
  // do some things
  const { html, head, css } = Page.render({ my: 'data' });
  res.end(html); // close — just missing layout and CSS
}
```

Just need to wrap it all up in a function that wraps the HTML in the page template, supports `preload`, and serves the 5xx page in case of errors:

```js
import Page from '../thanks.html';
import { render } from 'sapper/server-runtime.js'; // open to bikeshedding

export async function post(req, res) {
  // do some things
  const html = await render(Page, { my: 'data' });
  res.end(html);
}
```

mweststrate:
just for clarification, would `{my: data}` be merged with what gets passed to `payload`, to the page component, or both?

arxpoetica:
I just ran into this problem. I'm not sure I'm following the difference between "server side routes" and something like `routes/[email].json.js`.

Does the latter not support something like `export async function post...`?

davclark:
I'm trying to use sapper to develop something that (for now) will be deployed to netlify with export. This is a hangup for me - netlify handles forms specially by noticing a "netlify" tag in the form which it will automatically rewrite to hit it's back-end.

Right now, I'm experiencing a strange error where no matter how I specify action, I get:

```
   4.39 kB   contact/index.html
   3.54 kB   (404) contact/index.html/index.html
> EEXIST: file already exists, mkdir 'C:\Users\davcl\Projects\bead.glass\__sapper__\export\contact\index.html'
```

Which I don't quite understand. I've tried having no action, and making action "", "/contact", and "/dummy" (with a dummy.js file). All give the same error.

asppsa:
[...]

PR #770, opened 2019-06-27
[WIP] Pass data from server routes to page preload

asppsa:
This is a proof of concept for my proposed solution to #137.


asppsa:
Closed on the basis that this is not required -- see https://github.com/sveltejs/sapper/issues/137#issuecomment-511096819

Issue #590, opened 2019-03-03
`dev` flag used by Rollup/webpack configs seems reversed


jarrodldavis:
It appears that source maps are not enabled for `sapper dev` but are enabled for `sapper build`. I'm using the [sapper/svelte v3 rollup] template, and the issue appears to be at the call sites of `create_compilers` in the CLI command implementations.

[sapper/svelte v3 rollup]: https://github.com/sveltejs/sapper-template/tree/rollup-v3

Here's some relevant code from some GitHub spelunking:

`sapper dev`:

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/api/dev.ts#L7

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/api/dev.ts#L220

`sapper build`:

https://github.com/sveltejs/sapper/blob/e0de230e1391ec1b4a404faea57fd2cc8bbf9812/src/api/build.ts#L4

https://github.com/sveltejs/sapper/blob/e0de230e1391ec1b4a404faea57fd2cc8bbf9812/src/api/build.ts#L85

`create_compilers`:

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/core/create_compilers/index.ts#L4

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/core/create_compilers/index.ts#L14-L23

`env.ts`:

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/config/env.ts#L1-L7

Rollup Config (`sapper/config/rollup.js`):

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/config/rollup.ts#L1

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/config/rollup.ts#L20

https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/src/config/rollup.ts#L36

Should I go ahead and submit a pull request to reverse the values passed into `create_compilers`, or is there something I'm not understanding?

thgh:
JS client sourcemaps seem to be useful:
<details>

Dev js without sourcemap (current)

```
[...]

PR #772, opened 2019-06-27
[FEAT] make rollup sourcemap independent of env

Teoxoy:
Adds `sourcemap` arg to the output function.

This issue was brought up in https://github.com/sveltejs/sapper/issues/590#issuecomment-482396543

After #771 gets merged this is a crucial option to have imo

### DIFF
--- a/src/config/rollup.ts
+++ b/src/config/rollup.ts
@@ -1,7 +1,7 @@
 import { dev, src, dest } from './env';
 import { InputOption, OutputOptions } from 'rollup';
 
-const defaultSourcemap = dev ? 'inline' : false;
+const sourcemap = dev ? 'inline' : false;
 
 export default {
 	dev,
@@ -11,7 +11,7 @@ export default {
 			return `${src}/client.js`
 		},
 
-		output: (sourcemap: boolean | 'inline' = defaultSourcemap): OutputOptions => {
+		output: (): OutputOptions => {
 			let dir = `${dest}/client`;
 			if (process.env.SAPPER_LEGACY_BUILD) dir += `/legacy`;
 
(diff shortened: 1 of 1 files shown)


Teoxoy:
Currently the default value will be undefined which will not generate any source maps.
Should I make the default option the value of the `dev` var?

Teoxoy:
By default (i.e. no `sourcemap` option is passed in) it will generate **inline** sourcemaps for `dev` and **no** sourcemaps for `build`.

Conduitry:
Is there anything else that needs to happen for webpack do you think?

Conduitry:
Actually, looking at this more, this doesn't seem like the cleanest way to go about this. This adds an argument to the `config.xxx.output()` functions, which would have to be documented somewhere. I think the proper way to do this would be to just return the output config object with the default `sourcemap` value in it, and let people override it with, for example, an object spread in their own config.

Conduitry:
Pushed _that_ change and now I'm wondering what the purpose of this PR is now that I've changed it. Is there really a benefit to using inline sourcemaps? As far as I can tell, all this PR, as amended, does is default to inline sourcemaps, and also (conditionally) enable sourcemaps on the service worker.

As I mentioned above, I don't think we want to add an argument to the `.output()` functions that just overrides the value of a single field, when this is already possible using regular old js syntax.

Teoxoy:
Yeah, I guess overriding the `sourcemap` via object spread is an alternative. This is how I currently do it alltho I thought it was a workaround.

I would still merge this tho, it adds sourcemaps for the service worker by default and also uses inline sourcemaps for development which should cover the majority of use cases. Maybe change the title of the PR because it deviated from the initial intent.

PR #719, opened 2019-05-30
Prevent client to rerender page over SSR error - fixes #710

rixo:
The preload error was serialized to `null` in preloaded data, so the client thought it got error free preloaded data and rendered the page over the error.

That what's caused the error to disappear (coupled with custom `_layout.svelte`).

But to make things worse, the client doesn't even have data to render at this point.

Issue #366, opened 2018-08-20
Drop source-map-support?


Rich-Harris:
Currently, source-map-support is installed by default when you use the middleware. This probably doesn't make sense. It *is* useful to have sourcemapped stack traces (notwithstanding that it doesn't currently apply to components — see #117), but the middleware feels like the wrong place for it.

A few options:

* Expect app authors to include it (and put it in sapper-template)
* Keep it, but only activate if `process.env.NODE_ENV` is 'development' (not sure if there's really much point to that)
* Add it in dev mode via `sapper dev` (if that's possible, given that it's a child process?)
* Scrap it, and handle errors in the child process differently inside `sapper dev`

fusionstrings:
Not sure I understood correctly but if this is about not to generate sourcemaps at all then I would like to have the possibility to emit sourcemaps into a separate directory, which can be stored in protected environments.

Rich-Harris:
This happened a few released ago so I'll close this. @fusionstrings this isn't about not generating sourcemaps, it's about not *automatically* consuming them on the server

benmccann:
The dependency was removed in https://github.com/sveltejs/sapper/commit/cfd10c6f61a861f9b29081915085add31d7ca7fc#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 as being unused. It's hard to tell where the change actually occurred unfortunately

PR #773, opened 2019-06-28
[FEAT] Map stacktraces

### LINKED ISSUE #117: Emit and Consume Sourcemaps
sapper-error-trace

Currently when you have a bug in your code the server rendered code looks like the above.

It'd be very handy to emit and consume the sourcemaps so that stack trace gives us actionable info.

teoxoy:
closes #117

`sourcemap-stacktrace` takes in a stacktrace and uses sourcemaps (if it finds any) to map the paths to the sources

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "html-minifier": "^4.0.0",
     "http-link-header": "^1.0.2",
     "shimport": "^1.0.1",
+    "source-map": "^0.7.3",
     "sourcemap-codec": "^1.4.6",
     "string-hash": "^1.1.3"
   },
--- a/runtime/src/server/middleware/sourcemap_stacktrace.ts
+++ b/runtime/src/server/middleware/sourcemap_stacktrace.ts
@@ -1,0 +1,95 @@
+import fs from 'fs';
+import path from 'path';
+import { SourceMapConsumer, RawSourceMap } from 'source-map';
+
+function retrieveSourceMapURL(contents: string) {
+    const reversed = contents
+        .split('\n')
+        .reverse()
+        .join('\n');
+
+    const match = /\/[/*]#[ \t]+sourceMappingURL=([^\s'"]+?)(?:[ \t]+|$)/gm.exec(reversed);
+    if (match) return match[1];
+
+    return undefined;
+}
+
+const fileCache = new Map<string, string>();
+
+function getFileContents(path: string) {
+    if (fileCache.has(path)) {
+        return fileCache.get(path);
+    }
+    if (fs.existsSync(path)) {
+        try {
+            const data = fs.readFileSync(path, 'utf8');
+            fileCache.set(path, data);
+            return data;
+        } catch {
+            return undefined;
+        }
+    }
+}
+
+function sourcemapStacktrace(stack: string) {
+    const replaceFn = (line: string) =>
+        line.replace(
+            /^ {4}at (?:(.+?)\s+\()?(?:(.+?):(\d+)(?::(\d+))?)\)?/,
+            (input, varName, filePath, line, column) => {
+                if (!filePath) return input;
+
+                const contents = getFileContents(filePath);
+                if (!contents) return input;
+
+                const srcMapPathOrBase64 = retrieveSourceMapURL(contents);
+                if (!srcMapPathOrBase64) return input;
+
+                let dir = path.dirname(filePath);
+                let srcMapData: string;
+
+                if (/^data:application\/json[^,]+base64,/.test(srcMapPathOrBase64)) {
+                    const rawData = srcMapPathOrBase64.slice(srcMapPathOrBase64.indexOf(',') + 1);
+                    try {
+                        srcMapData = Buffer.from(rawData, 'base64').toString();
+                    } catch {
+                        return input;
+                    }
+                } else {
+                    const absSrcMapPath = path.resolve(dir, srcMapPathOrBase64);
+                    const data = getFileContents(absSrcMapPath);
+                    if (!data) return input;
+
+                    srcMapData = data;
+                    dir = path.dirname(absSrcMapPath);
+                }
+
+                let rawSourceMap: RawSourceMap;
+                try {
+                    rawSourceMap = JSON.parse(srcMapData);
+                } catch {
+                    return input;
+                }
+
+                const consumer = new SourceMapConsumer(rawSourceMap);
+                const pos = consumer.originalPositionFor({
+                    line: Number(line),
+                    column: Number(column)
+                });
+                if (!pos.source) return input;
+
+                const absSrcPath = path.resolve(dir, pos.source);
+                const urlPart = `${absSrcPath}:${pos.line || 0}:${pos.column || 0}`;
+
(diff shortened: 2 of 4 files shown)


Teoxoy:
@Conduitry has the issue been fixed in some other PR? Should I close this one?

If not, I can add the code from `sourcemap-stacktrace` package in sapper and delete the package.
I am not really up to maintaining it anymore, I don't know why I made it a package in the first place...

benmccann:
No, I don't believe this issue has been fixed yet

Teoxoy:
So, is there an issue with this PR? Do you guys want the code in sapper?

benmccann:
I think mostly everyone's just been busy.  Also, errors weren't showing up at all due to https://github.com/sveltejs/sapper/pull/719, which made this one hard to review. Now that that's been fixed I think it's time to revisit this one.

>I can add the code from sourcemap-stacktrace package in sapper and delete the package.
I am not really up to maintaining it anymore, I don't know why I made it a package in the first place...

You said your package was based on [node-source-map-support](https://github.com/evanw/node-source-map-support), but exposes the needed functionality. Would you be able to file an issue with [node-source-map-support](https://github.com/evanw/node-source-map-support) to ask them to expose that functionality? Then we could consider using that package instead of having to add the code directly into Sapper

Also, I would suggest splitting out the linting fixes into a separate PR. We've been working on linting and you can see some of our progress in https://github.com/sveltejs/sapper/pull/1198, but unfortunately that doesn't fix the linting issues you found and we should investigate why. Perhaps we should setup prettier on the codebase

Teoxoy:
I rechecked `node-source-map-support` and it has some node v8 stacktrace specifics + it doesn't look maintained to me.

I can remove the formatting changes and move the [src](https://github.com/Teoxoy/sourcemap-stacktrace/blob/835ea8be8f603b1d612be6f21e7ef527734e7148/src/index.ts) of `sourcemap-stacktrace` (95 LOC) in the repo.
Let me know if this is how we should move forward with this.

antony:
How did this pass me by? This would be incredibly useful.

I would agree that replicating the source of that module here is probably desirable rather than bringing the whole thing in. Can we include a note so that it's clear that this is the case, and if the source module improves we can incorporate at a later date.

Also as @benmccann says, we don't accept linting fixes in PRs, unless that is the sole purpose of the PR, so they should be removed for this one.

Thanks!

benmccann:
> it has some node v8 stacktrace specifics

V8 is the Javascript engine inside of node.js that parses and runs your Javascript. The same V8 engine is used inside of Chrome to run javascript in the Chrome browser. Google open-sourced the V8 engine and the builders of node.js used it to run Javascript in node.js. The current node.js binary cannot work without V8. So I'm not sure why it'd be an issue to have v8 specifics in the code?

If you're talking about Node 8.x instead, I'm also not sure it harms us at all if it supports Node 8.x

> it doesn't look maintained to me.

I'm not sure I'd call `source-map-support` unsupported. It does have a number of open issues and PRs, but also has had some commits regularly and was last updated only two months ago and has 20m weekly downloads.

Anyway, we don't have to block on getting the functionality implemented upstream. I wouldn't want to block this feature if we can't get the support we need. But we should at least ask for it before forking the project and copying the source here that way there's at least a possibility that we don't have to maintain a fork of the project long-term.

Teoxoy:
I am aware of the v8/node relationship. What I meant to say is that `source-map-support` is using the [V8 stack trace API](https://v8.dev/docs/stack-trace-api) which doesn't work for us (this is part of the reason it was removed in #366).

I'm not sure why you refer to `source-map-support` as upstream since I mention in the readme of `sourcemap-stacktrace` that it is a reimplementation.

The relevant part to us is 1/6th of `source-map-support`; sourcemapping a stacktrace is quite simple, it's a matter of going trough each line of the stacktrace and trying to replace the path, line no. and column no.. Doing that by searching in the file that thew the error for a `sourceMappingURL` and mapping the orignal line and column numbers using [`source-map`](https://github.com/mozilla/source-map).

benmccann:
> source-map-support is using the V8 stack trace API which doesn't work for us (this is part of the reason it was removed in #366).

Can you explain why it doesn't work for us? I don't see any reference to it in #366

Teoxoy:
> notwithstanding that it doesn't currently apply to components

It has been a year since I worked on this - so, I don't actually remember why it didn't work.

benmccann:
Ok. I'm fine with this PR if you're able to move the linting fixes to a new PR and rebase it

Teoxoy:
Rebased, removed formatting/linting fixes and moved the src of `sourcemap-stacktrace` in the repo.

package.json
```diff
@@ -21,6 +21,7 @@
     "html-minifier": "^4.0.0",
     "http-link-header": "^1.0.2",
     "shimport": "^1.0.1",
+    "source-map": "^0.7.3",
```
benmccann, line 24: we don't need this anymore since you're including the file directly, right?

Teoxoy: We do need that one. It's used to do the actual mapping.

teoxoy:
I can't seem to get the test suite running in WSL and I currently don't have the time to debug it and also add tests.

benmccann:
ok. thanks for all your work on this already. this lgtm

pngwn:
This needs tests before it can be merged.

benmccann:
I was just thinking about this some more... Does anyone know why we bundle the server? I wonder if it's necessary. If we didn't bundle it then I don't think we'd need source maps for the stack traces

pngwn:
Bundling, even for the server, generally makes things a bit faster and smaller. It also makes things like building a docker image simpler as the server bundle contains all of the necessary code so you don't need to ship any dependencies.

sdwvit:
please merge & release asap

benmccann:
@sdwvit you can help it get merged. the main thing this needs is some tests. if you'd like to contribute we'd be happy to see this get merged much sooner

runtime/src/server/middleware/sourcemap_stacktrace.ts
```diff
@@ -0,0 +1,95 @@
...
+
+                let rawSourceMap: RawSourceMap;
+                try {
+                    rawSourceMap = JSON.parse(srcMapData);
+                } catch {
+                    return input;
+                }
+
+                const consumer = new SourceMapConsumer(rawSourceMap);
+                const pos = consumer.originalPositionFor({
```
habibrosyad, line 74: Apparently, with the latest version of `source-map`, the `consumer` here returns promise, therefore call to `originalPositionFor` will fail 😕

teoxoy: You are right, it should be version 0.6.1

habibrosyad: I'm planing to continue the works that you've done if you don't mind, especially adding tests for this...

teoxoy: Not at all, I mentioned in one of my previous comments that I've been quite busy lately.

PR #779, opened 2019-07-01
Fix broken link to Svelte SSR documentation

Smona:
looks like that section was removed, and SSR information is spread around the docs now. This at least jumps to an SSR-related section.


Conduitry:
:+1: I think this is a reasonable place to link to

PR #780, opened 2019-07-01
Fixes #765. Support zero-length [...rest] params.

### LINKED ISSUE #765: Optional params/zero-length rest routes
Via [this Stack Overflow question](https://stackoverflow.com/questions/56403072/how-do-you-create-routes-with-optional-parameters-in-sapper). It would be nice if there was a way to do optional parameters — e.g. `/foo`, `/fr/foo` and `/ca/fr/foo` all resolving to the same page component (where `fr` is a named `lang` parameter and `ca` is a named `country` parameter).

@Timer from the Next.js team had a neat idea — we could mark optional parameters with double brackets: `routes/[[country]]/[[lang]]/foo.svelte`

At the same time, it might be nice if `routes/[...parts]/foo.svelte` matched all those pathnames. At present it will match `/fr/foo` and `/ca/fr/foo` but not `/foo`.

hitsthings:
This is a breaking change since it changes the behavior of [...rest] params to additionally match
nothing.

[...rest] test routes were moved under /rest-variable because zero-length support would otherwise conflict with [slug].svelte.
A [...rest].svelte was added in place so that comparator would continue to implicitly test ...rest routes against other routes.

A design decision was made that `abc-[...rest]-def.svelte` wouldn't match `abc-/foo/-def`, only `abc-foo/bar-def` (no slashes at edges of [...rest] params within a segment).


hitsthings:
On re-reading #765 , this only handles [...rest] and doesn't make any optional params otherwise. The support for `some-constant[[param]]whatever.svelte` makes all this stuff wonky, but I think it wouldn't be _too_ hard to add double bracket support if that's what you want. Let me know if you do.

hitsthings:
One thing I don't understand is the code to support something like `index.whatever.svelte`. If you trace the Blame back it attaches to https://github.com/sveltejs/sapper/issues/262 which as far as I know is now `_layout.svelte`.
The `is_index` `route_suffix` code looked somewhat relevant, but I didn't look deep enough to really understand it or whether it would complicate anything.

nickreese:
What's blocking the merge on this? Would love to be able to use this functionality.

Conduitry:
This would be a breaking change, which can always be a bit of a merge delayer. It removes the ability to have spread routes that only match one or more segment. I don't know whether preserving that should be a goal in particular, and I don't know whether we want to try to instead solve this in a more unified way along with optional regular params.

nickreese:
@Conduitry Got it. Thanks. I was actually looking for this to be implemented and commented on the wrong ticket. https://github.com/sveltejs/sapper/issues/545 Glad to see it was merged and working. :)

PR #788, opened 2019-07-03
update tag line

davidberglund:
Should be done, right? And the project description, too.


Rich-Harris:
yep, missed this. thank you!

PR #793, opened 2019-07-06
Fix for detaching meta in <svelte:head>

matyunya:
Fix for the problem discovered here:
https://github.com/sveltejs/sapper-template/pull/158

Page transition breaks if there's `meta` tag inside `<svelte:head>`.

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -386,5 +386,7 @@ export function load_component(component: ComponentLoader): Promise<{
 }
 
 function detach(node: Node) {
+	if (!node.parentNode) return;
+
 	node.parentNode.removeChild(node);
 }


Conduitry:
Do we know why these nodes don't have a `parentNode` by the time we call this? Have they already been detached by some other code?

evdama:
No idea, but svelte has the same issue https://github.com/sveltejs/svelte/pull/3188

mrkishi:
I believe the actual problem would be solved by https://github.com/sveltejs/svelte/pull/3172, no?

Conduitry:
See https://github.com/sveltejs/svelte/issues/3187#issuecomment-508942887 - this isn't the right place to solve this.

PR #811, opened 2019-07-19
Fix source maps for rollup builds

### LINKED ISSUE #808: "DevTools failed to parse SourceMap" for sapper-template#rollup
When I build sapper-template using the rollup configuration, loading the page in Chrome 75 prints `Devtools failed to parse SourceMap: http://localhost:3000/client/client.0e04fd89.js.map`.

Repro steps are just:

```
npx degit "sveltejs/sapper-template#rollup" my-app
cd my-app
npm install
npm run build
node __sapper__/build
```

and then open http://localhost:3000 in Chrome, and open the DevTools console.

Notably, if the webpack config is used and tweaked to enable source maps in production builds like so:

```
-		devtool: dev && 'inline-source-map'
+		devtool: dev ? 'inline-source-map' : 'source-map',
```

then there's no error message, and source maps seem to work correctly.

mdempsky:
When substituting \_\_SAPPER_CSS_PLACEHOLDER:foo\_\_, tighten the regexp
to require matching quotation marks and to not match too much if
there's an error.

Also, make the replacement logic smart about recognizing when we're
matching JavaScript code that's been quoted (e.g., in source maps),
and to generate a substitution that's still valid in that context.

Fixes #808.

### DIFF
--- a/src/core/create_compilers/extract_css.ts
+++ b/src/core/create_compilers/extract_css.ts
@@ -216,8 +216,21 @@ export default function extract_css(client_result: CompileResult, components: Pa
 
 		const source = fs.readFileSync(`${asset_dir}/${file}`, 'utf-8');
 
-		const replaced = source.replace(/["']__SAPPER_CSS_PLACEHOLDER:(.+?)__["']/g, (m, route) => {
-			return JSON.stringify(result.chunks[route]);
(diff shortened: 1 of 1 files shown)


mdempsky:
Alternatively, I'm wondering if the problem is that the PLACEHOLDER substitution needs to happen before source maps are generated for these JS files, and then sapper just wouldn't have to worry about touching .js.map files.

Issue #812, opened 2019-07-20
Proper error handling for server routes instead of sending all errors as a text response


racbart:
**Is your feature request related to a problem? Please describe.**

All errors originating in server routes are currently sent as a text response by Sapper and there is no way to prevent this. This might cause to leak some information which should not happen on production. It also requires somehow boilerplate-ish code in every server route if you need to either log the error or to format it in some other way (e.g. formulating a JSON response). 

**Describe the solution you'd like**

An option to provide a function to handle server route errors in a centralised way.

**Describe alternatives you've considered**

Injecting express/polka middleware before or after Sapper's cannot solve this problem, because Sapper handles server routing internally in its middleware, catches and handles all errors inside.

**How important is this feature to you?**

I believe it's critical for any real-world production deployment to be able to control what happens with errors. You usually need to log them and you need to control the output sent to the browser. At least, it is critical to me.

**Additional context**

I created a patch for this and will submit a pull request in a moment.

zolotokrylin:
This functionality is very much needed!

sdwvit:
please fix this, it's very hard to debug

PR #813, opened 2019-07-20
Custom error handler for server routes

racbart:
Regarding: #812 

This commit solves the problem of handling server route errors, which were not properly handled before. It provides opportunity to provide a function which can handle errors in a centralised way (e.g. log them, hide their content from end-user, send them to browser using JSON, etc).

The following code allows providing custom error handler function as an `error_handler` option to the middleware. I also included relevant documentation, which is self-explanatory, so won't repeat it here.

### DIFF
--- a/site/content/docs/02-routing.md
+++ b/site/content/docs/02-routing.md
@@ -118,6 +118,25 @@ The `error` object is made available to the template along with the HTTP `status
 
 
 
+### Error handling for server routes
+
+By default, errors occurred in server routes are sent as a text response with the HTTP 500 status. You can define your own error handler by passing it as an option to Sapper middleware:
+
+```js
+function on_error(err, req, res, next) {
+  res.setHeader('Content-Type', 'application/json');
+  res.statusCode = 500;
+	res.end(JSON.stringify({ error: err.message }));
+  // Log error here, etc
+}
+
+app.use(sapper.middleware({ error_handler: on_error }));
+```
+
+Your error handler should either serve the response by calling `res.end()` or pass the error to the default error handler by calling `next(err)`. If your handler throws an error, the default error handler will be used as a fallback (with the original server route error, not your error handler error).
+
+
+
 ### Regexes in routes
 
 You can use a subset of regular expressions to qualify route parameters, by placing them in parentheses after the parameter name.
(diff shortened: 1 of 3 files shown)


jamen:
Hey. I would like this feature too. You could setup error handling for a JSON API as:

```js
// src/routes/_error.json.js

export default function onError (error, request, response, next) {
    // If JSON is not accepted, pass to next error handler _error.svelte.
    if (request.headers.accept !== 'application/json') {
        return next(error)
    }

    // Send JSON error
    response.setHeader('Content-Type', 'application/json')
    response.statusCode = error.statusCode || 500
    response.end(JSON.stringify({ name: error.name, message: error.message }))
}
```
```js
// src/server.js

import polka from 'polka' 
import * as sapper from '@sapper/server'
import onError from './routes/_error.json.js'

const app = polka()
// ...

app.use(sapper.middleware({ onError }))
```

Then, when you write a server route, you can handle errors better:

```js
// src/routes/users.json.js

import { users } from '../database.js'

export async function get (request, response, next) {
    // There's no way to handle this without wrapping individual server routes.
    const { rows } = await users.list(...)
    request.end(JSON.stringify(rows))
}
```

Instead of wrapping the server route in an error handler:

```js
// src/routes/users.json.js

import { users } from '../database.js'
import route from '../route.js'

export const get = route(async function (request, response) {
  // ...
})
```

davemccrea:
This would be a welcome addition!

UnlimitedBytes:
Is this still in work?

site/content/docs/02-routing.md
```diff
@@ -118,6 +118,25 @@ The `error` object is made available to the template along with the HTTP `status
 
 
 
+### Error handling for server routes
+
+By default, errors occurred in server routes are sent as a text response with the HTTP 500 status. You can define your own error handler by passing it as an option to Sapper middleware:
+
+```js
+function on_error(err, req, res, next) {
+  res.setHeader('Content-Type', 'application/json');
```
benmccann, line 127: the formatting here is inconsistent. the spaces on this line and the next should replaced with tabs for consistency with the rest of the file

GiorgioG:
Is this dead?

benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit, so we'd be open to making sure this is possible there if you'd like to take a look

Issue #627, opened 2019-04-13
npm run dev: Error: ENOENT: no such file or directory, open '__sapper__/build/build.json'


btakita:
To reproduce, run

`npm run dev`

Then, while dev is running, run `npm run build`

The `npm run dev` process will get the following error:

`Error: ENOENT: no such file or directory, open '__sapper__/build/build.json'`

vladejs:
npm run build is an important ritual that should be run alone :)

thgh:
+1 for allowing parallel build & dev

btakita:
It seems like the issue is with

```js
fs.writeFileSync(path.join(dest, 'build.json'), JSON.stringify(build_info));
```

https://github.com/sveltejs/sapper/blob/master/src/api/build.ts#L113

There watch process is running in parallel. If a temp file were written & then moved, via the shell, to the dest/build.json, then the operation should be atomic from an OS perspective.

RailgunWinjoy:
How to solve this ?

evdama:
I tried with a fresh install and Sapper 0.28.1 and I'm getting this error as well 🤔
![Screen Shot 2020-08-26 at 07 19 28](https://user-images.githubusercontent.com/39374084/91258416-96c1f800-e76c-11ea-9974-e6219e897341.png)

MaxMilton:
I didn't realise this issues was already open... I've created a new one with more detailed information and a repro repo. The cause behind the error now is different to when this issue was opened so maybe it's time to close this one?

btakita:
@MaxMilton These appear to be separate issues. I'm still getting this (https://github.com/sveltejs/sapper/issues/627) issue with the 0.28.4 release. This issue is not a race condition. It occurs every time `npm run build` is invoked when `npm run dev` is already running.

yangdachaun:
root@dev-VirtualBox:/media/sf_work/ewechat# npm run dev
Error: ENOENT: no such file or directory, uv_cwd
    at process.wrappedCwd (internal/bootstrap/switches/does_own_process_state.js:128:26)
    at process.cwd (/root/.nvm/versions/node/v12.18.3/lib/node_modules/npm/node_modules/graceful-fs/polyfills.js:10:19)
    at Conf.loadPrefix (/root/.nvm/versions/node/v12.18.3/lib/node_modules/npm/lib/config/load-prefix.js:46:24)
[...]

PR #821, opened 2019-07-25
Use write-file-atomic to write template.html & build.json.

btakita:
Fixes https://github.com/sveltejs/sapper/issues/627


btakita:
I tested a hotfix branch & it does not seem to fix the issue. I'll leave this here, as there may be multiple places where it's needed for atomic wirtes. In which case, write-file-atomic would be of use

benmccann:
I'm going to go ahead and close this for now since the indication is that it doesn't fix the issue and it's pretty far out-of-date

PR #825, opened 2019-07-26
Allow export to take multiple entrypoints to use when crawling.

### LINKED ISSUE #749: export routes not reachable by <a> elements
For now, only way to include routes via export is to add `<a>` element with href attribute pointing that route in some other reachable route https://sapper.svelte.dev/docs#How_it_works

But there are cases when some routes cannot be reachable by `<a>`, and still needs to be exported.
In my case I have success page. When form was submitted correctly I do `goto('/success')`.

It should be a way to pass to `sapper export` routes that should be exported as well. Or maybe export should not rely on `<a>` tag only.

pngwn:
I am  PRing this again because I'm a genius and the original PR was from the master of my svelte fork. I was then too terrified to actually try to do anything on another branch in-case I broke something.

Original PR: #756.

---

This PR allows you to pass a list of entry points to export, which allows users to fully export apps that are not entirely linked by anchor tags.

I'm not sure if this is the best way to do it but it seems to work, I've added a test that provides a few unlinked entry points in the various styles people might try to pass in. I've done my best to handle unnecessary `index.html`s and `/`s (I'm no regex expert so that might need looking at). I also moved the `oninfo` call so users get a console message for each entry point, followed by the list of saved files for that 'crawl'.

I added `.vscode` to the gitignore as I was having trouble disabling my autoformatting with out it. I don't know what has happened with the `package-lock.json` it just seems to add some stuff when I install it.

Closes #749.


Conduitry:
What does the `currentRoot` variable do that's different from the `i` in the loop? I'm confused about how this is implemented.

pngwn:
From what I can tell, nothing, I probably just thought it was more descriptive. I don't really remember, I wrote this code quite a while ago.

In terms of implementation, It's just running `handle` in a loop instead of doing it just the once on `/`.

src/api/export.ts
```diff
@@ -201,7 +212,7 @@ async function _export({
 			type = 'text/html';
 			body = `<script>window.location.href = "${location.replace(origin, '')}"</script>`;
 
-			tasks.push(handle(resolve(root.href, location)));
+			tasks.push(handle(resolve(entryPoints[currentRoot].href, location)));
```
Conduitry: What's the reason for resolving this relative to the current entry point? I _think_ this should still always be relative to `root`. My understanding is that the multiple entry points is just so that the crawler can find pages it wouldn't ordinarily find. All of them should still have the same base href value.

Conduitry: The test you wrote still passes when using `root.href`, fwiw

pngwn: Yeah, I think you're right. I don't think I looked too deeply into this but I guess this clause is handling redirects/ 300s? I think it's using the `Location` header and building a new url to `handle` from that. My tests definitely don't test this behaviour, I don't know if any of the existing ones do either but resolving with anything but the root url would fail, I think.

pngwn: Yeah, looking back this clears up some of the confusion I had at the time with the whole `entryPoints[currentRoot]` thing, I wondered why that was necessary. This correction removes the need for that entirely. I probably should have taken more time to understand the code before implementing this.

onionhammer:
It looks like this was broken recently

kevlened:
@onionhammer I also thought this was broken; I was using the wrong syntax. I tried comma-separation and multiple `--entry` flags, but entries should be separated by spaces: `sapper export --entry="/ /products /products/example"`

Issue #829, opened 2019-07-29
Clicking two links in rapid succession causes the entire site to break


antony:
**Describe the bug**
When performing non-trivial tasks in a preload method, it is possible to click two links in quick succession before the page/view, which results in completely, irrecoverably breaking the Sapper site (other than full reload of the page)

**To Reproduce**
Use the repository at [https://github.com/beyonk-adventures/sapper-preload-pushstate-issue](https://github.com/beyonk-adventures/sapper-preload-pushstate-issue)

1. Visit the homepage, and click any link. Preload takes 5 seconds.
1. During this five second delay you should click some other links on the page.
1. Notice in the browser console, each consecutive click results in the page loading but the actual detach of the dom node is abandoned since a new route change has taken place.
1. Notice the URL in your browser (pushstate) changing regardless
1. Let the next page finally load.
1. When the word page shows up, hit back, you will notice the url changes to one of the abandoned requests
1. Notice that the word on the page you are looking at no longer relates to the word in the status bar - the page has entered a "broken" state.

**Expected behavior**
1. Hitting back should probably result in each of the previous of abandoned pages loading, or, the urls should not have been pushed into the browser history, and the homepage should load instead. 

**Information about your Sapper Installation:**
- Chrome (any version), seems worse on Safari.
- Ubuntu Linux 19.04
- Locally hosted
- Sapper 0.27.0
- Svelte 3.6.9
- Rollup

**Severity**
Blocker Severity - We are currently blocked from a v3 upgrade as the site quickly becomes broken for users on iPhone which is ~80% of our userbase. You can play with it at [https://qa.beyonk.com/uk](https://qa.beyonk.com/uk) and see that visiting an adventure and hitting back, and then hitting another, causes a scroll, making it very easy to click two links in rapid succession and completely break the site.

**Additional context**
1. The broken state of the page is intensified on a non-trivial app, the entire site's navigation is now broken and only a refresh will fix it.
1. The broken state of a page extends to all javascript in the client - in a non trivial site, this means that the site now appears "frozen/crashed" to a user. Any javascript on the page will no longer function.
1. A refresh of the browser is the only way to recover a Sapper site from this state.
1. The issue is far more prevalent on iOS Safari / Safari desktop - I'm not sure why.
[...]

PR #831, opened 2019-07-30
Feature/pushstate on successful navigation

antony:
This is a tentative fix for https://github.com/sveltejs/sapper/issues/829 - it *seems* to work but I am not in a position to assess the full impact of the change, or if there are edge cases I have missed.

WIP as I've just noticed failing tests that I need to fix.

### Tests


antony:
@Rich-Harris @Conduitry @pngwn would appreciate some feedback on whether my proposed solution is worth pursuing or whether I'm breaking something more fundamental in Sapper with this fix.

Right now I've rebuilt the whole of Beyonk in v3 but am completely blocked from going live with it because of this issue, so I'm very eager to fix it.

PR #850, opened 2019-08-18
Add documentation for Docker and Google Cloud deployment

sanderhahn:
Wanted to add gcloud to the sapper template, but was attended that this is probably a better place.

### DIFF
--- a/site/content/docs/11-deploying.md
+++ b/site/content/docs/11-deploying.md
@@ -62,3 +62,87 @@ now -e SAPPER_TIMESTAMP=$(date +%s%3N)
 ```
 
 [Now]: https://zeit.co/now
+
+
+## Deploying to Docker
+
+[Docker](https://docs.docker.com/install/) can be used to build an image and run it on your local system. Doing this locally allows testing the build process before building and deploying it on a cloud platform. The Docker image can be build by adding a `Dockerfile` into your project:
+
+```Dockerfile
+# Dockerfile
+FROM mhart/alpine-node:12 as build
+
+WORKDIR /app
+COPY . .
+RUN npm ci
+RUN npm run build
+
+FROM mhart/alpine-node:12 as prod
+
+WORKDIR /app
+COPY package.json package-lock.json ./
+RUN npm ci --production
+
+FROM mhart/alpine-node:slim-12
+
+WORKDIR /app
+COPY --from=build /app/__sapper__/build __sapper__/build
+COPY --from=build /app/static static
+COPY --from=prod /app .
+
+CMD ["node", "__sapper__/build"]
+```
+
+This Dockerfile is a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) that performs three steps:
+
+- Perform a Sapper build with development packages
+- Perform a npm install with only production packages
+- Aggregate the build artefacts and the production packages into a single image
+
+The three steps ensure that you have a very small image that only contains the necessary files for running Sapper in production. The image is based upon [mhart/alpine-node](https://hub.docker.com/r/mhart/alpine-node/) that is a small linux image that uses Alpine and Node.
+
+
+### Docker commands
+
+Commands available to build and run a local Docker image listed below. The names `sapper-image` and `sapper-container` are just examples that you can replace with something more appropiate.
+
+```bash
+# build a docker image
+docker build -t sapper-image .
+# run docker image locally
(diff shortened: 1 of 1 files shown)


QuantumInformation:
Awesome work

QuantumInformation:
Do you feel that this is more clear than the makefile here?

https://github.com/sveltejs/realworld/blob/master/Makefile#L6

sanderhahn:
Lots of developers might not know Makefile.

site/content/docs/11-deploying.md
```diff
@@ -61,4 +61,100 @@ When deploying to [Now][], you can pass the environment variable into Now itself
...
+```Dockerfile
+# Dockerfile
+FROM mhart/alpine-node:12 as build
+
+WORKDIR /app
+COPY . .
+RUN npm ci
+RUN npm run build
+
+FROM mhart/alpine-node:12 as prod
```
sanderhahn: This step can be skipped and the build can use [`npm prune --prod`](https://github.com/sanderhahn/cloud-run-button-dockerfile/blob/master/Dockerfile#L6) if the polka dependency is updated in the template.

Conduitry: If you mean doing `npm prune` instead of starting a starting from a new base image and doing `COPY --from`, that's not a good idea. While the files may not be present in the final image, they will still be present in one of its intermediate layers, and will contribute to its size.

Conduitry: Oh, do you mean still doing the separate `COPY --from` below?

sanderhahn: Yes, was thinking it would start out with the fresh `mhart/alpine-node:slim-12`, at least the image size didn't increase.

QuantumInformation:
I tried 
`docker run --name sapper-image -p 3000:3000 -d sapper-container` after building but get

Unable to find image 'sapper-container:latest' locally
docker: Error response from daemon: pull access denied for sapper-container, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.

sanderhahn:
Yes thanks, i mixed up the params...

QuantumInformation:
Did you have any luck getting it to run on Google Cloud run, other than port 8080? see
https://github.com/sveltejs/realworld/commit/f87e35796f53b646074816c9f9ea338b7c0ff2cb#commitcomment-34889238

sanderhahn:
The PORT in the Dockerfile is ignored by Google Cloud Run and the PORT is passed in thru the environment variable. See [Container runtime contract](https://cloud.google.com/run/docs/reference/container-contract#port) the section about PORT:

> In Cloud Run container instances, the PORT environment variable is always set to 8080, but for portability reasons, your code should not hardcode this value.

Having a PORT in the Dockerfile is still useful for running in Docker locally.

QuantumInformation:
I see thx

QuantumInformation:
Have you used datastore? I'm trying to use the emulator

https://github.com/gothinkster/gcp-datastore-cloud-functions-realworld-example-app/issues/40

site/content/docs/11-deploying.md
```diff
@@ -61,4 +61,100 @@ When deploying to [Now][], you can pass the environment variable into Now itself
...
+
+```
+# app.yaml
+runtime: nodejs10
+```
+
+A [custom build step](https://cloud.google.com/appengine/docs/standard/nodejs/running-custom-build-step) can be run by adding `"gcp-build": "run-p build"` as a `scripts` entry to your `package.json` scripts. This ensures that the build step will be performed before the application starts. Now you can deploy using the `gcloud app deploy` command.
+
+
+## Deploying to Docker
```
benmccann, line 79: nit: I might say "Deploying with Docker".  E.g. you're deploying to Google Cloud Run with Docker

sanderhahn: Do you really want an update? Maybe its better not to add this, because cloud providers are always changing things and it may become obsolete and will generate more questions to answer. Probably better to focus on building code instead...

benmccann: I imagine the Docker part should stay relatively constant, but if you want to drop the Google Cloud stuff from this PR that's fine with me

sanderhahn: Can do a new PR, because i might have removed my repo when i was cleaning up.

sanderhahn:
Replaced by PR #1228

PR #853, opened 2019-08-20
Reload site on changes in static directory

sanderhahn:
This makes the development experience similar to how the Svelte dev server is working.

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -212,6 +212,13 @@ class Watcher extends EventEmitter {
 
 		if (this.live) {
 			this.filewatchers.push(
+				fs.watch(`${static_files}/global.css`, () => {
+					this.dev_server.send({
+						action: 'reload'
+					});
+				})
+			);
+			this.filewatchers.push(
 				fs.watch(`${src}/template.html`, () => {
 					this.dev_server.send({
 						action: 'reload'
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -513,10 +513,7 @@ function watch_dir(
 
 		watch = new CheapWatch({ dir, filter, debounce: 50 });
 
-		watch.on('+', ({ isNew }: { isNew: boolean }) => {
-			// ignore isNew to refresh the manifest_data.components has_preload values
-			callback();
-		});
+		watch.on('+', callback);
 
 		watch.on('-', callback);
 
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -514,7 +514,8 @@ function watch_dir(
 		watch = new CheapWatch({ dir, filter, debounce: 50 });
 
 		watch.on('+', ({ isNew }: { isNew: boolean }) => {
-			if (isNew) callback();
+			// ignore isNew to refresh the manifest_data.components has_preload values
+			callback();
 		});
 
 		watch.on('-', callback);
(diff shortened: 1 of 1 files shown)


antony:
@sanderhahn This is far too specific, none of my sites have a global css file, and in fact use dynamically generated CSS, built from SASS files.

I like the idea, but this isn't the solution.

sanderhahn:
Well it doesn't solve your problem, but it does make the development experience friendlier for onboarding. Especially if you want the CodeSandbox to work nicely for Sapper templates (in the near future, because now it isn't working at all).

Do your sites output their css into the static folder? Because we can also reload on any change in that folder...

antony:
Generally people are discouraged from adding styles to the `global.css` file and are encouraged to use component scoped styles instead. I think to add reloading to the `global.css` file would convey the wrong message.

Sometimes I build css to the static folder, but that is not "live", it's a one-time update on build, I wouldn't want to watch that file, even in dev mode. I think watching files in a static folder might also be misleading.

sanderhahn:
Why don't you remove `global.css` if you want to discourage people for using it? Why do you offer livereloading for it in Svelte and not in Sapper? This seems inconsistent...

What about this is misleading? Don't understand that statement...

antony:
To be clear, I don't offer anything :)

Agreed that livereloading in one and not the other is inconsistent. I didn't know that it was being watched in Svelte.

My suggestion is that having a directory called `static` alludes to its content being static. Static content isn't something that you'd want to watch for changes. Additionally, watching very large files (binaries, videos etc) could cause a performance penalty.

sanderhahn:
The watcher is looking for changes in the directory structure and file modification dates. the project that is doing the actually watching is called `cheap-watch`. There is nothing inherently heavy about it and watching a single file that might not even be there.

sanderhahn:
Was also wondering why the directory is called `public` in Svelte and `static` in Sapper. Because its use seems similar apart from the fact that it is being watched for changes in one project and not in the other. But i suppose that is because you want to handle it differently...

pngwn:
I don't think we discourage using global CSS files *completely*. For example, we generally recommend people use CSS-vars, and the natural place for app-wide CSS-vars is in a global CSS file.

I agree with @sanderhahn that the inconsistency isn't ideal, and reloading when the `global.css` file is changed is probably a good idea. But I also agree that it is very specific, if you rename the file then obviously you lose the live-reload with no option to configure. 

I'm not sure what the best solution is here, we generally try to avoid too many configuration options where possible. If `global.css` is the only 'watched' css file, it almost becomes part of the API, and I don't know how I feel about that.

sanderhahn:
How about renaming the `static` directory to `public` and watching for file changes on the whole directory? This behavior is triggered by the `--live Reload on changes if not using --hot` flag that has a default value of `true` and doesn't state where this reloading is limited. The only thing i was thinking of that you want to explicitly not reload on changes in `manifest.json` because the server regenerates this file. Are there any other files that might live in the directory that can trigger unintentional reloads?

njbotkin:
I tend to think this should be a build configuration thing, since `global.css` isn't an intrinsically sapper thing.  It's easy enough to do this in rollup, don't know about webpack;

```js
server: {
	...
	plugins: [
		...
		{ buildStart() { this.addWatchFile('static/global.css') } }
	]
}```

Maybe put something like this in `sapper-template` to demonstrate?

sanderhahn:
Now its watching the `static` directory for changes in a similar fashion as Svelte is watching the `public` directory.

PR #854, opened 2019-08-21
Allow to use window.location.protocol on dev server

jancassio:
This is just a small feature to allow play with `https` on local environment.

I'm not sure how to test it since it's just change the dev client, not the application at all.

Feedbacks are always welcome :)

### DIFF
--- a/sapper-dev-client.js
+++ b/sapper-dev-client.js
@@ -13,7 +13,7 @@ function check() {
 export function connect(port) {
 	if (source || !window.EventSource) return;
 
-	source = new EventSource(`http://${window.location.hostname}:${port}/__sapper__`);
+	source = new EventSource(`${window.location.protocol}://${window.location.hostname}:${port}/__sapper__`);
 
 	window.source = source;
 


jancassio:
I'll leave it for a while because I figured out it won't work on many cases. Anyway, it could be solved using some http https proxy solution :)

sanderhahn:
Do think that this can work, however the `window.location.protocol` also contains a colon. Can you reopen this PR because it might allow the [codesandbox.io Sapper server template](https://codesandbox.io/s/github/codesandbox-app/sapper-template) to enable reloading on file changes.

sapper-dev-client.js
```diff
@@ -13,7 +13,7 @@ function check() {
 export function connect(port) {
 	if (source || !window.EventSource) return;
 
-	source = new EventSource(`http://${window.location.hostname}:${port}/__sapper__`);
```
sanderhahn: The [`window.location.protocol`](https://webplatform.github.io/docs/apis/location/protocol/) is of the format `http:` or `https:` so i think you should remove the extra colon.

PR #856, opened 2019-08-21
entry-only optional param for export

nealalpert:
This PR adds an optional parameter to the export command to allow export to be run without crawling the html of the entry pages for URLs to other pages.

In cases where only the data that drives a specific page or pages has been updated its more efficient to export just those pages instead of the entire site.

**Use case:** 
An author updates content on a single page of a CMS that drives the content for a sapper/svelte generated site with tens of thousands of pages.  Targeting just the specific page for export with a webhook can save time and lambda usage while exporting the static html that needs to be deployed to a CDN.

Added an optional parameter to export.ts:
```diff
-	entry?: string;
+	entry?: string,
+	entry_only?: boolean,
```

Added a check also in export.ts:
```diff
-	if (pathname !== '/service-worker-index.html') {
+	if (pathname !== '/service-worker-index.html' && !entry_only) {
```

Added a test to verify that only files needed to render the entry points are rendered when the `entry_only` parameter is true and that URLs in the page body are not crawled.


sw-yx:
instead of adding a boolean flag, which we may later want to tweak for future compatibility reasons (or offer more export options), may i suggest a union of string literals? also a naming that more closely reflects that this is explicitly for exporting. so something like

```
```

khrome83:
@sw-yx - the default behavior is all. This is overriding that and suggesting to not crawl anything else. It is a boolean by nature. 

We have another PR up for a `--filter` #859 option that provides alternatives to this boolean, like doing only `/en` or `/es` directories.

khrome83:
@Conduitry - can we take a lot at this as well. We would love to get this merged in.

sw-yx:
cool, just trying to help guard against combinatorial explosion of flags :) i agree this funcitonality is useful

sw-yx:
as i was just chatting with @pngwn today, [speed is an issue](https://github.com/sveltejs/sapper/issues/172#issuecomment-531301038) and it occurs to me that if we have `entry-only` then we dont really need puppeteer, do we? should we be a lot more aggressive with this?

khrome83:
@sw-yx yeah makes sense. Our primary use case is exporting Sapper. We will be doing some more PR most likely to try and give it some super powers.

sw-yx:
@Conduitry / @pngwn, hate to chime in with a +1, but this is relevant for the static site library i'm working on and obv Khrome and Neal are heavy users - would love some indication on whether this is inline with project goals or no

lyptt:
Similar to @sw-yx, my company has clients with 80k+ pages SSRed in production, and I'd like to be able to propose Sapper as an option alongside the usual React + Next.js / Gatsby options. I believe this is the only missing link to make that viable.

khrome83:
`entry-only` param is used by us for incremental builds.

To export the 30k pages we have, we had to also deal with the memory footprint. @nealalpert has another issue #869 that addresses this as well. You would need that to deal with 80k pages.

We also used the #859 to split the build. Our CI/CD takes our site, and using these params, spins up several FarGate instances at the largest size. We then build portions of the site on each one. Then we push them to s3 in about a minute (python lib called parallel-put). This lets us build and deploy 30k pages in ~6m. We have some more work to do with invalidating the CDN and getting the time below 5m. Overall it costs us about $0.04 cents each time we deploy.

If you really are going to export 80k pages, just know Sapper as-is is not performant enough. It uses too much memory and has a leak. It has a concurrency concept, but the file I/O prevents that from working as expected. It is not friendly for incremental builds, or parallel builds. And without additional performance work, exporting 30k pages took us closer to 30 minutes. The work we put in allows us to export our site in considerably less time.

Best of luck to you @lyptt. I know @sw-yx was building an SSG ontop of Sapper. I don't know the current state of it, and if it has grown to support these issues. 

If you're interested, our fork that addresses these issues is public. But we are building it to support our needs in a backward-compatible way. So it could be used if you desire. Our intent is not to maintain it long-term. But nothing has been addressed in Sapper in way of PR since September, and we fear the project is vaporware.

lyptt:
@khrome83 thanks for the insight. I may look at your fork as an alternative option. It seems like this repository whilst not quite dead is definitely stalled, so I'll probably hedge my bets for a while.

sw-yx:
well, ssg is lightly used by me and im just at the point of "idk if i want to make this a thing and have to maintain it for no money" stage haha. however i did merge in and publish https://www.npmjs.com/package/@ssgjs/sapper which has all of Neal's PRs merged for use. 

if y'all think http://npm.im/ssg is something you want to co-own, let me know.

benmccann:

PR #857, opened 2019-08-21
Mention `preload` being called on the server and on navigation only

thenoseman:
Hi!

This is a quick docs PR for an issue I had trying to understand the inner workings of `preload`.

For me it wasn't clear that it is only called on the server and on page navigation.

It's kind of implicitly stated [in the docs](http://localhost:3000/docs#Return_value) but I would find it useful to state that right at the beginning.

Cheers,
thenoseman


evdama:
Good point, I also think that one sentence could help save time in understanding this fact for a lot of people.

Panya:
I think your statement "only called on the server" is confusing, because `preload` is called on the client-side as well (when you navigate between routes).

benmccann:
I wasn't able to merge this since there was a merge conflict, but pushed a new commit with this change: https://github.com/sveltejs/sapper/commit/6c82d3c8298c4712e699c55f8539ef8a5b50016f

PR #864, opened 2019-08-30
Start with 'seen' set from copied static assets

jormaj:
I started working on this branch when I noticed that assets in static that were being linked to from a webpage got corrupted when running `sapper export` (pdf files in particular).
THEN I noticed that I was on an older version of sapper (0.27.4) and that this issue has been resolved since (ie pdfs appear fine in sapper exports in version 0.27.8)
Still, I think it makes sense to skip crawling files in `static`, since those got copied before so no point in going over those again.

Of course happy to get some feedback, still really new in this code base (and js in general) :)

### DIFF
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -68,7 +68,9 @@ async function _export({
 	rimraf(export_dir);
 
 	const seen: Set<string> = new Set();
-	copy(static_files, export_dir, seen, static_files + "/");
+	copy(static_files, export_dir)
+		.forEach((element:string) => seen.add(element.replace(static_files + "/", "")));
+
 	copy(path.join(build_dir, 'client'), path.join(export_dir, 'client'));
 	copy(path.join(build_dir, 'service-worker.js'), path.join(export_dir, 'service-worker.js'));
 	copy(path.join(build_dir, 'service-worker.js.map'), path.join(export_dir, 'service-worker.js.map'));
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -67,7 +67,7 @@ async function _export({
 	// Prep output directory
 	rimraf(export_dir);
 
-	copy(static_files, export_dir);
+	const seen = copy(static_files, export_dir);
 	copy(path.join(build_dir, 'client'), path.join(export_dir, 'client'));
 	copy(path.join(build_dir, 'service-worker.js'), path.join(export_dir, 'service-worker.js'));
 	copy(path.join(build_dir, 'service-worker.js.map'), path.join(export_dir, 'service-worker.js.map'));
@@ -98,7 +98,6 @@ async function _export({
 		}, process.env)
 	});
 
-	const seen = new Set();
 	const saved = new Set();
 	const q = yootils.queue(concurrent);
 
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -67,7 +67,8 @@ async function _export({
 	// Prep output directory
 	rimraf(export_dir);
 
-	const seen = copy(static_files, export_dir);
+	const seen: Set<string> = new Set();
+	copy(static_files, export_dir, seen, static_files + "/");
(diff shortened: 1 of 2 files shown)


src/api/utils/fs_utils.ts
```diff
@@ -31,16 +31,24 @@ export function rimraf(thing: string) {
...
 
 export function copy(from: string, to: string) {
-	if (!fs.existsSync(from)) return;
-
-	const stats = fs.statSync(from);
-
-	if (stats.isDirectory()) {
+	const copied = new Set();
+	const startDir = from + "/";
+	copyFiles(from, to);
```
khrome83: It's bad practice to call a function before it's declared. It also might make sense to not create a new method for this behavior, since the signature is still `from` and `to`. I would imagine copy would just call itself.

src/api/utils/fs_utils.ts
```diff
@@ -31,16 +31,24 @@ export function rimraf(thing: string) {
...
+  
+	function copyFiles(from: string, to: string) {
+	  if (!fs.existsSync(from)) return;
+  
+	  const stats = fs.statSync(from);
+  
+	  if (stats.isDirectory()) {
 		fs.readdirSync(from).forEach(file => {
-			copy(path.join(from, file), path.join(to, file));
+		  copyFiles(path.join(from, file), path.join(to, file));
```
khrome83: Your recursively calling a method, it would be better to call copy, instead of a method created within it.

src/api/utils/fs_utils.ts
```diff
@@ -31,16 +31,24 @@ export function rimraf(thing: string) {
...
-
-	const stats = fs.statSync(from);
-
-	if (stats.isDirectory()) {
+	const copied = new Set();
+	const startDir = from + "/";
+	copyFiles(from, to);
+	return copied;
+  
+	function copyFiles(from: string, to: string) {
```
khrome83: Avoid creating this function within another function. If you are going to/need to do something like this, use `const copyFiles = (from:string, to:string) => {}` instead.

src/api/utils/fs_utils.ts
```diff
@@ -31,16 +31,24 @@ export function rimraf(thing: string) {
 }
 
 export function copy(from: string, to: string) {
-	if (!fs.existsSync(from)) return;
-
-	const stats = fs.statSync(from);
-
-	if (stats.isDirectory()) {
+	const copied = new Set();
```
khrome83: Copy is used in a number of different places, but you don't need a `Set()` in any but the one. It would be better to leave the creation of the `Set` outside of copy.

src/api/export.ts
```diff
@@ -98,7 +98,6 @@ async function _export({
 		}, process.env)
 	});
 
-	const seen = new Set();
```
khrome83: Leave the creation of the `Set` here and find a different way to populate it.

jormaj:
I took another stab at it, and this refactor uses a recursive accumulator.
I think this one is the cleanest so far, but happy to hear other opinions :)

src/api/utils/fs_utils.ts
khrome83: It's just common design practice to declare functions before you call them. Similar to how Imports go at the top of the page. Yes, JS can hoist for you. But in general, like all programming languages, it's better to organize it different.

sw-yx:
oh this is very clever! altho i dont think we crawl non html assets anyway?

jormaj:
> oh this is very clever! altho i dont think we crawl non html assets anyway?

Normally, no. But if they're linked to from a page (f.ex. a pdf document in my case), they would be crawled. Hence this PR :)

benmccann:
https://github.com/sveltejs/sapper/pull/1105 / https://github.com/sveltejs/sapper/pull/938 look potentially related

benmccann:

Issue #746, opened 2019-06-17
basepath calculated relative to nesting for IPFS hosting


philholden:
A single site deployment may be hosted on both IPFS and a regular domain name using a gateway service:

https://blog.cloudflare.com/distributed-web-gateway/

However this means the `basepath` need to work both for:

https://mysite.io/about

and:

https://ipfs.io/ipfs/QmbHEhjtfuRrfhDWqDRQqpdX7EnE8ZzByT8t2YXCmfNDBt/about

in this case the following works:

basepath="./" in index.html
basepath="../" in about/index.html
basepath="../../" in blog/article1/index.html

Would it be possible to have a `--relative-basepath` export option that made the basepath of html files relative to their nesting?

philholden:
This worked for me as a post processing step to allow me to host export at root or in a nested directory:

```javascript
//walk.js in project root

const fs = require("fs"),
  path = require("path");
function walkDir(dir, callback) {
  fs.readdirSync(dir).forEach(f => {
    let dirPath = path.join(dir, f);
    let isDirectory = fs.statSync(dirPath).isDirectory();
    isDirectory ? walkDir(dirPath, callback) : callback(path.join(dir, f));
  });
}

walkDir(process.argv[2] || ".", function(filePath) {
  const fileContents = fs.readFileSync(filePath, "utf8");

  if (path.extname(filePath) !== ".html") return;
  const rel =
    "./" +
    filePath
      .replace(/^\//, "")
      .replace(process.argv[2], "")
      .replace(/^\//, "")
      .replace(/[^\/]+/g, "..")
      .replace(/\.\.$/, "");
  const contents = fileContents.replace(
    /<base href=[^>]*>/,
    `<base href="${rel}">`
  );
  fs.writeFileSync(filePath, contents, "utf8");
});
```

```json
// package.json
"scripts": {
  "export": "sapper export && yarn walk",
  "walk": "node walk __sapper__/export"
}
```

wighawag:
That would be great if there was an option indeed. I am also facing the same issue: deploying a website on ipfs while still having the website work in different situations.

@philholden I tried your script but I have still some references to `<link href="/client/..."` where the forward slash make it look up to the root.

I can probably fix it by modyfying the script but a built-in sapper solution would be great

PR #866, opened 2019-09-02
Support for relative base path to allow website to be served from different paths

wighawag:
This PR add code to support relative base path so that a website can be hosted on any sub path, including IPFS gateways

This fixes : https://github.com/sveltejs/sapper/issues/746

Note that all url will finish with "/" as this is how ipfs gateways handle folder path

We could potentially add an option to specify a different behavior (remove the slash at the end) if we want but I ll leave that as a potential addition to be done in another PR.

Note that any attempt to access the website pages without a "/" at the end or by specifying "/index.html", will be redirected to the canonical url (with a "/")
This is achieved via `location.replace`

Note that it won't solve https://github.com/sveltejs/sapper/issues/863 though as filesystem access has its own problem, push state does not work there and this require to have index.html on every path. 
This might be an extra option in the future if we can solve it.

### Tests

get a sapper website from the get started guide
and execute the following
```
sapper export --relative-basepath
```

You can now serve it from any parent/grandparent folder

Or upload it on ipfs

An example can be found here on cloudflare gateway : https://cloudflare-ipfs.com/ipfs/QmNVTWGZ4qoW4DwiTze3icNJKhKBArR4HrCR1sapAFaPYg/

To show how the website can be loaded from different sub path I also added it to a ENS name. You can use a browser that support ENS or you can use the following link : https://wighawag.eth.link 

You can also use a gateway like this : https://cloudflare-ipfs.com/ipns/wighawag.eth.link/

Note that I ll unpin the file and/or update that ENS domain in the future so appologies in advance for future readers :)

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -93,27 +93,59 @@ export function extract_query(search: string) {
 	return query;
 }
 
-export function select_target(url: URL): Target {
+function countBaseDepth(baseUrl: string): number {
+	let depth = 0;
+	const splitted = baseUrl.split('/');
+	for (let split of splitted) {
+		if (split === '..') {
+			depth++;
+		}
+	}
+	return depth;
+}
+
+function trimSlashes(path: string): string {
+	if (path.startsWith('/')) {
+		path = path.slice(1);
+	}
+	if (path.endsWith('/')) {
+		path = path.slice(0, path.length - 1);
+	}
+	return path;
+}
+
+let pathToCut;
+export function select_target(url: URL, start: boolean = false): Target {
 	if (url.origin !== location.origin) return null;
 
 	let path = url.pathname;
+	let pathname = url.pathname;
+
 	if (initial_data.baseUrl.startsWith('.')) {
-		let numBackward = 1;
-		const splitted = initial_data.baseUrl.split('/');
-		for (let split of splitted) {
-			if (split === '..') {
-				numBackward++;
-			}
+		if (start) {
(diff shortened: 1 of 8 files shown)


wighawag:
Did anyone look at the PR ?

I saw few people in the same situation as mine in the discord chat and having sapper website being able to be published on IPFS is I think a great feature that should not be discarded too quickly.

The changes are not invasive neither. I would love to get feedback if things can be improved so we can have that officially supported

Thanks

Clearwood:
I would really appreciate this becoming a part of the package.

benmccann:
We have several PRs dealing with `base` path (https://github.com/sveltejs/sapper/pull/1208#issuecomment-629792510). Would https://github.com/sveltejs/sapper/pull/984 also solve your problem?

benmccann:
@wighawag thanks for your response in https://github.com/sveltejs/sapper/pull/1208#issuecomment-633188671. I'm responding here so as not to go off-topic discussing this PR in the comments of another PR

What I'm wondering is if https://github.com/sveltejs/sapper/pull/984 makes `base` optional, then can you just use relative URLs? Do you need to use `basepath` at all in that case? Using relative URLs instead of relative `basepath` would make the site work in IE

spsaucier-bakkt:
This would be super useful - for instance, I am building for a client who will drop the folder wherever they want, and I have no control over where that will end up. It needs to be able to be dropped in anywhere on their server.

benmccann:

Issue #835, opened 2019-08-01
Preloads in layouts fire too eagerly when dynamic routes are involved


njbotkin:
Discussed in the discord, @Rich-Harris asked me to file an issue.  Repro repo here: https://github.com/njbotkin/sapper-template/tree/dynamic-route-layout-preload-bug

(this is my first time creating a repro repo, let me know if you need more information)

After you clone the repo, here are repro instructions:

Use the nav bar and go to `/one/two/ohai`.  Now navigate to `/one/two`.

Expected console output:

(empty)

Actual console output:

`/one/_layout.svelte preload fired`
`/one/two/_layout.svelte preload fired`

Note that (correctly) no preloads are fired when you navigate from `/one/two` to `/one`

Not terribly severe, worst case scenario, my app issues extra data queries.

Potentially helpful aside: if you remove `return {}` from `preload` in `_layout.svelte`, route preloads are **always** marked as dirty.

njbotkin:
The bug was introduced in 0.27.1, specifically https://github.com/sveltejs/sapper/pull/704, I think.

Conduitry:
#704 is something that there definitely have been some internal discussions about reverting, fwiw.

njbotkin:
Actually, this was introduced by https://github.com/sveltejs/sapper/pull/691

PR #868, opened 2019-09-02
tweak part invalidation logic

njbotkin:
Fixes https://github.com/sveltejs/sapper/issues/835.

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -262,7 +262,7 @@ function part_changed(i, segment, match, stringified_query) {
 	if (!previous) return false;
 	if (segment !== previous.segment) return true;
 	if (previous.match) {
-		if (JSON.stringify(previous.match.slice(1, i + 2)) !== JSON.stringify(match.slice(1, i + 2))) {
+		if (previous.match[0].split('/').slice(i+1, i+2)[0] !== match[0].split('/').slice(i+1, i+2)[0]) {
 			return true;
 		}
 	}


njbotkin:
Oh, hmm.  It seems that the tests added in https://github.com/sveltejs/sapper/pull/691 actually enforce the overeager invalidation behaviour described in https://github.com/sveltejs/sapper/issues/835.  I would argue that `h2` should *not* be updated in these cases (https://github.com/sveltejs/sapper/blob/master/test/apps/basics/test.ts#L292-L308).

runtime/src/app/app.ts
```diff
@@ -262,7 +262,7 @@ function part_changed(i, segment, match, stringified_query) {
 	if (!previous) return false;
 	if (segment !== previous.segment) return true;
 	if (previous.match) {
-		if (JSON.stringify(previous.match.slice(1, i + 2)) !== JSON.stringify(match.slice(1, i + 2))) {
+		if (previous.match[0].split('/').slice(i+1, i+2)[0] !== match[0].split('/').slice(i+1, i+2)[0]) {
```
benmccann, line 265: a space should be maintained on either side of the `+`. we've been playing around with setting up `eslint` to enforce style rules

benmccann:
I'm not sure if you have any interest in pursuing this PR still? I hadn't heard anything for awhile. We've moved our CI from Travis to Github, so it'd need to be rebased at a minimum. But maybe better just to try to make sure it works in SvelteKit at this point?

benmccann:

Issue #851, opened 2019-08-20
Memory leak in sapper export


nealalpert:
**Description:**
Sapper export's recursive fetching of URLs can exceed node's write capacity resulting in very large memory footprint.

The [handle function's](https://github.com/sveltejs/sapper/blob/master/src/api/export.ts#L138) recursive page crawling and URL fetch calls can end up overwhelming the [save function's](https://github.com/sveltejs/sapper/blob/master/src/api/export.ts#L105) fs.writeFileSync call's capacity to write to disk resulting in the yootil's queue memory footprint to exceed node's allowed amount.

**Steps To Reproduce:**
Run `sapper export --legacy` with a site with 30,000 pages where some of those pages are driven by sizable amounts of data.  Our current structure looks a bit like:

```
.
└── routes
    └── [lang]
        ├── things
        │   ├── [slug]
        │   │   ├── prices
        │   │   │   ├── _prices.js
        │   │   │   ├── index.json.js
        │   │   │   ├── index.svelte
        │   │   │   └── shop.json.js
        │   │   ├── index.json.js
        │   │   └── index.svelte
        │   ├── _validThings.js
        │   ├── index.json.js
        │   └── index.svelte
        ├── _error.svelte
        ├── _layout.js
        ├── _layout.svelte
        ├── layout.json.js
        └── index.svelte
```

I can't set up a repo with an example at the moment but I'll update this ticket once I can come up with an example that doesn't expose my company's data.

**Expected behavior**
Perhaps throttle the page crawling if the process is at risk of running out of memory?

**Version Info:**
Sapper 0.27.8
Svelte 3.8.0
Rollup 1.19.4

**Severity**
Moderate.  A work around is possible by passing the `--max-old-space-size` optional param to node and upping the memory allotment  when running `sapper export` if the environment can spare it.

**Stacktrace:**
<details>
<summary>JS stack trace</summary>
==== JS stack trace =========================================

Security context: 0x301abcea57c1 <JSObject>
[...]

PR #869, opened 2019-09-03
Handle i/o backpressure with sapper export

nealalpert:
Rewrite of sapper export to reduce memory footprint.  Related to issue #851.

Adds a queue that allows for any number of URLs to be queued up but limits the amount of data in memory based off of the [concurrent](https://github.com/sveltejs/sapper/blob/master/src/cli.ts#L198) export parameter.

In our test case a site with 69000 pages sapper export took 17 minutes and peaked memory usage around 3.8gb.  After the queue change the export took 13 minutes and peaked memory usage around 600mb.  In a use case where the export process is bottlenecked by the text parsing in the [handle](https://github.com/sveltejs/sapper/blob/master/src/api/export.ts#L138) function or file writing in the [save](https://github.com/sveltejs/sapper/blob/master/src/api/export.ts#L105) function this pattern offers performance gains and reduction in memory usage.

This pattern offered slight performance benefits over making the page export process (fetch, parse urls and file save) a single promise in the current queue.


khrome83:
@Conduitry - can we get this PR looked at?

The same unit test for this in AppVeyor seems to be failing for others as well. Running PR earlier did not trigger that failure, and we can not reproduce locally. From what we see it does not seem to be a valid failure caused by this code, as it only impacts `export` commands.

nealalpert:
updated the branch with a fix for the race condition brought up in https://github.com/sveltejs/sapper/issues/893

MaxMilton:
Would love to see some movement with this one. The issue in #893 is a blocker which this PR fixes. I'd rather not need to maintain a sapper fork just for fixes.

mhkeller:
@Conduitry would it be possible to get a review of this? I currently have to do exports in somewhat complicated batches to make sure I get all the posts.

khrome83:
Bumb...

src/api/export.ts
```diff
@@ -3,8 +3,8 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as url from 'url';
 import fetch from 'node-fetch';
-import * as yootils from 'yootils';
```
benmccann, line 6: this looks to be unused now and could be removed from `package.json`

jlkiri:
Any updates?

antony:
I've added the suggestion from @sw-yx in #864, as well as @benmccann 's suggestion.

I ran this PR with the existing export, the original PR, and my updated PR, and whilst non scientific, the original PR was 2x the speed of the original, and the updated PR seems to have shaved another 20% off (though this could be due to API warming up etc)

Either way this gets a :+1: from me.

Issue #489, opened 2018-10-23
Base url for server fetch requests


imbolc:
Now `fetch` used with an relative path requests different urls from browser and server. Eg if we're on page `foo.com` and requesting `/api/bar.json`, browser would request `foo.com/api/bar.json`, but server `127.0.0.1:PORT/api/bar.json`.

In my case (and I think it's a pretty common one) we already have api written on different language, so I'm using nginx to join it with sapper under the same domain name. But as api actually serves on a different port than sapper does, `fetch` from server side doesn't work. It requests `127.0.0.1:SAPPER_PORT` while api is on `127.0.0.1:API_PORT`.

I have `proxy_set_header Host $host;` in my nginx config and Express parses it into `req.hostname`. But as @mrkishi suggested we probably can't use it in `fetch` by default as anyone can forge this header and trigger server requests to any domains.

So it may be a setting to enable parsing of the header or just an `FETCH_ORIGIN=https://foo.com` environment variable. I'd vote for the second option as it doesn't rely on proper proxy server configuration.

stalkerg:
I have a similar problem. 
If you worry about security you can provide the same option as exist in express:
```js
app.set('trust proxy', 'loopback')
```

benmccann:
I'd like this as well. I've written a backend in Java and would like to send my fetch calls for JSON data to that server (e.g. api.mydomain.com or mydomain.com/api)

@Conduitry I see you marked this as "pending clarification". Is there anything in particular that you think needs to be clarified?

benmccann:
Fetching is done on both the client side and server side.

**Client side**

https://github.com/sveltejs/sapper/blob/fffdbc6282722083a4c750b20b9e7e8104fd8e47/runtime/src/app/app.ts#L284

For the client side, we could modify `start({ target })` to be `start({ target, apiBaseUrl })`.

`target` is passed to `set_target`. We could possibly rename this to `set_options` and pass all the options. Either way, this is done in `/runtime/src/app/app.ts`, so the new option would be available to `fetch`

**Server side**

https://github.com/sveltejs/sapper/blob/fffdbc6282722083a4c750b20b9e7e8104fd8e47/runtime/src/server/middleware/get_page_handler.ts#L107

For the server-side, I think we could similarly add an option to the `middleware` options. It'd then be straightforward to pass it on to `get_page_handler` in `server.mjs`

benmccann:
[...]

PR #882, opened 2019-09-09
Fix server fetch behaviour

### LINKED ISSUE #880: Authorization header doesn't get (automatically) sent by server fetch
**Describe the bug**
The server middleware doesn't add in the page request's Authorization header when fetching the same domain. This means that a site relying on the Authorization header will work when the client uses fetch, but will break when doing SSR on a page.

**To Reproduce**
Here's an example using `express` and `express-basic-auth` with the sapper template.
In `src/server.js`:
```js
import sirv from 'sirv';
import express from 'express';
import compression from 'compression';
import basicAuth from 'express-basic-auth';

const { PORT, NODE_ENV } = process.env;
const dev = NODE_ENV === 'development';

const staticUserAuth = basicAuth({
  users: {
    'username': 'password'
  },
  challenge: true
})

express() // You can also use Express
  .use(
    sirv('static', { dev }),
    staticUserAuth,
    sapper.middleware()
  )
  .listen(PORT, err => {
  });
```
In `src/routes/blog/[slug].svelte`, change line 5:
```js
const res = await this.fetch(`blog/${params.slug}.json`, { credentials: 'same-origin' });
```

You may also need to remove the service worker to stop it from interfering with the authentication process. The repo I used can be found [here](https://github.com/cyderize/sapper-auth).

After logging in with the username `username` and the password `password`, if you click the hyperlink to the blog page, the content loads as expected because the browser adds the Authorization header because the request is for the same domain.

However, if you refresh `/blog` to perform ssr then it will display a 500 error because the server does not send an Authorization header to `blog.json`.

**Expected behavior**
The server middleware should send the Authorization header when the domain of its initial request matches the domain of the URL being fetched.

**Additional context**
I'm developing a web app using Firebase, and making use of the service worker to [inject an Authorization header](https://firebase.google.com/docs/auth/web/service-worker-sessions) with the logged in user's authentication ID Token. This then gets verified by some middleware on the server. This approach means I don't have to deal with session cookies and keeping them in sync with Firebase auth and also negates the need for CSRF tokens. However, the server throws away the Authorization header at the moment, making it impossible for me to use ssr for authenticated routes.

### LINKED ISSUE #881: Server fetch doesn't use same-origin by default
I noticed while poking around the server middleware that when fetching, Sapper doesn't seem to send any credentials by default (i.e. credentials: 'omit'). Since whatwg/fetch#585, however, the default in browsers is to have credentials: 'same-origin'.

Is there a reason why this is the case, or should this be changed to be consistent with what browsers now do?

cyderize:
Sets `{ credentials: 'same-origin' }` as default for server fetch(). Closes #881.

When including credentials, also send authorization header. Closes #880.


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -106,38 +106,42 @@ export function get_page_handler(
...
-							.join('; ');
+					opts.headers.cookie = str;
 
-						opts.headers.cookie = str;
+					if (!opts.headers.authorization && req.headers.authorization) {
+						opts.headers.authorization = req.headers.authorization;
 					}
 				}
-
+
```
benmccann, line 144: nit: don't need to add the whitespace here

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -106,38 +106,42 @@ export function get_page_handler(
...
-				if (opts) {
-					opts = Object.assign({}, opts);
-
-					const include_cookies = (
+				const includeCredentials = (
+					opts.credentials !== 'omit' && (
 						opts.credentials === 'include' ||
-						opts.credentials === 'same-origin' && parsed.origin === `http://127.0.0.1:${process.env.PORT}`
-					);
+						parsed.origin === `http://127.0.0.1:${process.env.PORT}`
```
benmccann, line 115: it might be nice to encapsulate this in a method like `isSameOrigin` so that in the future we can more easily make the URL to fetch from configurable. It also would help better explain the logic here

benmccann: Actually, isn't this a no-op?  `parsed` was hardcoded to this above, right? So you could probably simplify by just making this check be `opts.credentials !== 'omit'`

EddyVinck:
I'm having trouble with getting my cookies during SSR as well in the boilerplate I'm working on https://github.com/EddyVinck/sapper-graphql-template/pull/20

I think this issue might solve my problem, so I will be checking this out after work or at the very least later this week.

Let me know if there's anything I can do to help.

runtime/src/server/middleware/get_page_handler.ts
Conduitry: No, `parsed` is constructed from `url`, relative to `http://127.0.0.1/...` if it did not contain a domain.

benmccann: oops. sorry. accidentally deleted my comment that @Conduitry was replying to, so his comment might not make sense without that context. in any case, I was mistaken so you can ignore this thread

EddyVinck:
I just checked this PR out locally and linked it with `yarn link` to try it out for my project and it fixes my problem! I'm so happy :smile:

EddyVinck:
Actually I was mistaken, the behavior seems to be the same for me. I am using [`svelte-apollo`](https://github.com/timhall/svelte-apollo), so maybe there is a problem there as well. Seems it doesn't include the cookie on the initial request (SSR)

In case it might be of interest for this PR you can check out what I'm working on here: https://github.com/EddyVinck/sapper-graphql-template/pull/20

benmccann:
@EddyVinck I just took a look at [your SO post](https://stackoverflow.com/questions/60892831/how-to-include-credentials-in-fetch-preload-with-svelte-sapper). This PR only helps when your frontend and backend are on the same domain. For querying across domains you will need another solution. I've written about that here: https://github.com/sveltejs/sapper/issues/489#issuecomment-628369826. tl;dr: I would recommend using the same domain and using something like nginx to route to the appropriate server based on URL path

EddyVinck:
@benmccann thanks for your help.

I just tried it with `this.fetch` and it didn't solve my problem. My middleware did not give me my cookie on SSR, even with `this.fetch`.
<details>
<summary>in async preload function</summary>

```javascript
let result2 = await this.fetch("/graphql", {
  method: "POST",
  headers: {
    "Content-Type": "application/json"
  },
  body: JSON.stringify({ query: "{ me { id } }" })
});
result2 = await result2.json();
console.log({ thisFetch: result2.data });
```
</details>

I don't want to hijack this PR with my problems, but I do have one more question if you don't mind. When you say that the domain should be different, should it not be working if my frontend is on `http://localhost:3000/` and my GraphQL endpoint is on `http://localhost:3000/graphql`?

benmccann:
That case should work with this PR, or you can do `{credentials: 'include'}` as a workaround in the meantime. You can find me on discord with the same username to avoid too much discussion via PR comments

EddyVinck:
Oh thanks, I forgot that. That was indeed the issue!

benmccann:
@Conduitry this PR looks good to me. What do you think?

test/apps/credentials/test.ts
```diff
@@ -11,35 +11,79 @@ describe('credentials', function() {
...
+	it('sends cookies to same origin when using this.fetch without credentials', async () => {
+		await r.load('/credentials?type=cookie');
+
+		assert.equal(
+			await r.text('h1'),
+			'a: 1, b: 2, max-age: undefined'
+		);
+	});
+
+	// TODO: does not send cookies to different origin with credentials: "same-origin"
```
Conduitry, line 49: This comment (and the other TODO below) are making me nervous. We definitely do not want to be sending cookies or authorization headers to external servers during SSR by default. This isn't an acceptable known limitation to be addressed later on.

The code in `get_page_handler` doesn't look like it would do that, but I haven't actually tested it. This behavior would need to be fixed (if it is indeed broken), and the tests would need to be updated.

benmccann: I think the TODO is a bit poorly worded. My understanding of the TODO is that it was a TODO to add a test to assert that it does not do that. This PR doesn't actually send credentials to other origins unless the user specifically sets `opts.credentials === 'include'`

I think that the best course of action here is to just drop these TODOs while merging. I'm not sure that adding such tests are necessary for a couple reasons. First is that the current code does not have such tests. Second is that Sapper's tests are integration tests and it doesn't seem very straightforward to setup a another server on another origin and assert that it receives a request without credentials. Though if you do think there's some change that would be useful here, I'm happy to open up a new version of this PR with those changes.

Conduitry: Thanks. No I don't think better tests should block this PR. I appreciate the sanity check and the confirmation that we're not leaving a big security hole as a TODO. The updating of the tests I was asking for can be as simple as fixing the comments.

Issue #885, opened 2019-09-10
[Request] Document CLI flag configuration options


sw-yx:
**Is your feature request related to a problem? Please describe.**

there is no place on the docs to document notes for each of the CLI flags. there are places for `sapper build` and `sapper export` but none for `sapper dev` and no good way to document flags that are just common to all of them. this argues for a distinct "CLI API" section.

Example issues that could be addressed:

- `--ext` flag: https://github.com/sveltejs/sapper/issues/827
- `--legacy`: https://github.com/sveltejs/sapper/issues/739
- `ignore`: https://github.com/sveltejs/sapper/issues/736
- `--entry`

**Describe the solution you'd like**

this argues for a distinct "CLI API" section.

**Describe alternatives you've considered**

misery

**How important is this feature to you?**
low importance but easy to do

pngwn:
I think this is a good idea, we don't have a great place to document these currently, and the CLI args have grown somewhat recently.

I think a "CLI API" section with links from various sections down to the CLI docs makes the most sense (so that users reading specific sections don't miss out on anything if they aren't in the habit of `ctrl-f`-ing).

I think it makes sense to have these grouped in a section but am wary of people who just read the 'relevant' not realising that the solution to their problem is a CLI argument that they may not even know exists. A link in the relevant section may go some way to solving this.

- A link in the "Exporting" section to the CLI args (with a quick note about `--entry`).
- A link in the "Routing - Pages" section to the CLI args (with a quick note about `--ext`)

`legacy` doesn't really have a natural home to link from but we should add details about browser compatibility (or at least feature compatibility) somewhere in the Sapper docs. This is an important point for many people and comes up relatively often.

`ignore` isn't a CLI argument, so I'm not sure what should happen with it.

sw-yx:
@pngwn please, you look
![image](https://user-images.githubusercontent.com/6764957/64629827-ec698d00-d3c1-11e9-8c37-b65cc730b59d.png)

PR #888, opened 2019-09-10
add CLI API docs

sw-yx:
Closes https://github.com/sveltejs/sapper/issues/885

I used my best judgment as to where in the docs the CLI API docs should go, so i had to rename a bunch of files, but I'm happy to respond to feedback.

cc @pngwn let me know if this looks reasonable


site/content/docs/04-cli-api.md
```diff
@@ -0,0 +1,46 @@
...
+Lastly, `sapper export` inherits the same commands as `sapper build`, plus these specific ones:
+
+- `build_dir` (`string`): set build dir. Default `__sapper__/build`
+- `export_dir` (`string`): set export dir. Default `__sapper__/export`.
+- `host_header` (`string`): set host header
+- `timeout` (`number | false`): set timeout for prefetches. Default `5000`.
+- `concurrent` (`number`): set number of concurrent builds. Default `8`.
+- `oninfo` (`({ message }: { message: string }) => void;`): callback
+- `onfile` (`({ file, size, status }: { file: string, size: number, status: number }) => void;`): callback
+- `entry` (`string`): Set entry point of your static site, in case it is hosted at a subdirectory. Default `/`
```
pngwn: `entry` also accepts a space-separated string as you can define multiple entry points, probably worth clarifying this.

pngwn:
This looks good to me, the only thing I would add is that any space-separated values need to be wrapped in quotes `--entry "foo bar"` as this is a common point of confusion.

@Conduitry Does this look okay to you?

Conduitry:
I guess I find it a bit weird to mention there's a programmatic API and then to say basically nothing about it besides 'look at the source'. I'd probably prefer to not mention it at all. We get lots of questions that would be already answered if people knew about the command line options to the CLI - but I would bet that almost no one is using the API or could benefit from it. Maybe this could be revisited later, but I think for now only describe the CLI.

site/content/docs/04-cli-api.md
```diff
@@ -0,0 +1,46 @@
...
+
+Sapper ships a small CLI that you can use to build, serve, and export your Sapper apps:
+
+- `sapper dev`: spin up a local dev server locally with hot reloading
+- `sapper build`: build a Sapper app that can be deployed on a Node server
+- `sapper export`: export a Sapper app as a static site, prefetching data
+
+You may call these via API or programmatically by importing the API (check the Sapper source for details on this).
+
+In the CLI, you can pass flags to each command, e.g. `sapper dev --legacy --ext ".svelte .svexy"`. Here is a full list of common API flags/options and what they do:
```
Conduitry: I do love mdsvex's default file extension, but this feels a little unprofessional lol. Maybe a different example?

sw-yx: will use `.html`

site/content/docs/04-cli-api.md
```diff
@@ -0,0 +1,46 @@
...
+- `cwd` (`string`): the current working directory. Default `.`
+- `src` (`string`): the source to build from. Default `src`
+- `dest` (`string`): the folder to build to. Default `__sapper__/build`
+- `routes` (`string`): the routes folder to build from. Default `src/routes`
+- `output` (`string`): the output folder to build to. Default `src/node_modules/@sapper`
+- `static` (`string`): the output folder for static files. Default `static`
+- `bundler`: can be either `rollup` or `webpack`. Default: automatically determined
+- `ext` (`string`): space separated string of file extensions to read, e.g. `--ext ".svelte .svexy"`. Default: `.svelte`. Remember you will also have to configure your webpack/rollup client and server `extensions` accordingly.
+
+Commands specific to `sapper dev`:
```
Conduitry: Should probably say 'options' and not 'commands'. Ditto below.

sw-yx:
@Conduitry one reason to document it is the callbacks.. should I just leave them off?

site/content/docs/04-cli-api.md
pngwn: I beg your very pardon!

site/content/docs/04-cli-api.md
```diff
@@ -0,0 +1,46 @@
...
+
+- `dev-port` (`number`): set the port for dev
+- `devtools-port` (`number`): set the port for devtools
+- `port` (`number`): set yet another port. Default `process.env.PORT`
+- `hot` (`boolean`): toggle hot reloading
+- `live` (`boolean`): toggle live
+
+Commands specific to `sapper build`:
+
+- `oncompile` (`({ type, result }: { type: string, result: CompileResult }) => void;`): callback
```
Conduitry: Mentioning an API callback as an option to `sapper build` (a CLI command) seems confusing. As I mentioned in another comment, it might be best to leave out the API stuff entirely here. Ditto below.

sw-yx: alright

site/content/docs/04-cli-api.md
sw-yx: ok

sw-yx:
all comments addressed!

sw-yx:
has a bunch of conflicts now and i'm giving up

Issue #890, opened 2019-09-13
sapper export `--no-timeout` does not disable timeout


swyxio:
**Describe the bug**

this is what I thought the bug was

1. in the CLI we tell people to [pass `--no-timeout` to disable](https://github.com/sveltejs/sapper/blob/910d28e3419409b5498933e8c13d37cc03b652a0/src/cli.ts#L199)
2. however that just gives us a `opts.timeout` value of `false`
3. this is [passed into setTimeout](https://github.com/sveltejs/sapper/blob/910d28e3419409b5498933e8c13d37cc03b652a0/src/api/export.ts#L151) and times out immediately

however i tried [this](https://github.com/sveltejs/sapper/pull/891) and STILL got timeouts... which means there is another layer of timeouts and i dont know where it is

**Logs**
n/a

**To Reproduce**

`sapper export --no-timeout` with a lot of pages.

**Expected behavior**

it should continue exporting with no timeout

**Severity**
How severe an issue is this bug to you? Is this annoying, blocking some users, blocking an upgrade or blocking your usage of Sapper entirely?

moderate severity. obv not having a timeout is not ideal but a  clear bug in the cli interface is also not good

PR #891, opened 2019-09-13
fix timeout bug

sw-yx:
fix timeout bug

closes https://github.com/sveltejs/sapper/issues/890

### DIFF
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -146,9 +146,10 @@ async function _export({
 
 		const r = await q.add(async () => {
 			const timeout_deferred = new Deferred();
+			
 			const the_timeout = setTimeout(() => {
 				timeout_deferred.reject(new Error(`Timed out waiting for ${url.href}`));
-			}, timeout);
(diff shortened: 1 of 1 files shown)


src/api/export.ts
```diff
@@ -146,9 +146,10 @@ async function _export({
 
 		const r = await q.add(async () => {
 			const timeout_deferred = new Deferred();
+			
 			const the_timeout = setTimeout(() => {
 				timeout_deferred.reject(new Error(`Timed out waiting for ${url.href}`));
-			}, timeout);
+			}, timeout === false ? undefined : timeout);
```
Conduitry: Shouldn't you just be not creating the timeout at all if `timeout === false`? I'd expect calling `setTimeout(..., undefined);` would have the same problem of it being called right away.

sw-yx:
on further inspection there is ANOTHER timeout at play that is not working here... so i am closing this and amending my issue as i havent nailed down what the cause is

PR #907, opened 2019-09-20
add support for url with suffix of index.html in export mode

shuizhongyueming:
I have uploaded my exported site to a third side cloud storage to make a static site.

But the cloud storage will automatically transfer url `/some/path` to `/some/path/index.html`

The HTML content is loaded correctly, but the script for that route is only loaded in URL `/some/path` but not the `/some/path/index.html`, even they are the same.

So, I change the `select_target` function, add 

```javascript
	path = path.replace(/index\.html$/, '');
```

before the match of path and route.

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -99,6 +99,9 @@ export function select_target(url: URL): Target {
 
 	let path = url.pathname.slice(initial_data.baseUrl.length);
 
+	// support url with suffix index.html in export mode
+	path = path.replace(/index\.html$/, '');
+
 	if (path === '') {
 		path = '/';
 	}


benmccann:
This line might be related?

https://github.com/sveltejs/sapper/blob/4a0ffd1b10604812657d8842058e9bc24e516134/src/api/export.ts#L40

shuizhongyueming:
@benmccann 

` return path.replace(/^\/|\/$|\/*index(.html)*$|.html$/g, '') `

I think this is a better regx for remove some unnecessary suffix

benmccann:
Sorry this had gone unreviewed for so long. This one has a merge conflict. Maybe better to add this feature to SvelteKit at this point than trying to get it in here?

PR #921, opened 2019-09-30
prefetch on focus

### LINKED ISSUE #900: Shouldn't prefetch trigger on focus event as well?
Currently the prefecth feature [triggers](https://github.com/sveltejs/sapper/blob/master/runtime/src/app/start/index.ts#L29) on hover and touch:

```javascript
addEventListener('touchstart', trigger_prefetch);
addEventListener('mousemove', handle_mousemove);
```

I propose it should also be invoked on `focus` to provide the feature to users using keyboard navigation.

Would it cause any issues?

mdauner:
This PR adds an event listener in order to prefetch links on focus. Closes #900.

### DIFF
--- a/runtime/src/app/start/index.ts
+++ b/runtime/src/app/start/index.ts
@@ -28,6 +28,7 @@ export default function start(opts: {
 
 	// prefetch
 	addEventListener('touchstart', trigger_prefetch);
+	addEventListener('focusin', trigger_prefetch);
 	addEventListener('mousemove', handle_mousemove);
 
 	return Promise.resolve().then(() => {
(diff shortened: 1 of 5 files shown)


benmccann:
Sorry this had gone unreviewed for so long. It seems like a decent idea. It's got some merge conflicts, so I can't just hit merge on it. This code exists in SvelteKit as well. I'd hate to add new features to Sapper that don't get added in SvelteKit. If we're going to add it to just one place I'd rather add it there. What do you think?

https://github.com/sveltejs/kit/blob/84e9023c06dc3579972f93e0d5cf1f2481aa3be4/packages/kit/src/runtime/client/renderer.js#L101

PR #938, opened 2019-10-15
Sapper export also follows <img src='...'/> and exports these

grjzwaan:
This PR is not related to any issue that I could find, I just wanted something like this and made a PR.

It might not be the cleanest way to solve this issue, or even desired, so I welcome feedback.

### Current behaviour
Sapper export follows `<a href='>` links, but not ``. 

### New behaviour
Sapper export follows `` links and exports the image.

### Changes

* Also follow the url  in `src` attributes;
* When fetching results, determine based on the mime to use `r.text()` or `r.arrayBuffer()`;
* Write files using a buffer (then writing the file works the same for text or arrayBuffer);
* Added a test

### Why 
For myself I have a static website with blog posts in the following structure:
```
/blog/2019/11/01/index.md
/blog/2019/10/12/index.md
/blog/2019/10/12/catlolz.png
...
```
The `index.md` file refers to the image. I could put the image in the `static` folder and refer to it but the image _belongs_ to this post. It seems nicer to also intercept requests for images in the export.

You do need an endpoint that serves the images, but this is also an advantage because now we can optimize those images dynamically (resize, compress, etc) by giving information from the front end.

### Points of attention
You probably do not want to wildly grab all image links on your website. I **think** there was a filter in Sapper to exclude files from outside your own website, so that should apply. I didn't test this yet.


aloker:
Sapper should also follow image links in <source srcset/> and .

aloker:
I took the liberty to implement a comparable solution myself, including srcset support (#1105)

Conduitry:
Closing in favor of #1398, which is built on this. I've included you as a co-authored-by on the commit. Thank you!

Issue #923, opened 2019-10-03
req.originalUrl not provided from vanilla HTTP server.


devgar:
**Describe the bug**
`sapper.middleware` needs `req.originalUrl` for routing, this property is defined by **polka**/**express** but undefined in **vanilla** HTTP Request object.

**Logs**
```
webpack:///./src/node_modules/@sapper/server.mjs?:2554
				if (req.url === '/' && originalUrl[originalUrl.length - 1] !== '/') {
				                                               ^

TypeError: Cannot read property 'length' of undefined
    at Array.eval (webpack:///./src/node_modules/@sapper/server.mjs?:2554:52)
    at nth_handler (webpack:///./src/node_modules/@sapper/server.mjs?:2609:14)
    at eval (webpack:///./src/node_modules/@sapper/server.mjs?:2613:25)
    at eval (webpack:///./src/server.js?:27:4)
    at cc (/Users/ed/Documents/Projects.nosync/Webs/Svelte/Sapper/magician/node_modules/sirv/index.js:93:18)
    at eval (webpack:///./src/server.js?:26:3)
    at compression (/Users/ed/Documents/Projects.nosync/Webs/Svelte/Sapper/magician/node_modules/compression/index.js:220:5)
    at Server.eval (webpack:///./src/server.js?:25:2)
    at Server.emit (events.js:196:13)
    at parserOnIncoming (_http_server.js:708:12)
```

**To Reproduce**
Changing default **server.js** in sample-template with this:

```
import sirv from 'sirv';
import {createServer} from 'http';
import compression from 'compression';
import * as sapper from '@sapper/server';

const { PORT, NODE_ENV } = process.env;
const dev = NODE_ENV === 'development';
const compressionCB = compression({ threshold: 0 })
const sirvCB = sirv('static', { dev })
const sapperCB = sapper.middleware()

createServer((req, res) => {
	compressionCB(req, res, () =>
		sirvCB(req, res, () =>
			sapperCB(req, res)
		)
	)
}).listen(PORT)
```

**Information about your Sapper Installation:**

- Using the sapper-template

- This occurs in Webpack and Rollup

- Sapper version 0.27.9

- Svelte version 3.12.1

**Severity**
This only occurs in non polka/express apps which are the middlewares sapper focuses in, but I think a vanilla HTTP support should be taken into account.
It can be fastly solved adding ```req.originalUrl = req.url``` before calling the middleware stack.

PR #942, opened 2019-10-17
Define req.originalUrl if not defined when needed.

devgar:
This enables sapper to be used with nodejs HTTP module as mentioned in #923 .

### Tests

### DIFF
--- a/runtime/src/server/middleware/index.ts
+++ b/runtime/src/server/middleware/index.ts
@@ -17,6 +17,9 @@ export default function middleware(opts: {
 	return compose_handlers(ignore, [
 		(req: Req, res: Res, next: () => void) => {
 			if (req.baseUrl === undefined) {
+				if (!req.originalUrl) {
+					req.originalUrl = req.url
+				}
 				let { originalUrl } = req;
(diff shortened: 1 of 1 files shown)


runtime/src/server/middleware/index.ts
```diff
@@ -17,6 +17,9 @@ export default function middleware(opts: {
 	return compose_handlers(ignore, [
 		(req: Req, res: Res, next: () => void) => {
 			if (req.baseUrl === undefined) {
+				if (!req.originalUrl) {
+					req.originalUrl = req.url
+				}
 				let { originalUrl } = req;
```
benmccann, line 23: @Rich-Harris was proposing the same change in https://github.com/sveltejs/sapper/pull/682/files#r432806606. This seems pretty reasonable to me. I think his way of setting it was a bit more succinct though. What do you think about updating it to that?

devgar: Yes, this was only a proposition to solve #923, it's clearer and could be safer.
Should I close the pull or update it to match his changes?

PR #944, opened 2019-10-18
remove references to 'now'

antony:
Remove all references to `now` as they relate to `v1`, we should have some considered instructions about deploying to `v2`, however until route splitting is done, this probably isn't ideal.

The main reason is that new users to now can't even spin up v1 apps, as it won't even appear as an option for them - only for people like me who have been `grandfathered` in.

Now v1 is due to be officially deprecated *very* soon I've been told, and the last week has taught me that they've essentially abandoned the platform, and it's a struggle to get anything to deploy there recently. It's probably not a great idea to endorse it in the docs for this reason.

### DIFF
--- a/site/content/docs/11-deploying.md
+++ b/site/content/docs/11-deploying.md
@@ -4,22 +4,6 @@ title: Deployment
 
 Sapper apps run anywhere that supports Node 8 or higher.
 
-
-### Deploying to Now
-
-> This section relates to Now 1, not Now 2
-
-We can very easily deploy our apps to [Now][]:
-
-```bash
-npm install -g now
-now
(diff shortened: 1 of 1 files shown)


sw-yx:
render.com is probably the new best place

antony:
@sw-yx I was complaining to support that I hadn't been able to deploy any of my applications in 5 days, and they said "you should upgrade to v2, as we will be deprecating v1 imminently"

antony:
Closing in favour of merging https://github.com/sveltejs/sapper/pull/979/files and then removing the now v1 information and adding a new PR for adding Docker, Netlify.

PR #945, opened 2019-10-18
Close event source before window is unloaded

klaussner:
If Sapper is run in development mode, Firefox outputs a (meaningless) error in the console every time the page is reloaded manually or after a reload action was received by the dev client's `EventSource`:

sapper-firefox-console

This seems to be a bug in Firefox because the error doesn't occur in Chrome or Safari. I've added an event listener that closes the event source before the window is unloaded, which prevents the error.

### DIFF
--- a/sapper-dev-client.js
+++ b/sapper-dev-client.js
@@ -37,4 +37,11 @@ export function connect(port) {
 			check();
 		}
 	};
+
+	// Close the event source before the window is unloaded to prevent an error
+	// ("The connection was interrupted while the page was loading.") in Firefox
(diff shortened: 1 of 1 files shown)


thgh:
Fixes https://github.com/sveltejs/sapper/issues/981

ryanatkn:
Just wanted to confirm for any reviewers that I was experiencing this issue and had an identical fix.

grundmanise:
The same is observed in Chrome when the connection is not explicitly closed.
`Version 83.0.4103.61 (Official Build) (64-bit)`

PR #953, opened 2019-10-21
cli: add --silent option to sapper build

marinintim:
This PR adds --silent options to `sapper build`. The motivation is that `rushjs` build system treats any output as warnings, thus failing the build. This is similar to `rollup --silent` option.

I didn't pass this option to bundler itself (as I've added `silent` to my rollup config` and wanted to avoid passing option three levels deep).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -155,6 +155,7 @@ prog.command('build [dest]')
 	.option('--routes', 'Routes directory', 'src/routes')
 	.option('--output', 'Sapper intermediate file output directory', 'src/node_modules/@sapper')
 	.option('--ext', 'Custom page route extensions (space separated)', '.svelte .html')
+	.option('--silent', 'Do not log on zero errors')
 	.example(`build custom-dir -p 4567`)
 	.action(async (dest = '__sapper__/build', opts: {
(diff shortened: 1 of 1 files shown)


cfey:
Hi,
I got a similar issue.
Why isn't this merged yet? It seems like a really small addition which doesn't break anything

schnubor:
Why isn't this merged after more than a year? T_T

antony:
I'm not sure this feature is necessary. It is adding more cli options to do something which is built into any decent shell.

`npm run build  > /dev/null`

See https://www.baeldung.com/linux/silencing-bash-output

This even works on Windows, if this article is to be believed

https://stackoverflow.com/questions/1420965/redirect-windows-cmd-stdout-and-stderr-to-a-single-file

benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, [that's been addressed in SvelteKit](https://kit.svelte.dev/docs#configuration). I'm going to go ahead and close this PR since we won't be adding the feature to Sapper, but if you'd like to check out SvelteKit we will be reviewing PRs for features there

PR #954, opened 2019-10-22
Add example for importing client-side library

mateomorris:
Maybe I'm just a dummy, but it took me a while to figure out that this would let me use a client-side library and avoid the `document is not defined` errors. Let me know if I'm off on anything 😄

### Tests

### DIFF
--- a/site/content/docs/06-server-side-rendering.md
+++ b/site/content/docs/06-server-side-rendering.md
@@ -6,56 +6,29 @@ Sapper, by default, renders server-side first (SSR), and then re-mounts any dyna
 
 ### Making a component SSR compatible
 
-Sapper works well with most third-party libraries you are likely to come across. However, sometimes, a third-party library comes bundled in a way which allows it to work with multiple different module loaders. Sometimes, this code creates a dependency on `window`, such as checking for the existence of `window.global` might do.
+Sapper works well with most third-party libraries you are likely to come across. However, sometimes a third-party library depends on browser objects like `window` or `document`.
 
-Since there is no `window` in a server-side environment like Sapper's, the action of simply importing such a module can cause the import to fail, and terminate the Sapper's server with an error such as:
+Since there are no browser objects in a server-side environment like Sapper's, importing a module dependant on them can cause the import to fail and terminate the Sapper's server with an error such as:
 
 ```bash
 ReferenceError: window is not defined
 ```
 
-The way to get around this is to use a dynamic import for your component, from within the `onMount` function (which is only called on the client), so that your import code is never called on the server.
+The way to get around this is to dynamically import the library from within the `onMount` function (which is only called on the client) so that the import is never called on the server.
+
+For example, importing and mounting [Quill.js](https://github.com/quilljs/quill/)
 
 ```html
 <script>
-	import { onMount } from 'svelte';
-
-	let MyComponent;
-
-	onMount(async () => {
-		const module = await import('my-non-ssr-component');
-		MyComponent = module.default;
-	});
+	import {onMount} from 'svelte'
+        let editor
+	
+	onMount(async() => {
+		const { default: Quill } = await import('quill')
+		new Quill(editor)
+	})
 </script>
 
-<svelte:component this={MyComponent} foo="bar"/>
-```
-
-This method can also be used to import client-side libraries. For example, importing and using [Quill.js](https://github.com/quilljs/quill/)
-
-```html
-<script>
-	import { onMount } from 'svelte'
-
-	onMount(async () => {
-
-		const module = await import('quill');
-		let Quill = module.default;
-
-		let editor = new Quill('#editor', {
-			modules: { toolbar: '#toolbar' },
-			theme: 'snow'
-		});
-	});
-</script>
-
-<div id="toolbar">
-  <button class="ql-bold">Bold</button>
-  <button class="ql-italic">Italic</button>
-</div>
-
-<div id="editor">
-  <p>Hello World!</p>
-</div>
+<div bind:this={editor}></div>
 
 ```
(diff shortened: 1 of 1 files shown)


benmccann:
What makes something a client-side library or what were you trying to accomplish? I'm not familiar with Quill in particular. But speaking generally there may be some benefit to running an editor JS script on the server. If you run it on the server-side, it could server-side render (SSR) editor layout, controls, etc. Then when the page loads all it would need to do is attach event handlers for the controls. So I think you might only want to skip SSR when the script tries to access `window` as previously described

mateomorris:
By 'client-side library' I mean a library intended to run in a browser which depends on `window` or `document`.  The things you mentioned would be beneficial, but Quill and many other libraries can't be SSR'd at all because of those requirements. 

But either way, I think it would be helpful to offer a better example for importing a client-side library since the current example unnecessarily uses `<svelte:component>` and imports `my-ssr-component` where importing/mounting an actual library would be more descriptive to a real-world application.

benmccann:
> By 'client-side library' I mean a library intended to run in a browser which depends on window or document

We already have an example of that. I'm not sure sure that adding two examples of it is really useful. If the current example isn't a good one, I'd rather update it than add a second.

> the current example unnecessarily uses <svelte:component> and imports my-ssr-component where importing/mounting an actual library would be more descriptive to a real-world application.

I'm afraid I don't really see how using a real library is an improvement. That vast majority of users will not be using that particular library and so in either case it's a component name that doesn't match what they will be using.

Using Quill as an example adds stuff that's unnecessary such as the usage of `toolbar`. I think that having unneccesary code present makes it harder to see exactly what it is that is necessary

mateomorris:
I understand. Now that I'm looking at it again, the section seems to be talking about two different uses for the dynamic import: 
1. Importing your own Svelte component which depends on browser objects (this is what the example is showing)
2. Importing a third-party library which depends on browser objects ("Sapper works well with most third-party libraries you are likely to come across")

At the very least, I think the first example could be improved by referencing a Svelte component instead of `my-ssr-component`:
```
<script>
   import {onMount} from 'svelte'
   let MyComponent
	
   onMount(async() => {
	MyComponent = (await import('./MyComponent.svelte')).default
   })
	
</script>

<svelte:component this={MyComponent} />
```

And if it would be helpful (which I think it would be, especially for more junior devs), a minimal second example showing how to import/mount a library:
```
<script>
	import {onMount} from 'svelte'
	let Quill
	
	onMount(async() => {
		Quill = (await import('quill')).default
		new Quill('#editor')
	})
	
</script>

<div id="editor"></div>
```
https://svelte.dev/repl/5aaf090e2f83429a997eac91f6c39a3c?version=3.23.0

I hope I'm not belaboring the point, I'm just trying to make it a easier for newcomers to user Sapper because it's so amazing.

antony:
I think importing a svelte component with async imports is a bad example, because it encourages a pattern where people lazily write non-ssr compatible components, and then leave it to the consumer to async import them to work around the fact they're not ssr compatible.

The ideal solution in this scenario is to tell the author to make their component SSR compatible, or fork it and do it yourself.

I feel very uncomfortable about adding workarounds for low quality third-party components into the official documentation.

The quill example is fine - except the example isn't written ideally - don't reference element IDs, simplify the import syntax, and don't have the Quill variable declared outside of onMount (since it doesn't need to be):

```
<script>
	import {onMount} from 'svelte'
       let editor
	
	onMount(async() => {
		const { default: Quill } = await import('quill')
		new Quill(editor)
	})
</script>

<div bind:this={editor}></div>
```

mateomorris:
Thanks for the edit, that makes sense. Included your edits, cleaned up some verbiage, and removed the original example (since it's sounds like that's what you're advising).

site/content/docs/06-server-side-rendering.md
```diff
@@ -6,27 +6,29 @@ Sapper, by default, renders server-side first (SSR), and then re-mounts any dyna
...
+	
+	onMount(async() => {
+		const { default: Quill } = await import('quill')
+		new Quill(editor)
+	})
 </script>
 
-<svelte:component this={MyComponent} foo="bar"/>
+<div bind:this={editor}></div>
+
```
benmccann, line 33: no need for a blank line here

site/content/docs/06-server-side-rendering.md
```diff
@@ -6,27 +6,29 @@ Sapper, by default, renders server-side first (SSR), and then re-mounts any dyna
...
-	import { onMount } from 'svelte';
-
-	let MyComponent;
-
-	onMount(async () => {
-		const module = await import('my-non-ssr-component');
-		MyComponent = module.default;
-	});
+	import {onMount} from 'svelte'
+        let editor
```
benmccann, line 24: this should be tabs instead of spaces

mateomorris: good catch

site/content/docs/06-server-side-rendering.md
```diff
@@ -6,27 +6,28 @@ Sapper, by default, renders server-side first (SSR), and then re-mounts any dyna
...
-
-	let MyComponent;
+	import {onMount} from 'svelte'
+	let editor
 
-	onMount(async () => {
-		const module = await import('my-non-ssr-component');
-		MyComponent = module.default;
-	});
+	onMount(async() => {
```
benmccann, line 26: Sapper always puts a space between `async` and `()`. For example:

https://github.com/sveltejs/sapper/blob/674727587bc0b5f28352443e2d9f999d27da66d7/src/api/export.ts#L147

benmccann:
Sorry this had gone unreviewed for awhile. It looks like there's a merge conflict here and at this point I'm not really sure we'll be updating the Sapper docs much more with the focus being on SvelteKit, so I'm just going to go ahead and close this. Thanks for the suggestion on this

mateomorris:
Totally understandable! Appreciate the explanation & all your work on sveltekit

PR #958, opened 2019-10-23
[R+M] Adding clarity to the docs

forrestblade:
I had a hard time understanding the components folder, but I finally figured it out. One thing I noticed is the docs said nothing about it, so here is hopefully at least one small reference that my aid someone learning.


GGAlanSmithee:
this is a bit missleading since you can add that to any link to make it prefetch

Conduitry:
Indeed, it works with any link. And we don't want to be mentioning specific components that exist only in sapper-template and aren't part of an example anywhere else on the docs page.

PR #960, opened 2019-10-24
added errorRedirect to middleware opts

thebells1111:
errorRedirect is an object that can be passed to middleware that allows server side redirect on an error code.

errorRedirect: {errorStatus: 404, statusCode: 301, location: '/'} 

The above object can be passed to allow a server side redirect to the specified location based on the errorStatus.
errorStatus can be a number, string, or array of number/string, and will default to 404
statusCode can be a number/string for the reroute status code; i.e. 301, 302, and will default to 301
location will be a string with the path to redirect to, and will default to '/'

### Tests

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -11,7 +11,8 @@ import App from '@sapper/internal/App.svelte';
 
 export function get_page_handler(
 	manifest: Manifest,
-	session_getter: (req: Req, res: Res) => any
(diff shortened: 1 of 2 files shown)


benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit by adding a `svelte.config.cjs` file and adapters. I'm going to go ahead and close this PR since we won't be adding the feature to Sapper. You might like to check out SvelteKit instead

Issue #948, opened 2019-10-20
Make layout know if it's in error page or home page


hgl:
**Is your feature request related to a problem? Please describe.**
I asked this question on [SO](https://stackoverflow.com/questions/58463102/how-does-layout-svelte-know-if-its-in-the-error-page), but doesn't get much attentions.

I want to style the page differently in home and error page, but `segment` is `undefined` in both pagas.

**Describe the solution you'd like**
Either make segment return different values (e.g., `""` vs `undefined`), or another variable saying if the current is error or not.

I lean toward the former.

**How important is this feature to you?**
Giving different layouts to home and error page layout is pretty common I think, and given there doesn't seem to be a workaround, I feel it's pretty important.

intrikate:
On a side note, the `page` store from `@sapper/app` may offer a (slightly hacky) workaround. That being said, the problem surely seems common enough to warrant a better solution. I’d go as far as to say `_error.svelte` should not inherit the common layout by default (we can still import and use it explicitly if we want to).

silllli:
There’s a pull request https://github.com/sveltejs/sapper/pull/962

thebells1111:
PR #1250 may be the solution to this. It separates the `_layout` styling from the `_error` styling.

Conduitry:
`$page.error` is now available in 0.28.9.

Issue #824, opened 2019-07-26
RFC: Remove `segment` prop from layouts


Conduitry:
With the introduction of the `page` store in Svelte v3 Sapper, the `segment` prop passed to `_layout.svelte` components feels somewhat vestigial. I don't think there's anything that can be done with it that can't be done (in a more uniform, if slightly more verbose, way) with `$page.path`.

There seems to be a bit of confusion around `segment`, about how to gain access to other parts of the URL, and about how to inject other bits of data into the layout. These would hopefully all be avoided if we never actually passed any props into `_layout` and we instead had the docs tell people to just always use the stores.

Thoughts?

Conduitry:
We could perhaps ease people into this a bit, by issuing warnings rather than immediately breaking. It should be pretty straightforward to check at build time whether a given `_layout` component defines a `segment` prop, by inspecting the `vars` array returned by the compiler. We're already doing this elsewhere to check for the presence of `preload`.

pngwn:
As I mentioned, I have moved away from using `segment` at all. I find it easier to work with the `page` store which covers all `segment` usecases and many more besides. Layout's are pretty static so working out which part of the path they correspond to is pretty straightforward. 

Removing `segment` would make the API more consistent, would be easier to teach and, as you said, remove the question of 'can other props be passed to layouts'. While the change might be slightly controversial, I think in the long term it will make for a clearer more consistent API.

If we do choose to adopt this proposal, transitioning to it gently with a deprecation notice seems like a nice idea, especially since the `segment` prop is probably quite widely used.

Rich-Harris:
I do find it useful for e.g. controlling which nav button is selected (including nested, secondary navs). If we were to remove this — and I agree, it's weird and vestigial — what about adding a `$page.segments` array in its place? Or is it better for the user to derive that from `$page.path`?

pngwn:
I wondered about adding a segments array to the `page` store, it would obviously be useful but its also incredibly straight-forward to derive from the `path`. Maybe if you needed to do it extensively across an app it could get tedious but in my mind the segments are really just used for navigation or navigation related pieces of UI. I don't know if people use them much outside of the odd navigation component.

njbotkin:
That's funny, I just started using `segment` as a clean way to ask, "are there any sub routes mounted to this `_layout`?"  In `segment`'s stead, I might suggest `current`; the `_layout`'s path.  All child segments can be had with `$page.path.slice(current.length)`.

[...]

PR #962, opened 2019-10-27
changed handle_error and index segments

thebells1111:
Right now, the 'segment' prop that is passed to `_layout.svelte` will be undefined for both the index page, and for the error handling page. This will change the segments to `['index']` and `['_error']` respectively. Being able to differentiate between the two pages in the `segment` prop will allow the designer greater control over determining how the page is displayed, such as changing the Nav appearance based upon the segment sent.

### Tests

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -145,7 +145,7 @@ export function handle_error(url: URL) {
 			},
 			component: ErrorComponent
 		},
-		segments: preloaded
+		segments: ['_error']
 
 	}
 	const query = extract_query(search);
@@ -274,7 +274,8 @@ export async function hydrate_target(target: Target): Promise<{
 	branch?: Array<{ Component: ComponentConstructor, preload: (page) => Promise<any>, segment: string }>;
 }> {
 	const { route, page } = target;
-	const segments = page.path.split('/').filter(Boolean);
+	const segments = page.path === '/' ? ['index'] : page.path.split('/').filter(Boolean);
+	
 
 	let redirect: Redirect = null;
 


runtime/src/app/app.ts
```diff
@@ -274,7 +274,8 @@ export async function hydrate_target(target: Target): Promise<{
 	branch?: Array<{ Component: ComponentConstructor, preload: (page) => Promise<any>, segment: string }>;
 }> {
 	const { route, page } = target;
-	const segments = page.path.split('/').filter(Boolean);
+	const segments = page.path === '/' ? ['index'] : page.path.split('/').filter(Boolean);
```
krzysztof-grzybek, line 277: So, this change will not allow us to have a route like `index`? Why not just pass the empty string `''` instead? It's basically what we have in url.

Conduitry:
Given that we're probably going to be removing or at least deprecating `segment` (#824), I don't think this is the proper solution to #948, or at least isn't the entire solution. We probably want an additional field on the `$page` store that indicates whether this is an error page (and possibly also what the error is? I have a less strong opinion about that).

krzysztof-grzybek:
@Conduitry Sounds reasonable, thank You!

thebells1111:
It seems like it would be easy enough to put `error` and `status`  inside of the `$page` object. If this is done, the need for an `error` prop and `status` prop is probably eliminated, since they would be accessible in the `$page` object.

benmccann:
I'll go ahead and close this given the feedback above. We can consider https://github.com/sveltejs/sapper/pull/1498 as an alternative

PR #975, opened 2019-11-03
[Docs] add note about installing Cypress

rdela:
...to `content/docs/14-testing.md`

See
- <https://github.com/sveltejs/sapper-template/issues/22#issuecomment-500228964>
- <https://github.com/sveltejs/sapper-template/issues/177#issuecomment-548165305>

### DIFF
--- a/site/content/docs/14-testing.md
+++ b/site/content/docs/14-testing.md
@@ -8,8 +8,12 @@ You can use whatever testing frameworks and libraries you'd like. The default in
 ### Running the tests
 
 ```bash
-npm i -D cypress
+npm install --save-dev cypress
 npm test
 ```
 
-The first command installs Cypress as a dev dependency. `npm test` starts the server and opens Cypress. You can (and should!) add your own tests in `cypress/integration/spec.js` — consult the [docs](https://docs.cypress.io/guides/overview/why-cypress.html) for more information.
+`npm install --save-dev cypress` installs Cypress as a dev dependency.
+
+`npm test` starts the server and opens Cypress.
+
+You can (and should!) add your own tests in `cypress/integration/spec.js` — consult the [docs](https://docs.cypress.io/guides/overview/why-cypress.html) for more information.
--- a/site/content/docs/14-testing.md
+++ b/site/content/docs/14-testing.md
@@ -8,7 +8,8 @@ You can use whatever testing frameworks and libraries you'd like. The default in
 ### Running the tests
 
 ```bash
+npm i -D cypress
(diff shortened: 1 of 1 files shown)


rdela:
> I think this could be formatted better, a new paragraph perhaps, to explain the commands.

Not sure what you mean by this, could you try to be more specific?

rdela:
Clarified note and split into 3 paragraphs, and expanded the command (`npm i -D` => `npm install --save-dev`) per discussion in https://github.com/sveltejs/sapper-template/pull/194

@antony was this what you had in mind?

benmccann:
Related: https://github.com/sveltejs/sapper-template/pull/194

thgh:
I would not recommend to add Cypress by default, it slows down the default install significantly and is probably not used that often.

How about adding a cypress command that just tells (or lets) you to install the real cypress?

benmccann:
> I would not recommend to add Cypress by default, it slows down the default install significantly and is probably not used that often.

I think if it's not used that often we should just remove it. It seems too opinionated to be in the default template to me. It's not a good fit for too many projects: if you need to support IE 11, need to test iframes or popups, use Jest, etc.

rdela:
In the interest of incorporating as many points of view as possible, I agree with [Conduitry's comment](https://github.com/sveltejs/sapper-template/pull/227#issuecomment-632700810) on Ben's PR:

> Having the skeleton of a setup for Cypress without having Cypress as a devdep does seem bad, but adding Cypress as a devdep for everyone seems worse, because it is huge. I don't know what the right solution is here. We don't currently have a mechanism for generating different templates with different features toggled on and off, apart from the very simple Rollup vs webpack template thing.

I found the Cypress setup to be really informative and am using it on all my Sapper projects now. I agree with [Rich's 2017 assessment that it is the best way to test Sapper apps](https://github.com/sveltejs/sapper-template/issues/1#issue-282729139):

> Gleb Bahmutov was kind enough to write up his experience of using Cypress to test a Svelte app, and I think Cypress is probably the right go-to choice for testing Sapper apps.

It was crucial to getting GitLab CI to work when I was building my [Sapper GitLab Pages Markdown example](https://rdela.gitlab.io/sapper-v3/). The only stumbling block for me was having to install it and the lack of documentation around that. If we could get this PR and https://github.com/sveltejs/sapper-template/pull/194 merged then the lack of documentation problem would be solved at least.

benmccann:
Cypress was removed from the template in https://github.com/sveltejs/sapper-template/pull/263

PR #979, opened 2019-11-06
Add documentation for now v2 deployment

diverent2:
V1 is now offically deprecated and can no longer be used in new projects.
Add new documentation for V2

- [X] Related to https://github.com/sveltejs/sapper-template/issues/102
- [X] This message body should clearly illustrate what problems it solves. If there are related issues, remember to reference them.
- [X] Ideally, include a test that fails without this PR but passes with it. PRs will only be merged once they pass CI. (Remember to `npm run lint`!)
### Tests
-  [X] Run the tests tests with `npm test` or `yarn test`)

### DIFF
--- a/site/content/docs/11-deploying.md
+++ b/site/content/docs/11-deploying.md
@@ -7,9 +7,55 @@ Sapper apps run anywhere that supports Node 8 or higher.
 
 ### Deploying to Now
 
-> This section relates to Now 1, not Now 2
+We can very easily deploy our apps to [Now][https://zeit.co/dashboard]:
 
-We can very easily deploy our apps to [Now][]:
+#### Using Now v2 (recommended)
+
+We can use the [now-sapper loader](https://www.npmjs.com/package/now-sapper) to deploy our project and start the runtime.
+
+To do so, we will need set up our environment to work with `now`.  
+We do this by adding a `now.json` file to our project that specifies our loader and scripts to use. (`build` and `start` for runtime)
+
+```json
+{
+  "version": 2,
+    "builds": [
+    { "src": "package.json", "use": "now-sapper" }
+  ],
+}
+```
+
+We should also ignore some files using the `.nowignore`
+
+```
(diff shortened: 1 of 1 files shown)


thgh:
With the latest version of now-sapper, you don't need `.handler` => just `export default app` (same for express and polka)

site/content/docs/11-deploying.md
```diff
@@ -7,9 +7,57 @@ Sapper apps run anywhere that supports Node 8 or higher.
...
+```js
+polka() //same with Express
+  .use(
+    [...]
+  )
+```
+
+with
+```js
+const app = polka //same with Express
```
Vehmloewff: Should there be parentheses here?

```suggestion
const app = polka() //same with Express
```

diverent2:
@thgh I've updated to docs accordingly :)  
Thanks for your feedback!

diverent2:
*ping* :)

ryanditjia:
> With the latest version of now-sapper, you don't need `.handler` => just `export default app` (same for express and polka)

Perhaps the README in the repo could use an update? Polka still says `export default app.handler` 😄

brancusi:
Works great thank you

Conduitry:
Since this is such a desired feature, I think it's probably acceptable to have a link to direct people to using a third party solution (`now-sapper`) for this - but the official docs shouldn't take on the responsibility of explaining how to use it. We don't want people to come here if they're having trouble with it, and we don't want to have to worry about keeping the explanation in sync with now-sapper's real docs.

We should just have a link to `now-sapper` and its documentation, and mention that it is an unofficial project.

benmccann:
@diverent2 will you have a chance to consider @Conduitry's suggestion?

diverent2:
Hello there, since so much changed from the original PR I've opened a new one about the recommended changes here: 👉  https://github.com/sveltejs/sapper/pull/1199
See you there!

Issue #180, opened 2018-03-09
Support base paths


Rich-Harris:
Use case: you want to serve a Sapper app from somewhere other than the root of your website. On the server this is relatively straightforward:

```js
import polka from 'polka';
import sapper from 'sapper';
// ...
import { routes } from './manifest/server.js';

const app = polka();
const subapp = polka();

// ...

subapp.use(sapper({
  routes
}));

app.use('/subapp', subapp);

app.listen(4567);
```

This will (or should — I haven't *actually* tested it) mount e.g. `routes/foo.html` to `/subapp/foo`.

All that's missing, in theory, is for the client-side app to know about the base path. Proposed API:

```js
import { init } from 'sapper/runtime.js';
import { routes } from './manifest/client.js';

init(document.querySelector('#sapper'), routes, {
  base: 'subapp'
});
```

PR #984, opened 2019-11-09
Make <base> optional

### LINKED ISSUE #904: make sapper.base optional
sapper.base was introduced to solve https://github.com/sveltejs/sapper/issues/180 over a year ago. it is not optional for sapper to compile (i'm not sure why, but im sure there are good reasons). however, this causes screwy things to happen for relative/anchor links.

in discord, rich has voiced willingness:

>  unindented03/15/2019
> is there a way to avoid including %sapper.base% and have prefetching still work? that <base> tag is making me really unhappy
it means I can't have relative links anywhere
it also fucks up my anchor links
 
> rich_harris03/15/2019
> not in the current version. would love to get rid of it, it needs some care though

make `<base>` be optional, if not supplied by the user.

none

moderate. i'm autogenerating some links and having to account for base is making it pretty tricky.

thgh:
Adds an option `baseTag: false` to sapper.middleware to opt out of the `<base>` tag

Resolves #904 => issues with svg and css `url()`
Base tag was introduced in #180

Could perhaps be extended to support #866 (I have not checked effect on sapper export)
Could also be extended to support #600 (By adding a baseOrigin option)


benmccann:
@thgh I think this PR will end up being blocked until we figure out how to handle configuration in Sapper. [I raised an RFC](https://github.com/sveltejs/rfcs/pull/23) since I'm also blocked by that decision. Please take a look and feel free to leave any feedback there

benmccann:
It was reported in https://github.com/sveltejs/sapper/pull/1237 that export won't work if `base` is not present. I haven't tested myself, but assuming that's true than I imagine this will break export

thgh:
Looks like export only supports `base` indeed, that should be easily fixed by using [this `url`](https://github.com/akvadrako/sapper/blob/55ba30b6ceb6d8dec1e713046ff6ad48c94a2e25/src/api/export.ts#L154) as `base`.

benmccann:
https://github.com/sveltejs/sapper/pull/1208 is trying to solve the export problem, so I think if it gets merged first then it will be fine

benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit which has a `svelte.config.cjs` file. I'm going to go ahead and close this PR since we won't be adding the feature to Sapper, but if you'd like to check out SvelteKit we will be reviewing PRs for features there

Issue #730, opened 2019-06-09
"Server is not listening on port 3000" but application is running.


TheBeachMaster:
This  is **sort of** similar to [this issue](https://github.com/sveltejs/sapper/issues/431) and [this issue](https://github.com/sveltejs/sapper/issues/486) as well as this [issue](https://github.com/sveltejs/sapper/issues/278) . 

This is on a fresh setup as detailed [here](https://sapper.svelte.dev/docs#Getting_started). 

## Environment   

`Node Version: v8.16.0` , `OS: Arch Linux` , `npm version : 6.9.0` .

## Setting up (server.js) 

The default template comes with the following: 

```js 
import sirv from 'sirv';
import polka from 'polka';
import compression from 'compression';
import * as sapper from '@sapper/server';

const { PORT, NODE_ENV } = process.env;
const dev = NODE_ENV === 'development';

polka() // You can also use Express
	.use(
		compression({ threshold: 0 }),
		sirv('static', { dev }),
		sapper.middleware()
	)
	.listen(PORT, err => {
		if (err) console.log('error', err);
	}); 
```  

Then running `npm install` 

## Running 

`export HOST=127.0.0.1 PORT=3000`  

`npm run build` 

`npm run dev`  you get the following error 

```
 thebeachmaster@127  ~/test/svelte/hellosvelte  npm run dev                                                                                                                                  ✔  7070  10:50:04

> TODO@0.0.1 dev /home/thebeachmaster/test/svelte/hellosvelte
> sapper dev

✔ server (995ms)
✔ client (998ms)
✔ service worker (24ms)
> Server is not listening on port 3000
``` 
I went and checked if there was any process listening  on port 3000
`lsof -i :3000 | grep LISTEN` only to find that a `node` process was ... 

``` 
 thebeachmaster@thebeachmaster  ~/test/svelte/hellosvelte  lsof -i :3000 | grep LISTEN                                                                                                   1|1 ↵  7073  10:52:10
node    314 thebeachmaster   18u  IPv6 382263      0t0  TCP *:hbci (LISTEN)
 thebeachmaster@thebeachmaster  ~/test/svelte/hellosvelte   
```  

as seen here when I opened the browser...

![Sapper project template](https://user-images.githubusercontent.com/9129843/59156507-4b5af580-8aa5-11e9-8c1b-37719314c14e.png)

## The problem 
This results in hot-reloading failing.

[...]

PR #985, opened 2019-11-09
Wait upto 30 seconds for server ready

thgh:
Resolves sveltejs/sapper#730

Related: #1007


benmccann:
@thgh thanks for this. I think it might be the same as https://github.com/sveltejs/sapper/pull/1239 which was recently merged?

benmccann:
Could we do `ports.wait(this.port, { timeout: 30000 })` instead? It looks to me like you can from the [port-authority README](https://github.com/Rich-Harris/port-authority/) though I'm not 100% sure. If not, I think making that configurable in `port-authority` first would be cleaner than trying to work around it here

thgh:
I like seeing that some process is taking longer than expected, I think in this case it informs the user something's up with their underlying system. Too bad it's not very actionable, but I haven't seen an actual solution so not sure if there is a more helpful message.

benmccann:

thgh:
Sure!

benmccann:
Thanks!

Issue #993, opened 2019-11-18
sapper.middleware catches all server errors, would rather pass through


briancray:
**Is your feature request related to a problem? Please describe.**
I have a error handler in my express app that sends server errors to Sentry and does some custom error handling. Sapper's error handler catches all server errors and doesn't pass them through to my default.

**Describe the solution you'd like**
To pass through sapper server route errors to pass through to my error handling middleware

**Describe alternatives you've considered**
None

**How important is this feature to you?**
Error handling is muy importante

**Additional context**
No extra context

UnlimitedBytes:
Would be a nice feature!

koenkivits:
This feels like a must-have. Commenting on this issue to link #812, which is also about error handling and also contains a PR.

Either solution would work for me; I just want (Sentry) error logging to work.

soullivaneuh:
Is there any alternative to catch Sentry errors?

Trying to use the sentry error handler before the middleware:

```
import sirv from 'sirv';
import polka from 'polka';
import compression from 'compression';
// eslint-disable-next-line import/no-unresolved
import * as sapper from '@sapper/server';
import * as Sentry from "@sentry/node";

if (process.env.SENTRY_DSN) {
  console.log('sentry init', process.env.SENTRY_DSN);
  Sentry.init({
    dsn: process.env.SENTRY_DSN,
  });
}

const { PORT, NODE_ENV } = process.env;
const dev = NODE_ENV === 'development';

polka() // You can also use Express
  .use(
    Sentry.Handlers.requestHandler(),
    compression({ threshold: 0 }),
    sirv('static', { dev }),
    Sentry.Handlers.errorHandler(),
    sapper.middleware(),
  )
  .listen(PORT, (err) => {
    // eslint-disable-next-line no-console
    if (err) console.log('error', err);
  });
```

But this to anonymous weird errors.

If no workaround is available, then yes, the feature is quite a must have.

bamadesigner:
I came up with a simple fix for anyone who is interested/needs something to hold you over until SvelteKit.

I added the status code to the response sent during `handle_page()`, so when its passed back through middleware you can check response.statusCode inside the middleware function, e.g. `if (500 === response.statusCode)` equals true when a 500 error has been thrown.

I decided to fork sapper for myself since they might not do any more work here anyway. This is the only change I've made to my fork. Feel free to copy.

https://github.com/bamadesigner/sapper/pull/1/files

vish01:
[...]

PR #994, opened 2019-11-18
handleServerErrors parameter for sapper.middleware

briancray:
### Tests

Issue: https://github.com/sveltejs/sapper/issues/993

Sapper's server errors need to pass through to my server's error handler middleware instead of immediately responding with a 500.


dimfeld:
In order to maintain backwards compatibility, I think you want `handleServerErrors` have a default value of `true`.

benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit and we'd like to make sure this is possible there. I'm going to go ahead and close this PR since we won't be adding the feature to Sapper, but if you'd like to check out SvelteKit will be reviewing PRs for missing features there

Issue #547, opened 2019-01-16
Error when using route with multiple parameters in single segment


6eDesign:
This issue can be reproduced by creating a route such as `routes/blah.[slug].[id].html` which results in an error like this one: 
```
> Server crashed
webpack:///./__sapper__/server.js?:1
throw new Error("Module parse failed: Unexpected token (60:124)\nYou may need an appropriate loader to handle this file type.\n| \t\t\tpattern: /^\\/blah.([^\\/]+?)\\/?$/,\n| \t\t\tparts: [\n> \t\t\t\t{ name: \"blah_$slug$93_$91id\", file: \"blah.[slug].[id].html\", component: __blah_$slug$93_$91id, params: match => ({ slug].[id: d(match[1]) }) }\n| \t\t\t]\n| \t\t},");
^

Error: Module parse failed: Unexpected token (60:124)
You may need an appropriate loader to handle this file type.
|                       pattern: /^\/blah.([^\/]+?)\/?$/,
|                       parts: [
>                               { name: "blah_$slug$93_$91id", file: "blah.[slug].[id].html", component: __blah_$slug$93_$91id, params: match => ({ slug].[id: d(match[1]) }) }
|                       ]
|               },
    at eval (webpack:///./__sapper__/server.js?:1:7)
    at Object../__sapper__/server.js (C:\dev\homeadvisor\sapper-play\__sapper__\dev\server\server.js:97:1)
    at __webpack_require__ (C:\dev\homeadvisor\sapper-play\__sapper__\dev\server\server.js:21:30)
    at eval (webpack:///./src/server.js?:8:75)
    at Module../src/server.js (C:\dev\homeadvisor\sapper-play\__sapper__\dev\server\server.js:109:1)
    at __webpack_require__ (C:\dev\homeadvisor\sapper-play\__sapper__\dev\server\server.js:21:30)
    at C:\dev\homeadvisor\sapper-play\__sapper__\dev\server\server.js:85:18
    at Object.<anonymous> (C:\dev\homeadvisor\sapper-play\__sapper__\dev\server\server.js:88:10)
    at Module._compile (module.js:643:30)
    at Object.Module._extensions..js (module.js:654:10)
```

Conduitry:
I _think_ this might be fixed with something as simple as a non-greedy wildcard [here](https://github.com/sveltejs/sapper/blob/e2193a608076df5ac9186c66464816e737f6d5e5/src/core/create_manifest_data.ts#L257).

[...]

PR #1000, opened 2019-11-23
Allow multiple slugs in route segment

JohnPeel:
### Description
Currently you are not allowed to have multiple slugs in a single segment. (e.g. `src/routes/[file].[ext].js`)

This merge request relates to issue #547.

The regex used to find these parts will detect `file].[ext` as a slug because it is greedy.
I've made it non-greedy..

I also fixed an issue with non-dynamic parts of the segment (e.g. the `.` in the above filename) that are being treated as regex.
  * Example: http://localhost:3000/blog/what-is-sapperajson when using the example code given on new project. (Note the "a" instead of a ".", the "." in the pattern can be any character because it is treated as regex)

### Issues
* The change to sanitize non-dynamic parts of the segment may break projects that take advantage of the bug.

### Changes
* Regex for getting parts from a file segement was made non-greedy.
* Added regex sanitizing to non-dynamic parts of segment.
  * This broke a few tests (e.g. `.` => `\.`), these were fixed.
* Added tests for multiple slugs (with and without nested square brackets).

### Note
There is no lint script in your package.json.

---

### Tests

### DIFF
--- a/src/core/create_manifest_data.ts
+++ b/src/core/create_manifest_data.ts
@@ -298,7 +298,7 @@ function comparator(
 }
 
 function get_parts(part: string): Part[] {
-	return part.split(/\[(.+\(.+\)|.+?)\]/)
+	return part.split(/\[(.+?\(.+?\)|.+?)\]/)
 		.map((str, i) => {
 			if (!str) return null;
 			const dynamic = i % 2 === 1;
--- a/src/core/create_manifest_data.ts
+++ b/src/core/create_manifest_data.ts
@@ -84,16 +84,16 @@ export default function create_manifest_data(cwd: string, extensions: string = '
 					? basename
 					: basename.slice(0, -ext.length);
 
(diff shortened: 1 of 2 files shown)


JohnPeel:
The issue with nested square brackets may be mitigated by making the `.*` inside the parenthesis non-greedy.

src/core/create_manifest_data.ts
```diff
@@ -298,7 +298,7 @@ function comparator(
 }
 
 function get_parts(part: string): Part[] {
-	return part.split(/\[(.+)\]/)
+	return part.split(/\[(.+\(.+\)|.+?)\]/)
```
JohnPeel: Leaving myself a note to test `/\[(.+\(.+?\)|.+?)\]/`.

JohnPeel: It fixed issue 1 above. =)

PR #1001, opened 2019-11-24
Added options.preprocess and tests

DPr00f:
### Description

Currently you're not allowed to replace/add custom data to the generated html.

### Issues

Since a developer can't replace values in the `template.html` they'll end up with hacky solutions when wanting to replace values onto the generated html.

This is specially important for localisation and if the developer wants to replace `<html lang="%lang%">` in the template itself.

### Changes

- Added a custom option to the `sapper.middleware` called `preprocess`
- `preprocess` function gets called before the `status` and the response is ended
- Added tests called `preprocess` that illustrate a replacement

### Tests


DPr00f:
As a side note to this PR I understand many assumptions and a lot of liberty was taken.

I just want this discussion to not be ignored as it's important for i18n.

Possibly this could be done like the `nonces` option is done as well or with a middleware system for the `sapper.middleware`.

Just thought of preprocess for the name of the property as the whole project is surrounded by preprocessers.

Also, would like to point out that this can now be achieved with post-processing but it feels wrong IMHO.

## Other possible solutions and I would be happy to contribute to them:

### Add variables to the locals

Adding to `res.locals.sapper` object.

i.e.
```
// template.html
...
<html lang="%sapper.lang%">
...
```

```
// server.js

...
.use((req, res) => {
  res.locals.sapper = {
    lang: extractLangFromUrl(req.url)
  };
})
...
```

### Middlewares

```
// server.js
...
.use(
  sapper.middleware({
    middleware: [
      (body, req, res) => body.replace("%sapper.lang%", extractLangFromUrl('req.url')),
      otherMiddleware()
    ]
  })
)
```

benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit, so hopefully we can address this there. Someone's already sent a PR for it: https://github.com/sveltejs/kit/pull/641

PR #1021, opened 2019-12-07
sapper export --no-subfolders

bfanger:
**The problem**
I want `src/routes/foo.svelte` to match `/foo` in the url.

**Current situation**
`src/routes/foo.svelte` exports to `foo/index.html`
Certain static hosting providers (github-pages in my case) will see the generated foo folder and will redirect to `/foo/` (with a trailing slash) before displaying the contents of the index.html.

**This PR**
This PR add a `--no-subfolders` option to `sapper export`

The option wil generate `/export/foo.html` from `src/routes/foo.svelte`
Github pages and others will match the `/foo` url to the `foo.html` file without redirect.

The feature behaves similar to the [subFolders option in nuxt](https://nuxtjs.org/api/configuration-generate/#subfolders)

The patch itself is very small, couple of lines in cli.ts and export.ts
But i've also included tests (based on the regular export test) and added a line to the documentation.


src/api/export.ts
```diff
@@ -18,7 +18,8 @@ type Opts = {
 	cwd?: string,
 	static?: string,
 	basepath?: string,
-	host_header?: string,
+  host_header?: string,
```
bfanger: Looks like i'm using the wrong number of spaces or tabs, indentation looked fine in my editor.
I'll clean it up and `git push -f` the fixed commit.

bfanger:
I've rebased the upstream changes (fixed the merge conflict)

benmccann:
Thanks @bfanger. I think folks are a bit nervous about adding more configuration options until we figure out how exactly to handle configuration. I created an [RFC](https://github.com/sveltejs/rfcs/pulls) to try to move that conversation along, so please feel free to take a look in case you have any comments about it

bfanger:
After the sapper 0.28+ rebase, the CI failed on (14, macOS-latest) which is weird, because thats the platform i develop and test on, looking at the error i looks like the tests would pass if the CI server was little faster. 

And although the test itself is fast, the `before('build app')` and `before('export app')` steps are slow. The `tests/apps/export` takes about 3s to 7s on my beefy machine. That's too close the the 10s limit imposed by the `this.timeout(10000);`

benmccann:
The tests seem a bit flaky on Mac. I wouldn't worry about it failing on this PR. We're totally open to improvements to make it more robust though :slightly_smiling_face:

benmccann:

bfanger:
👍 i'll be switching to SvelteKit. Hope this feature wil be part of it.

Issue #1034, opened 2019-12-24
Utilise <base> tag in goto() in case of IE11


Case09:
When trying to use goto() in IE11, whole page will reload instead of navigation. As i understand this is because `document.baseURI` is undefined in IE11, so goto() function logic falls back to `location.href = href;`.

This can of course be fixed by setting `document.baseURI` in IE to value of `<base>` tag that sapper adds, but maybe this should be added in sapper code base.

In goto() definition new URL is made with
`const target = select_target(new URL(href, document.baseURI));` This can maybe be changed to call function like this instead of `document.baseURI` directly:
```
function getBaseURI() {
  if (document.baseURI) return document.baseURI;
  const base = document.getElementsByTagName('base');
  if (base.length > 0) return base[0].href;
  return document.URL;
}
```

Conduitry:
This should be fixed in 0.28.9.

Issue #657, opened 2019-05-04
Usage of shimport makes legacy support prohibitively challenging


jacobmischka:
Shimport is a super neat project, but its critical integration in recent Sapper apps makes supporting IE11 very challenging. While shimport may work in legacy browsers when the APIs it uses are properly polyfilled, there's no convenient way currently to include such polyfills in Sapper apps, as Sapper uses shimport to load all of the user code.

The best method of working around this issue I've found is to include a polyfill.io script tag before `%sapper.scripts%`, but even so I've failed in coming up with the correct set of polyfills to include to get Shimport to work correctly.

Currently, I'm using the following features:

- Blob
- fetch
- Promise
- URL
- Map
- String.prototype.startsWith
- Array.from
- Array.prototype.keys (may be unnecessary)
- classList (pretty sure unnecessary)

Resulting in the following script tag:

```html
	<script crossorigin="anonymous" src="https://polyfill.io/v3/polyfill.min.js?features=Blob%2Cfetch%2CPromise%2CURL%2CElement.prototype.classList%2CArray.prototype.keys%2CMap%2CArray.from%2CString.prototype.startsWith"></script>
```

Using these polyfills, I get an `Unsupported BodyInit type` error in IE11 when Shimport tries to load my client code, which seems to be caused when the fetch polyfill tries to use a body of type Blob. Am I missing any polyfills that would allow Shimport to work? Are there any other escape hatches I may be missing?

Ideally I'd like to remove the reliance on polyfill.io, as I fear it may not be polyfilling correctly which is what's causing the error, but I can't think of a better way to integrate polyfills _before_ shimport is loaded using the current configuration options supported by Sapper. I'm currently using the rollup template, though I'd be happy to switch to webpack instead if that makes a workaround any easier.

An automatic workaround when compiling with `--legacy` would be amazing, but if we can come up with a manual one in the meantime that would be really great.

Thanks, as always!

gnuletik:
Hey @jacobmischka, I'm also trying to make sapper works on IE11.
Your issue helped me understanding shimport, thanks !

I was able to solves the `Unsupported BodyInit type` error which was caused by polyfill.io 's implementation of fetch.
The workaround was to use [github/fetch](https://github.com/github/fetch)'s implementation and keep polyfill.io for the dependencies of fetch.

I inserted this before `%sapper.scripts%`
```html
[...]

PR #1037, opened 2019-12-27
feat(template): support for arbitrary replacements

### LINKED ISSUE #1036: ability to replace text on template through middleware
I want to set up (through middleware) the `lang` and `dir` attributes for tags in `template.html`, and Sapper doesn't give me a chance to do so.

I already implemented a solution on a PR I'm about to submit.
Basically, what it does is, look for the property `replacers` in the response object (affectionately known as `res`) and replace text on the template for every key on that property.
So, if any middleware has put anything on that property, you're good to go.

I need this to implement better internationalization. And internationalization is VERY important to me.

vilarfg:
Fixes #1036


DominikGuzei:
Thanks @vilarfg, works great for SSR rendering of JSS styles into the template 👍

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -320,12 +320,14 @@ export function get_page_handler(
...
 			const nonce_attr = (res.locals && res.locals.nonce) ? ` nonce="${res.locals.nonce}"` : '';
 
-			const body = template()
-				.replace('%sapper.base%', () => `<base href="${req.baseUrl}/">`)
-				.replace('%sapper.scripts%', () => `<script${nonce_attr}>${script}</script>`)
-				.replace('%sapper.html%', () => html)
-				.replace('%sapper.head%', () => `<noscript id='sapper-head-start'></noscript>${head}<noscript id='sapper-head-end'></noscript>`)
-				.replace('%sapper.styles%', () => styles);
+			const body = replace(template(), {
+				...res.replacers,
```
arve0, line 324: I suggest moving this below defaults, such that you can override them. Overriding defaults could be a flexible way to solve #1034, #904, #657 and similar issues. For example: including polyfills for certain user-agents, excluding scripts on others, using absolute urls in `<base>`, etc.

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -356,6 +358,18 @@ export function get_page_handler(
 	};
 }
 
+function replace(
+    template: string,
+    replacers: Record<string, string | CallableFunction>
+) {
+    for (const key in replacers) {
+        if (replacers.hasOwnProperty(key)) {
+            template = template.replace(new RegExp(`%sapper.${key}%`, 'g'), replacers[key] as any);
```
arve0, line 367: Have replacers less coupled to the `%sapper.key%` pattern? It would be flexible to let replacers define their own patterns to replace.

happycollision: I didn't realize that this PR was around when I started my work, but I've done something similar over at #1642, which happens to address your concern here, @arve0

runtime/src/server/middleware/get_page_handler.ts
happycollision: Same here as my other comment, actually.

benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit by adding a `svelte.config.cjs` file and adapters. I'm going to go ahead and close this PR since we won't be adding the feature to Sapper. You might like to check out SvelteKit instead. Someone's sent a PR for this feature already actually https://github.com/sveltejs/kit/pull/641

PR #1052, opened 2020-01-12
Wait for window load event before registering the service worker

### LINKED ISSUE #970: wait for window load event before registering the service worker?
Hi guys,

I was wondering, should we wait for the window load event before registering the service worker as it seems to be common practice to do so?

```js
if ('serviceWorker' in navigator) {
  window.addEventListener('load', function() {           //  <-- wait for window to load before registering the sw
    navigator.serviceWorker.register('/service-worker.js');
  });
}
```

Most service worker implementations I've seen do, sw libs like workbox do it as well so I was wondering if that small change is worth adding?
https://developers.google.com/web/tools/workbox/guides/service-worker-checklist

https://github.com/sveltejs/sapper/blob/a52bdb2f4e1a722f06134b4065da2a32969e12e2/runtime/src/server/middleware/get_page_handler.ts#L278

NOtherDev:
Fixes #970 and related.

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -275,7 +275,7 @@ export function get_page_handler(
 			].filter(Boolean).join(',')}};`;
 
 			if (has_service_worker) {
-				script += `if('serviceWorker' in navigator)navigator.serviceWorker.register('${req.baseUrl}/service-worker.js');`;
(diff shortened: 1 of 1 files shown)


benmccann:
Sorry this had gone unreviewed for so long. I'm hesitant to touch Sapper more than necessary at this point. Maybe it's better to try to make sure SvelteKit works as desired instead at this point?

benmccann:

PR #1055, opened 2020-01-15
Adds example to goto documentation

kevmodrome:
I was a bit confused as to where I would import goto from so I added a small usage example.

### DIFF
--- a/site/content/docs/03-client-api.md
+++ b/site/content/docs/03-client-api.md
@@ -33,6 +33,20 @@ Programmatically navigates to the given `href`. If the destination is a Sapper r
 
 Returns a `Promise` that resolves when the navigation is complete.
 
+```html
+<script>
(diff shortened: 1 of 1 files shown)


Conduitry:
Maybe an example with `await` to indicate that this navigation is not synchronous, and that you can use the promise to deal with that,

kevmodrome:
I changed the example a bit and added a sentence to the promise documentation. I'm not sure how elaborate of an example is warranted. Should I do one that does something both on resolve as well as reject or is this fine?

Conduitry:
I went to push a couple of tweaks before squashing merging this, but apparently you didn't enable allowing maintainers to push to the branch this PR is from.

kevmodrome:
My bad @Conduitry, enabled it.

PR #1067, opened 2020-01-23
Remove head hydration workaround

### LINKED ISSUE #1066: Remove head hydration workaround
Since Svelte 3.17.3 hydration now works with `head` elements. This removes the need for the `noscript` workaround.

I will open PRs for `sapper` (sveltejs/sapper#1067) and `sapper-template` (sveltejs/sapper-template#195)

Use the workaround in perpetuity since it works now and probably forever.

Not very; it is just a minor cleanup for the next time breaking changes to Sapper are up for consideration.

**Additional context**
- sveltejs/svelte#1607
- sveltejs/svelte#4082
- sveltejs/svelte#4250
- sveltejs/svelte#4311

johnmuhl:
Closes #1066

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "yootils": "0.0.16"
   },
   "peerDependencies": {
-    "svelte": "^3.5.0"
+    "svelte": "^3.17.3"
   },
   "scripts": {
     "test": "mocha --opts mocha.opts",
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -228,16 +228,6 @@ async function render(redirect: Redirect, branch: any[], props: any, page: Page)
 			props: await root_preloaded
 		};
 
-		// first load — remove SSR'd <head> contents
-		const start = document.querySelector('#sapper-head-start');
-		const end = document.querySelector('#sapper-head-end');
-
-		if (start && end) {
-			while (start.nextSibling !== end) detach(start.nextSibling);
-			detach(start);
-			detach(end);
-		}
-
 		root_component = new App({
 			target,
 			props,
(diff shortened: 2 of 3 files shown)


benmccann:
@johnmuhl should this also update the `package.json` to set svelte 3.17.3 as the minimum version required?

Conduitry:
I'm taking a look at this again now that we're committed to the next release containing breaking changes - and I'm bothered that removing this isn't breaking any tests, even though we aren't updating any of the tests to include `hydratable: true` in the bundler's compiler config.

benmccann:
Every test already has `hydratable: true` in its `rollup.config.js` file, so I think that's to be expected

Conduitry:
The version of Svelte used in the tests in this branch is ancient though - https://github.com/sveltejs/sapper/blob/fcdd67be73b5ff9eb67ad57251952af8db1e8024/package-lock.json#L5312 - much older than the feature that this change relies on, which leads me to think this isn't being tested at all.

benmccann:
Yeah, there's just one test for hydration and it does't test the `head` element: https://github.com/sveltejs/sapper/blob/a5a421f4b689eb6030db592f6dc85178d5c05d68/test/apps/basics/test.ts#L170

Do we need to add a new test for it though? Svelte should have tests for `head` hydration and I'm not sure if we're doing anything special once the workaround is removed

Conduitry:
The `pkg.peerDependencies` should at least be updated to reflect the new requirement. This probably doesn't need to be called out in the migration section in the docs. I guess I'm okay without a test for this, since Svelte should be already testing it.

benmccann:
Ok, I just bumped the `peerDependencies`, so I'll go ahead and merge this one

PR #1071, opened 2020-01-24
Doc fix: how to use regex is not clear

KristerV:
I spent hours trying to get the `/routes/[regex]/[another].svelte` to work. Hoping this will help someone.

### DIFF
--- a/site/content/docs/02-routing.md
+++ b/site/content/docs/02-routing.md
@@ -122,6 +122,16 @@ The `error` object is made available to the template along with the HTTP `status
 
 You can use a subset of regular expressions to qualify route parameters, by placing them in parentheses after the parameter name.
 
-For example, `src/routes/items/[id([0-9]+)].svelte` would only match numeric IDs — `/items/123` would match, but `/items/xyz` would not.
+For example, `src/routes/items/[id([0-9]+)].svelte` would only match numeric IDs — `/items/123` would match, but `/items/xyz` would not. These dynamic parameters will be available in `page.params`. See the [Pages section](docs#Pages) for more details.
+
+Examples:
+
+```
+index.html                         # normal
+[page].html                        # match anything. page.params.page holds the value
+[lang([a-z]{2})]/index.html        # match two letters in directory. page.params.lang holds the value
+[lang([a-z]{2})]/[page].html       # match two letters. page.params is {lang: String, page: String}
+[category([a-z-]{3,})]/[page].html
+```
 
 Because of technical limitations, the following characters cannot be used: `/`, `\`, `?`, `:`, `(` and `)`.


site/content/docs/02-routing.md
```diff
@@ -122,6 +122,16 @@ The `error` object is made available to the template along with the HTTP `status
...
-For example, `src/routes/items/[id([0-9]+)].svelte` would only match numeric IDs — `/items/123` would match, but `/items/xyz` would not.
+For example, `src/routes/items/[id([0-9]+)].svelte` would only match numeric IDs — `/items/123` would match, but `/items/xyz` would not. These dynamic parameters will be available in `page.params`. See the [Pages section](docs#Pages) for more details.
+
+Examples:
+
+```
+index.html                         # normal
+[page].html                        # match anything. page.params.page holds the value
+[lang([a-z]{2})]/index.html        # match two letters in directory. page.params.lang holds the value
+[lang([a-z]{2})]/[page].html       # match two letters, page.params is {lang: String, page: String}
```
benmccann, line 133: nit: I might use a `.` instead of `,` here

KristerV: No problem, done.

Conduitry:
I'm not really convinced of the need to remind readers about page params a few paragraphs after just talking about them, nor of the need to several examples of what regular expressions look like.

benmccann:
Yeah, I was debating that as well. There's not a lot of value add in having more examples in my mind. If we wanted to be even more explicit we could update the existing sentence to something like:

>For example, `src/routes/items/[id([0-9]+)].svelte` would only match numeric IDs — `/items/123` would match and make the value `123` available in `page.params.id`, but `/items/xyz` would not match.

That'd be a lot more concise than adding additional sentences to describe features that were already explained, but still be more explicit so that there's less for users to piece together.

KristerV:
mm, as the new user who did in fact get confused and spent hours looking for answers (finding it in the original issue where feature was proposed) am saying the readme is not clear enough and needs _explicit_ examples. Unless the readme is not meant for new users coming to use this library for the first time I'd highly suggest "reminding" some fundamental principles.

antony:
@KristerV I think the reasoning here is that if we were to document how to use regex, then why not include how to use HTTP, NPM, maybe a quick intro to javascript, etc.

It's out-of-scope, and far better explained and demonstrated on https://regex101.com which could maybe be linked from the words "regular expressions" as an absolute maximum.

If we make the docs too big by going into explicit detail about everything, it actually has a negative impact on new users, as people don't enjoy reading reams of documentation.

kevmodrome:
Maybe a good middle ground is to link to something like: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions ? This could be done for most of the out-of-scope things in order to help the user understand what they're reading.

KristerV:
yeah I see your point i guess. I mean my confusion was never about regex, but more about the `name(regex)` syntax. SO maybe I did put too many examples, just wanted to be extra clear.

benmccann:
@KristerV would my suggested text in https://github.com/sveltejs/sapper/pull/1071#issuecomment-629802267 have cleared things up for you?

KristerV:
oh sorry, forgot to mention - yes. it does since it mentions params.id.

Issue #1074, opened 2020-01-27
`sapper build` breaks when passing --cwd and [dest]


mikeparisstuff:
**Describe the bug**
I believe I have found a small bug in how paths are resolved when using both `--cwd` and a `[dest]` in `sapper build`. I have been trying to move a sapper project to build within bazel which led me to discover this.

I am trying to run this sapper command:

```
sapper build relative/path/to/output/dir --legacy --cwd ./relative/path/to/sapper-app --bundler rollup
```

My `rollup.config.js` is at path `./relative/path/to/sapper-app/rollup.config.js` and if I don't provide a `cwd` then sapper is unable to find it. 

The problem is that if you run this, then you will see an `index.js` at `relative/path/to/output/dir/index.js` but the rest of the sapper project build is saved to `./relative/path/to/sapper-app/relative/path/to/output/dir`. It looks like the [path resolved here](https://github.com/sveltejs/sapper/blob/d99d67bbb1ca0bc1cbb923392c6f5530518763ff/src/cli.ts#L174) is going to end up being different than the [path resolved here](https://github.com/sveltejs/sapper/blob/d99d67bbb1ca0bc1cbb923392c6f5530518763ff/src/api/build.ts#L43) which results in the problem that I am seeing.

Without thinking too hard, I think there may be a few fixes.

1. Update the [the first path](https://github.com/sveltejs/sapper/blob/d99d67bbb1ca0bc1cbb923392c6f5530518763ff/src/cli.ts#L174) to take into account the `--cwd`.
2. Add more options to `build`. E.G. if I could set a `--config` path and a `--src` path then I wouldn't need to change the `--cwd` and all relative paths would work again.
3. Force `[dest]` into an absolute path in the `cli` [perhaps here?](https://github.com/sveltejs/sapper/blob/d99d67bbb1ca0bc1cbb923392c6f5530518763ff/src/cli.ts#L172) using `path.resolve`. I think this would require changing how `--cwd` works tho so option (1) is probably better.

I have looked into alternatives to try to force absolute paths in bazel but this is not only much harder as you need to slightly tweak a lot of default build rules and seems to be a bad practice.

**To Reproduce**
To reproduce, run sapper build with both a `[dest]` and a `--cwd` from a parent directory. E.G. if you have a project at `/Users/me/projects/sapper-app` run the following from the path `/Users/me/projects`.

```
sapper build ./sapper_out --legacy --cwd "./sapper-app" --bundler rollup
```

You will see the following:

[...]

PR #1075, opened 2020-01-28
Add --bundlerConfig to cli commands so you can build a nested dir w/o changing --cwd.

mikeparisstuff:
# The problem

**Issue: https://github.com/sveltejs/sapper/issues/1074**

The launcher script created in `cli.ts` is placed in a different `[dest]` path than the rest of the build when you provide both a `--cwd` and `[dest]` in the command. This change causes the launcher file to also take the `--cwd` into account and places the launcher file in the correct location.

## Example

If you have a sapper project at `/Users/me/projects/sapper_app` and you run this command: 

```
sapper build ./sapper_out --cwd ./sapper_app
```

from `/Users/me/projects` then you get the following build output

```
/Users/me/projects/sapper_out
   /index.js
/Users/me/projects/sapper_app/sapper_out
   /build.json
   /template.html
   /service-worker.js
   /client/
   /server/
```

 This PR fixes a few issues with the `--cwd` but does not fix all of them. I switched gears midway through to solve the problem with a new argument `--bundlerConfig`, but I'm pretty sure the fixes to `--cwd` are still in the right direction so I left them in place for now.

This PR makes three changes:

1. The fix is a simple change to include the `--cwd` in the launcher path, if one exists, and places `index.js` in the same directory as the rest of the build. This was not enough to fix the problem but I believe is still a bug fix.
2. Adds a new `--bundlerConfig` arg allows you to point to the `rollup.config.js` or `webpack.config.js` which is still expected to be in the root relative to the `--cwd` and is the last missing arg outside of `--src`, `--output`, `--static`, and `--routes`. 
3. Changes the logic in the generated `server.js` to find `build.json` based on a path relative to `server.js` instead of relative to the `--cwd` at build time. This allows you to build a sapper app in one place and then run it in another.

Changes 2 & 3 allow you to run:

```
sapper build __sapper__ --src client/sapper/src --routes client/sapper/src/routes --static client/sapper/static --output client/sapper/node_modules/@sapper --bundlerConfig client/sapper/rollup.config.js --bundler rollup
```

which results in a build that can be packaged, moved, and run in the same way. This means you can run the app using `node __sapper__` or, if the directory is somewhere else, `node some/nested/path/__sapper__`.

### Tests

The tests pass and I have a successful build & run working now 👍


mikeparisstuff:
After fixing `build --cwd [dest]` in the build, I was trying `dev --cwd sapper_app --build-dir ../sapper_dev` and it failed with error:

```
> ENOENT: no such file or directory, open 'src/template.html'
```

The issue was very similar and was that the `src` variable in the constructor in `dev.ts` was being used in one location without taking `--cwd` into account and another where it was.

I went ahead and added the fix to this PR and have confirmed that `dev --cwd` now works.

mikeparisstuff:
The original fixes in this CR that tried to fully support `--cwd` were not enough to solve the problem. I shifted gears and instead decided to try to use `--src`, `--routes`, `--static`, and `--output` to change the source locations without changing the `--cwd`. To make this work, I had to change how the `rollup.config.js` and `webpack.config.js` as those were expected to be in the root of the `--cwd`. I added a `--bundlerConfig` option to build, dev, and export so that you can specify where the bundler config file should be found. This change allowed me to get a working build going.

After the --bundlerConfig was added, I went and changed how the `build.json` file was found in the generated `server.js`. This was trying to use a hardcoded path relative to the `--cwd` at build time which introduced some issues in sandboxed build environments like bazel where paths are not easily assumed. The `build.json` is always in the same spot relative to the `server.js` file so I changed it to look relative to that file. After these two changes, I am able to build the project and run it successfully. Success :)

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "sapper",
+  "name": "motley-sapper",
```
benmccann, line 2: this shouldn't change

pngwn:
API additions have to meet a pretty high bar in general but we won't be looking at this until some higher-level questions about how we handle configuration have been answered.

ajbouh:
I've applied this PR to my own build and discovered that switching to `__dirname` in the generated code requires an additional bit of config if using the webpack bundler.

To prevent webpack from hardcoding `__dirname` to `"/"`, you need to modify your webpack.config.js to look something like:

```
module.exports = {
  // ...
  server: {
    // ...
    node: {
      __dirname: false
    }
  }
}
```

benmccann:

PR #1085, opened 2020-02-08
Document the actual use of `export --entry`

happycollision:
The option for `--entry` is mentioned in the docs, but the actual api for it is not mentioned. Now it is.

### DIFF
--- a/site/content/docs/10-exporting.md
+++ b/site/content/docs/10-exporting.md
@@ -50,8 +50,6 @@ The `--entry` option expects a string of space-separated values. Examples:
 
 ```bash
 sapper export --entry "some-page some-other-page"
-
-sapper export --entry single-entry
 ```
 
 Additionally, any non-page routes should be requested in `preload`, *not* in `onMount` or elsewhere.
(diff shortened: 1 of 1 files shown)


site/content/docs/10-exporting.md
```diff
@@ -44,7 +44,17 @@ You can also add a script to your package.json...
...
 
-Because of this, any pages you want to be included in the exported site must either be reachable by `<a>` elements or added to the `--entry` option of the `sapper export` command. Additionally, any non-page routes should be requested in `preload`, *not* in `onMount` or elsewhere.
+Because of this, any pages you want to be included in the exported site must either be reachable by `<a>` elements or added to the `--entry` option of the `sapper export` command.
+
+The `--entry` option expects a string of space-separated values. Examples:
+
+```bash
+sapper export --entry "some-page some-other-page"
+
+sapper export --entry single-entry
```
benmccann, line 54: nit: I think the first example is sufficient and we probably don't need both

site/content/docs/10-exporting.md
```diff
@@ -44,7 +44,17 @@ You can also add a script to your package.json...
...
+```
+
+Additionally, any non-page routes should be requested in `preload`, *not* in `onMount` or elsewhere.
```
benmccann, line 57: I think it would be useful to explain why

happycollision: I actually no longer remember why myself, and don’t really have time to dig into it again. But I agree it would be useful.

logannc: I think onMount is run only in the browser so it would not be picked up if you wanted onMount resources exported.

logannc: Could you also clarify the example with exposition for what files are there? That is, what does `some-page` map to? `src/routes/some-page.svelte` or `src/routes/some-page/index.svelte` or potentially both?

PR #1094, opened 2020-02-21
Added "any" method and fixed `res.write()`

Saiv46:
1. Added "any" field
Many web frameworks allowing you to match all HTTP methods.
https://expressjs.com/en/4x/api.html#app.all for example.

2. See https://nodejs.org/api/http.html#http_response_write_chunk_encoding_callback
`res.write()` must return boolean, but it doesn't because data got intercepted.

### Tests


benmccann:
Can you update the documentation in https://github.com/sveltejs/sapper/blob/master/site/content/docs/02-routing.md#server-routes ?

thgh:
Could we change `any` to `default` so it's aligned with Next.js? (Next.js only has export default which acts as any)

benmccann:

benmccann:

Issue #947, opened 2019-10-19
service worker: Does not register a service worker that controls page and start_url


evdama:
Hi folks,

after about ten days of asking Google, Stackoverflow and on discord this issue is my last resort of hope for any pointers for an issue that I'm seeing with my sapper application.

I've not changed or original `service-worker.js` file from the sapper template:
```js
import { timestamp, files, shell, routes } from '@sapper/service-worker';

const ASSETS = `cache${timestamp}`;

// `shell` is an array of all the files generated by the bundler,
// `files` is an array of everything in the `static` directory
const to_cache = shell.concat(files);
const cached = new Set(to_cache);

self.addEventListener('install', event => {
	event.waitUntil(
		caches
			.open(ASSETS)
			.then(cache => cache.addAll(to_cache))
			.then(() => {
				self.skipWaiting();
			})
	);
});

self.addEventListener('activate', event => {
	event.waitUntil(
		caches.keys().then(async keys => {
			// delete old caches
			for (const key of keys) {
				if (key !== ASSETS) await caches.delete(key);
			}

			self.clients.claim();
		})
	);
});

self.addEventListener('fetch', event => {
	if (event.request.method !== 'GET' || event.request.headers.has('range')) return;

	const url = new URL(event.request.url);

	// don't try to handle e.g. data: URIs
	if (!url.protocol.startsWith('http')) return;

	// ignore dev server requests
	if (url.hostname === self.location.hostname && url.port !== self.location.port) return;

	// always serve static files and bundler-generated assets from cache
	if (url.host === self.location.host && cached.has(url.pathname)) {
		event.respondWith(caches.match(event.request));
		return;
	}

	// for pages, you might want to serve a shell `service-worker-index.html` file,
	// which Sapper has generated for you. It's not right for every
	// app, but if it's right for yours then uncomment this section
	/*
	if (url.origin === self.origin && routes.find(route => route.pattern.test(url.pathname))) {
		event.respondWith(caches.match('/service-worker-index.html'));
		return;
	}
	*/

	if (event.request.cache === 'only-if-cached') return;

	// for everything else, try the network first, falling back to
	// cache if the user is offline. (If the pages never change, you
	// might prefer a cache-first approach to a network-first one.)
	event.respondWith(
		caches
			.open(`offline${timestamp}`)
			.then(async cache => {
				try {
					const response = await fetch(event.request);
					cache.put(event.request, response.clone());
					return response;
				} catch(err) {
[...]

PR #1095, opened 2020-02-21
exclude .DS_Store from Service Worker

### LINKED ISSUE #667: Remove .DS_Store (etc?) from service worker manifest
If the service worker tries to cache `.DS_Store` on installation, installation will fail

russelgal:
Fixes #667

exclude .DS_Store

### Tests


src/utils.ts
```diff
@@ -35,7 +35,7 @@ export function walk(cwd: string, dir = cwd, files: string[] = []) {
 		}
 	});
 
-	return files;
+	return files.filter(v=>!v.includes('.DS_Store'));
```
benmccann, line 38: will this handle `.DS_Store` in nested directories?

it might be slightly more efficient to change the `else` in this method to `else if` and do the check there so that you don't have to do an extra iteration over the list of files

bc-jasond:
Can this go out?  I tested it locally by just pasting the `filter()` line into the `walk()` function.  Looks like it will fix this issue too https://github.com/sveltejs/sapper/issues/947

src/utils.ts
benmccann: yeah, I guess this happens during build time, so no need to prematurely optimize performance here. lgtm

src/utils.ts
```diff
@@ -35,7 +35,7 @@ export function walk(cwd: string, dir = cwd, files: string[] = []) {
 		}
 	});
 
-	return files;
+	return files.filter(v => !v.includes('.DS_Store'));
```
pngwn, line 38: Instead of filtering the array after we have just created it, could we just prevent the addition of `.DS_Store` files to the array in the first place? Additionally, comparing only the `file` value will gives us a bit more certainty that we are only omitting `.DS_Store` files and nothing else. 

False positives seem unlikely, but I've stopped trying to second-guess all of the weird things people do with Sapper.

Conduitry:
Are there other files that should be excluded here as well? The original issue has "(etc?)" in the name, which indicates that there's more that should be looked into here. What exactly is it about this file that makes installation of the service worker fail?

pngwn:
`Thumbs.db` could be another. But they haven’t been stored in the folder itself since Vista. I’m not sure if any other hidden files cause issues.

Conduitry:
On Windows 10, I've actually had a couple folders get Thumbs.db files added to them in recent months, to my surprise. For that, though, I'd have to say that I'd expect it to be my responsibility to make sure I didn't have any such files in an app I was trying to build with Sapper.

I still don't understand what it is about `.DS_Store` that causes the service worker installation to outright _fail_, and not just that causes it to cache extra files we don't need. If we can distill that, then IMO we should attempt to exclude files that make it fail, but it would be the user's responsibility to make sure other files they don't want aren't there.

pngwn:
Do servers generally ignore hidden dot files by default? I guess this is the issue, when trying to cache the `.DS_Store` there is a 404 that the file cannot be found. So either the files are not uploaded (and so cannot be found when trying to precache) or they are hidden by the server (with the same result).

The big difference between `.DS_Store` and `Thumbs.db` is that one is a hidden file, which might explain the issues.

src/utils.ts
benmccann: yeah, I agree that seems a bit better. I've tweaked the PR to do that

benmccann:
I don't have a Mac, so can't easily test why it would fail. @Rich-Harris filed the bug originally though, so maybe he has more details

benmccann:
If we say it's the user's responsibility, how would we expect them to handle that? Modify their `package.json` to do something like the following? Or tell them never to open a directory with an image in it :smile: 

```
"build": "find . -name .DS_Store | xargs rm && sapper build",
```

russelgal:
> If we say it's the user's responsibility, how would we expect them to handle that? Modify their `package.json` to do something like the following? Or tell them never to open a directory with an image in it 😄
> 
> ```
> "build": "find . -name .DS_Store | xargs rm && sapper build",
> ```

very user friendly )

antony:
As an aside, these files should be ignored by gitignore, so for people building and deploying from CI it shouldn't be an issue.

However. I'm not sure I agree with the basis of this PR. Maintaining yet another filter of files dumped all over the filesystem by some specific OS is just painful.

It might be better to understand the root cause of the DS_Store issue.

benmccann:
@russelgal can you clarify what error you get if a `.DS_Store` file is encountered?

benmccann:
@russelgal bc-jasond can you clarify what error you get if a `.DS_Store` file is encountered?

benmccann:
@russelgal @bc-jasond can you clarify what error you get if a `.DS_Store` file is encountered?

russelgal:
> @russelgal @bc-jasond can you clarify what error you get if a `.DS_Store` file is encountered?

404 error in service worker

pngwn:
I sympathise with the view that this should be handled by the user but on the same token it is a confusing error if you don't really know much about these hidden files. It is a very simple fix that would improve the experience for a number of users. I don't really think this amounts to a large filter of files but if that trend begins to develop we can assess it when it does.

russelgal:
I think there is too much discussion for such a small fix )

benmccann:
@russelgal I think the reason you're getting a 404 is because sirv doesn't serve dot files by default. If we're going to change the code here, there's an argument to be made that we should ignore all dot files instead of just `.DS_Store`. Though it's unclear if we should be making that decision on behalf of our users as some may have a use case for serving a particular dot file. Another solution to avoid the 404 would be to specify [sirv's `dotfiles` option](https://github.com/lukeed/sirv/tree/master/packages/sirv#optsdotfiles):

```
sirv('static', { dev, dotfiles: true }),
```

Does that work for you?

russelgal:
My deploy workflow:
1. run build production on mac
2. auto upload build folder on vps (excluding macOS .DS_Store files)
3. graceful reload app with PM2

Hence dotfiles serving has nothing to do with it

Conduitry:
If your deployment workflow specifically involves not copying certain files that were present in `static` during the build, them being missing on the server and breaking the service worker shouldn't be much of a surprise. I still maintain that singling out one file in Sapper for special handling, and making this behavior be baked into Sapper itself, is weird.

antony:
@Conduitry yep that's exactly the issue. This would happen to any sort of file.

I think we need clearer documentation around how the basic service-worker operates, or dare I say it, comments in the code around exactly what the `files` code does. It depends on how we feel about the service worker being more than just a starter-for-ten.

But this PR doesn't fix the issue it is designed to (because there isn't an issue) - it's a user misunderstanding, so I think it should be closed.

russelgal:
Okay, I'll continue to fix this file manually, should we close this PR?

antony:
@russelgal just delete the file *before* build and you won't have to worry about it.

PR #1109, opened 2020-03-06
Add query string versioning for static assets in build process

nipinium:
This makes browser and cloudflare refresh the cache when the content of `global.css` or `favicon.ico`  changed.

### DIFF
--- a/src/api/build.ts
+++ b/src/api/build.ts
@@ -9,6 +9,7 @@ import { noop } from './utils/noop';
 import validate_bundler from './utils/validate_bundler';
 import { copy_runtime } from './utils/copy_runtime';
 import { rimraf, mkdirp } from './utils/fs_utils';
+import add_version_to_assets from "./utils/add_version_to_assets";
 
 type Opts = {
 	cwd?: string;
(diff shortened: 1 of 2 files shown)


benmccann:
The tests seem to be failing with this PR

I think I would instead prefer that the cache TTL is set to 0 on dynamic assets and forever on static assets along with using webpack and rollup to inject hashes into the static asset file names

benmccann:
https://github.com/sveltejs/sapper/pull/1173 seems like it might solve this issue for you as well and is probably going in a better overall direction

nipinium:
Yeah the tests were failed, I didn't even know the reason: it worked when I ran it using `yarn dev`, but it failed every time I ran `yarn test`, maybe because of the linting step, maybe it just didn't work using yarn.

Anyway, I figured it's better to make a post-build script instead of make a modification in sapper source code, so I kinda abandoned this PR.

Should I just leave it here for reference purpose, or just close it already?

benmccann:
Probably better to close it if it won't be pursued. It'll make it easier to keep track of which PRs need review. And it can still be referenced even if closed

nipinium:
Closed it.

PR #1118, opened 2020-03-13
Fix 2 issues with custom html files in export

jamesjnadeau:
1. While exporting, if a route such as `[slug].html.js` is made that serves it's own html pages(with correct Content-Type header), those pages url's were being rewritten even though they were valid. Example: `simple.html` was being rewritten to `simple.html/index.html`. We now check for a `.html` on the end of the path before adding a `/index.html` to the end

2. When parsing exported html, the `<base>` element was checked for in the head of the doc. If this element didn't exist(for example if you generate your own html and miss adding this element) the error message is not clear what is wrong and is hard to track down. This change should avoid that confusion and use a sane default if not available.

### Tests
-  [*] Run the tests tests with `npm test` or `yarn test`)
I added a `test.html` file to the export test that should test both of these scenarios


src/api/export.ts
```diff
@@ -186,7 +187,7 @@ async function _export({
 					const cleaned = clean_html(body);
 
 					const base_match = /<base ([\s\S]+?)>/m.exec(cleaned);
-					const base_href = base_match && get_href(base_match[1]);
+					const base_href =  base_match ? get_href(base_match[1]) : '/'
```
benmccann, line 190: nit: don't need the extra space before `base_match` and should add a semi-colon to the end of the line

src/api/export.ts
```diff
@@ -112,7 +112,8 @@ async function _export({
 		const is_html = type === 'text/html';
 
 		if (is_html) {
-			if (pathname !== '/service-worker-index.html') {
+			if (pathname !== '/service-worker-index.html'
+			   && pathname.substr(pathname.length - 5) !== '.html') {
```
benmccann, line 116: maybe we don't need the `pathname !== '/service-worker-index.html'` check anymore since this would cover it?

Conduitry:
Some portion of this has been resolved in the recently merged #1049 - which is also part of what's making this PR conflict. So that should be bullet point 1 addressed.

Conduitry:
Issue #1202 has been opened, which sounds like the second, unaddressed issue this PR is intended to fix.

src/api/export.ts
Conduitry: Yep, I did the same tidying in #1049 before merging that in.

jamesjnadeau:
This patch looks wonky now, when all I did was try to merge in master and rebase, oh well, i'm going to close this and submit the one liner in another pull request so it's easier for you folks to review.

PR #1123, opened 2020-03-16
move %sapper.scripts% to <head> with defer

### LINKED ISSUE #1121: %sapper.scripts% should be in <head>
Unless there is something i totally misunderstand about the use of `defer`, i don't really get why scripts are still being placed at the end of the body.
[CanIUse](https://caniuse.com/#search=defer) shows a 98% compatibility rate of the `defer` attribute, the remaining 2% being mostly Opera mini.

In `template.html`, `%sapper.scripts%` should be located inside the `<head>` tag, with a `defer` attribute.

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
    <meta name="theme-color" content="#333333" />

    %sapper.base%

    <link rel="stylesheet" href="global.css" />
    <link rel="manifest" href="manifest.json" crossorigin="use-credentials" />
    <link rel="icon" type="image/png" href="favicon.png" />

    %sapper.styles%
    %sapper.head%
    %sapper.scripts%
  </head>

  <body>
    <div id="sapper">%sapper.html%</div>
  </body>
</html>
```

iOiurson:
This PR moves `%sapper.scripts%` to `<head>` with `defer` attribute for the main script.

[Some parts of the inserted scripts](https://github.com/sveltejs/sapper/blob/master/runtime/src/server/middleware/get_page_handler.ts#L270-L290) are inlined, in particular when using Rollup as the bundler.

Inline scripts are not affected by `defer`. 

But it seems in this case the inline scripts mainly import the main script dynamically,  thus it should be executed asynchronously.

Fixes #1121

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -289,7 +289,7 @@ export function get_page_handler(
 					script += `var s=document.createElement("script");try{new Function("if(0)import('')")();s.src="${main}";s.type="module";s.crossOrigin="use-credentials";}catch(e){s.src="${req.baseUrl}/client/shimport@${build_info.shimport}.js";s.setAttribute("data-main","${main}")}document.head.appendChild(s)`;
 				}
 			} else {
-				script += `</script><script src="${main}">`;
+				script += `</script><script src="${main}" defer>`;
 			}
 
 			let styles: string;
--- a/site/src/template.html
+++ b/site/src/template.html
@@ -38,15 +38,14 @@
 	<!-- This contains the contents of the <svelte:head> component, if
 	     the current page has one -->
 	%sapper.head%
+	<!-- Sapper creates a <script> tag containing `app/client.js`
+			 and anything else it needs to hydrate the app and
+			 initialise the router -->
+	%sapper.scripts%
 </head>
 <body>
 	<!-- The application will be rendered inside this element,
 	     because `app/client.js` references it -->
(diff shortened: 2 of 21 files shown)


artemjackson:
AppVeyor build failed cuz of `net::ERR_UNSAFE_PORT at http://localhost:2049`.
@Conduitry  Should we fix this unpredictability via `--explicitly-allowed-ports=2000` puppeteer flag?

benmccann:
@artemjackson https://github.com/sveltejs/svelte has moved to GitHub actions, so maybe we should do the same and just delete appveyor. I sent a PR to do the migration here: https://github.com/sveltejs/sapper/pull/1224

dzbarsky:
Can we merge this PR? It should be a nice perf win.

artemjackson:
@iOiurson could yu please rebase the PR against latest master? (`git pull master`)

Conduitry:
If this requires a corresponding change to a consumer's `src/template.html` to keep their app working, this is a breaking change.

benmccann:
Adding the `defer` attribute will stop the script from blocking, but `DOMContentLoaded` will still not fire until the script is finished executing. So if the user's `src/template.html` looks like the one in https://github.com/sveltejs/sapper-template/blob/master/src/template.html then this change will not affect the app. The case where it may be a breaking change would be if the user had modified their `template.html` to add a script without `defer` attribute after the Sapper script that both depended on it having run and wasn't listening for a page load event.

Issue #1134, opened 2020-03-30
Clarify in the docs that preload cannot guarantee access to either client-side-only or server-side-only information


natevaughan:
## Sapper `preload` has unpredictable behavior

Consider this common web application setup:
```
[browser]  < — > [api server]
     ^
     |
     v
[frontend server or CDN]
```
The api server may be running any technology and serving protected resources. The frontend server is Sapper, and is performing dynamic routing and/or server-side rendering, or may be a statically-generated sapper `export` hosted on a CDN.

In this setup, many use cases are not supported by Sapper `preload`:
 - Fetch protected data from the API server
 - Check a cookie and take a different action depending on the value set
 - Use a secret in an environment variable

These problems all stem from one fact: Sapper does not guarantee where `preload` will run. In a simple two-route application (say `index.svelte` and `me.svelte`), `preload` will run in the following places in the following scenarios:
 - User clicks a link from `index.svelte` to `/me` -> `preload` runs in browser
 - User clicks a link from another site to `/me` -> `preload` runs on server
 - User refreshes `/me` -> `preload` runs on server

This results following behavior:

 - Fetch protected data from the API server (fails on server, succeeds in browser)
 - Check a cookie and take a different action depending on the value set (fails on server, succeeds in browser)
 - Use a secret environment variable (succeeds on server, fails in browser; a naive dev might use `sapper-env` to solve and thus expose the secret to the client)

This can be a confusing experience that frustrates and alienates devs as svelte begins to gain more attention and adoption.

## Exacerbating the problem: solutions for narrow use-cases

Some proposed solutions work if an application happens to use Sapper as both an application's frontend server AND api server for all protected resources:
```
[browser]
     ^
     |
     v
[frontend server AND api server]
```
For example this thread https://github.com/sveltejs/sapper/issues/178 discusses and proposing solutions for authentication, but all solutions assume that the Sapper `server.js` will contain code that sets up and handles authentication and will use e.g. passport.js for configuring authentication.

## The solution: Clarify that `preload` should be used only for server-side rendering of static data and static site generation

If we zoom out from the narrow use-case (where Sapper happens to be all three of the frontend server, api server, and source of authentication), then `preload` should really only be used for one thing: server-side rendering of truly static data and static site generation.

Other Svelte features are available that are guaranteed to run only on the broswer (e.g. a function returning a promise and the `{#await}` block) and these should be emphasized for all use cases involving dynamic data from authenticated api servers.

[...]

PR #1135, opened 2020-03-30
Clarify assumptions made by `preload`

natevaughan:
Included here are a few assumptions that I stumbled over when getting to know `preload`. I've updated the docs to spell them out to hopefully save others a little bit of time.

See https://github.com/sveltejs/sapper/issues/1134

### DIFF
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -84,9 +84,9 @@ To fix this, Sapper provides `this.fetch`, which works on the server as well as 
 ```
 
 It is important to note that `preload` may run on either the server or in the client browser. Code called inside `preload` blocks:
-  - should run on the same domain as any upstream api servers requiring credentials; otherwise `credentials: 'include'` cannot guarantee access to 3rd party session cookies
+  - should run on the same domain as any upstream API servers requiring credentials; otherwise, `credentials: 'include'` cannot guarantee access to 3rd party session cookies
   - should not reference `window`, `document` or any browser-specific objects
-  - should not reference any api keys or secrets, which will be exposed to the client
+  - should not reference any API keys or secrets, which will be exposed to the client
 
 If you are using Sapper as an authentication/authorization server, you can use session middleware such as [express-session](https://github.com/expressjs/session) in your `app/server.js` in order to maintain user sessions.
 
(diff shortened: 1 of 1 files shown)


antony:
Thanks @natevaughan - this is very clear, and I think is a big tripping point for new users.

natevaughan:
Great, glad it helps. And thank _you_ for your help today @antony

natevaughan:
What's the Sapper release cycle like?

antony:
@natevaughan right now there isn't really one. We're focussed on Svelte for the time-being. We're slowly merging PRs as time allows.

site/content/docs/04-preloading.md
```diff
@@ -83,7 +83,12 @@ To fix this, Sapper provides `this.fetch`, which works on the server as well as
 </script>
 ```
 
-Note that you will need to use session middleware such as [express-session](https://github.com/expressjs/session) in your `app/server.js` in order to maintain user sessions or do anything involving authentication.
+It is important to note that `preload` may run on either the server or in the client browser. Code called inside `preload` blocks:
+  - should run on the same domain as any upstream api servers requiring credentials; otherwise `credentials: 'include'` cannot guarantee access to 3rd party session cookies
+  - should not reference `window`, `document` or any browser-specific objects
+  - should not reference any api keys or secrets, which will be explosed to the client
```
abastardi, line 89: "explosed" should be "exposed"

natevaughan: Done

site/content/docs/04-preloading.md
```diff
@@ -83,7 +83,12 @@ To fix this, Sapper provides `this.fetch`, which works on the server as well as
 </script>
 ```
 
-Note that you will need to use session middleware such as [express-session](https://github.com/expressjs/session) in your `app/server.js` in order to maintain user sessions or do anything involving authentication.
+It is important to note that `preload` may run on either the server or in the client browser. Code called inside `preload` blocks:
+  - should run on the same domain as any upstream api servers requiring credentials; otherwise `credentials: 'include'` cannot guarantee access to 3rd party session cookies
```
benmccann, line 87: shouldn't `credentials: 'include'` send the cookies even to other domains?

EddyVinck: @benmccann Doesn't that depend on how a project has their CORS settings configured? I'm pretty new to NodeJS so I might be mistaken

antony: cookies can't be cross-domain. Cookies from `.domain.com` will only ever be sent to `.domain.com`

Issue #1136, opened 2020-03-30
mime-types.md is missing mime-type for wasm file type


sajuthankappan:
**Describe the bug**
mime-types.md is missing mime-type for wasm file type. This causes 404 error when responding to .wasm files

**Logs**
No errors in console

**To Reproduce**
Try importing any wasm files. e.g., 
    const { greet } = await import("xyz-wasm");
    greet();

**Expected behavior**
responding with 200 code with the wasm file

**Information about your Sapper Installation:**
- Your browser and the version: FF74

- Your operating system: Ubuntu Linux 19.10

- Your hosting environment: Local

- Sapper version: 0.27.11

- Svelte version: 3.0.0

- Whether your application uses Webpack or Rollup: Webpack

**Severity**
blocking your usage of Sapper entirely?

**Additional context**
I checked by adding the following line to mime-types.md, and the wasm file is loaded successfully, in my local cloned code.
```
application/wasm				wasm
```

Conduitry:
Fixed in 0.27.12.

PR #1137, opened 2020-03-30
Fix missing wasm mime-type. Issue #1136

sajuthankappan:
Mime-types.md is missing the wasm file type, which ends up with 404 not found error, when importing wasm modules. This pull request adds mime-type for .wasm files and solves issue #1136


Conduitry:
Thanks, I've opened #1138 which I think is a better solution to this.

Conduitry:
Closing in favor of #1138.

Issue #784, opened 2019-07-02
Anchor linking and scroll position after refresh not working


staus:
Currently it seems the natural Sapper behaviour when you refresh a page is to jump to the top. I'm guessing this is because the dom doesn't exist initially so there is no dom to be scrolled down to.

Even more so I'm concerned this impacts direct linking to anchor links on a page as they neither seem to work naturally. I could check the url hash and do the scroll in code, but I'm concerned it might still hurt SEO score.

To replicate the issue, simply get the Sapper boilerplate and add a bunch of content so the page overflows

Unless I'm missing something I'd consider this a pretty big issue with sapper.

willdante:
Same here, keeping the scroll position when refreshig a page is very important in a website.

When changing route then going back to previous route with browser back button, the scroll position is correctly kept.
But not when refreshing a page.

How can we keep scroll position when refreshing ?

antony:
Just realised this is still an issue - anchor links just don't work as part of an "a" tag.

From a "refresh" point of view, it's clear - Sapper stores the `scroll_position`s in memory, so if you refresh the browser, the application reloads and the value is reset.

It would be good to understand the technical reasons why this doesn't work if you click an 'a' tag on a different page though. My content is very much static, stored inside a `const` variable in the page, so it is rendered on the server-side. This means that the position of the anchor tag would be known before the page renders in the browser.

How do other frameworks handle this? Nuxt? Next?

willdante:
Thanks for the info.

Is there an easy way to store the scroll position in local storage for example, or anywhere it wouldn't reset at refresh ?

antony:
Since Sapper is essentially serving an SPA, a refresh is "out of bounds" for its concern. The only reason a user should be refreshing is if the application broke or similar, so I can't see this scenario being handled/being a concern of Sapper.

willdante:
In my point of vue (maybe it won't be accurate as I'm relatively new to modern javascript and frameworks), Svelte is serving SPA and Sapper is here to handle SEO and multiples routes, which is required for websites.

When I'm on a website, I quite often refresh the page when a lazy loaded image fails to appear for example.
Or as a web developer, to test my lazy loading or other animations.

Going back to top when refreshing a page in a website is bad in a UX perspective.

So I would argue it should be a concern for Sapper, as it was for Nuxt which handles scroll position correctly 😉

staus:
[...]

PR #1139, opened 2020-04-01
Fix deeplinks no matter previous scroll position

ArkaneMoose:
Hello, and thanks for this awesome tool! I found what appears to be a bug in anchor deeplinks.

See a demo of the [broken behavior](https://arkanemoose.github.io/sapper-anchor-scroll-demo/broken/) and the [fixed behavior](https://arkanemoose.github.io/sapper-anchor-scroll-demo/fixed/).

If you're scrolled down on one page, and you click a deeplink that goes to a different page, the scroll position on the new page is wrong. This might be what is being referred to in the "anchor linking" part of #784, but I'm not quite sure.

This broken behavior seems to be because of [this line](https://github.com/sveltejs/sapper/blob/5e2c68606124a7c4407250c2889a7e23cf2c3d73/runtime/src/app/app.ts#L203) which computes the new scroll position according to the `getBoundingClientRect().top` of the deeplinked element. [According to MDN,](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) this value is relative to the top-left of the viewport, so if you want a position that is relative to the top-left of the document, you need to add the value of `window.scrollY`.

This PR implements that, and it also adds a test that verifies that the behavior is broken prior to this PR and fixed afterwards.

Let me know if there's anything you'd like me to change. Thanks!

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -200,7 +200,7 @@ export async function navigate(target: Target, id: number, noscroll?: boolean, h
 			if (deep_linked) {
 				scroll = {
 					x: 0,
-					y: deep_linked.getBoundingClientRect().top
+					y: deep_linked.getBoundingClientRect().top + scrollY
 				};
 			}
 		}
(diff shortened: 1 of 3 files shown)


Conduitry:
Released in 0.27.13, thanks!

PR #1154, opened 2020-04-14
Support ES6 Modules in Sapper

DominusVilicus:
Not sure if this will break commons, but I believe it should not. 

One more note, the __sapper__ build directory should contain a `package.json` that contains
```json
{
   "type": "commonjs"
}
```

### Tests

### DIFF
--- a/src/core/create_compilers/index.ts
+++ b/src/core/create_compilers/index.ts
@@ -41,7 +41,7 @@ export default async function create_compilers(
 	}
 
 	if (bundler === 'webpack') {
-		const config = require(path.resolve(cwd, 'webpack.config.js'));
+		const { default: config } = await import('file:' + path.resolve(cwd, 'webpack.config.js'))
 		validate_config(config, 'webpack');
 
 		return {
--- a/src/core/create_compilers/index.ts
+++ b/src/core/create_compilers/index.ts
@@ -41,7 +41,7 @@ export default async function create_compilers(
 	}
 
 	if (bundler === 'webpack') {
-		const { default: config } = await import('file:' + path.resolve(cwd, 'webpack.config.js'))
+		const { default: config } = await import('file://' + path.resolve(cwd, 'webpack.config.js'))
 		validate_config(config, 'webpack');
 
 		return {


DominusVilicus:
Hey, just would like to leave some notes.

I added support for ES6 node 13 ecmascript modules. (`import` instead of `require()`)

I'm pretty sure this doesn't break existing commonjs integrations, but I also had to add a `package.json` to the __sapper__ build directory which contains
```json
{
   "type": "commonjs"
}
```

This file should be added to the __sapper__ build directory, because if it isn't there then it will say module is not defined in module.exports = ...

DominusVilicus:
Noticed it does create some errors because ES6 requires `file:` to be appended to the file paths. I might need to look into how to work around this.

DominusVilicus:
**Steps to reproduce**

add `"type": "module"` to `package.json` inside the template, update `webpack.config.js` requires to `import`s and then it will cause an error in `core.js` as it is trying to require the `webpack.config.js` but it cannot as it is using ES6 imports. Adding my code will fix the issue, but a `package.json` with type `commonjs` is required to prevent the app treating that folder as ES6

antony:
@DominusVilicus you've ticked the "relates to an existing bug" box but not said which bug.

Can you explain what this PR solves?

DominusVilicus:
@antony Sorry, there was no github issue.

The bug on the other hand is that sapper does not support ES6 Modules

Conduitry:
Whenever I've tried to use native ESM in Node, the experience has been sketchy enough that I immediately went back to compiling to CJS. I don't see a concrete benefit to this right now - it's not like we'd be able to avoid bundling, because all of the Svelte components still need to be compiled, and there's a lot of other stuff that needs to happen too. On the whole, this is not a complication I think it is remotely worth inviting at this time.

DominusVilicus:
@Conduitry ESM modules are supported by default in node 14

It works fine with the change I made

akvadrako:
There is anyway a fairly easy workaround if you want to use ES6 modules with webpack config:
```js
const xxx = require('esm')(module)('xxx').default;
```

Issue #1155, opened 2020-04-14
500 error on hard reload in fetch due to hard coded http://127.0.0.1:${process.env.PORT}


velusgautam:
**Describe the bug**
When reload the blog page the fetch request is trying the hard coded base path 
<img width="806" alt="Screenshot 2020-04-14 at 1 05 37 PM" src="https://user-images.githubusercontent.com/8556085/79198793-ad1c8680-7e51-11ea-9df1-a27d0fdcc6ec.png">

**Logs**
<img width="405" alt="Screenshot 2020-04-14 at 1 08 11 PM" src="https://user-images.githubusercontent.com/8556085/79198389-0801ae00-7e51-11ea-8c9e-4ec20c854e80.png">

**To Reproduce**
Goto below link and hard reload

https://sapper-btmh9v5am.now.sh/blog 

**Expected behavior**
It should request in the base path of the route instead of 127.0.0.1:3000

**Stacktraces**

**Information about your Sapper Installation:**
- Your browser and the version: Chrome Version 80.0.3987.163

- Your operating system: OS X 10

- Your hosting environment: Now

- Sapper version : ^0.27.0

- Svelte version:  "^3.0.0"
- If it is an exported (npm run export) or dynamic application.

- Whether your application uses  Rollup

**Severity**
Its a bug that is not allowing for all SEO and social media to render the 500 output.

velusgautam:
https://github.com/sveltejs/sapper/pull/1156

velusgautam:
<img width="1341" alt="Screenshot 2020-04-14 at 4 32 14 PM" src="https://user-images.githubusercontent.com/8556085/79218304-d3501f80-7e6d-11ea-8459-51cb16f4914b.png">
@Conduitry  We have this link in the production build. How can I fix that?

Can you please advise

antony:
@velusgautam This is a support question. Please use the https://svelte.dev/chat for support questions.

PR #1156, opened 2020-04-14
Fixed fetch hardcoded reload issue in now Bug: #1155

velusgautam:
Fixed the path to **req.headers.host** from  **http://127.0.0.1:${process.env.PORT}** to take it dynamic


Conduitry:
I don't think this is something we want to always be doing - we don't want `this.fetch` for local requests during SSR to be making DNS lookups and for these requests to be making a full roundtrip out to the real internet. Where Sapper is running might not even have full access to the real internet. We also definitely don't want to always be requesting `http://` - but it's impossible for us to know whether there is in fact an HTTPS reverse proxy in front of us.

PR #1165, opened 2020-04-21
Added pageStore to handle page changes better.

### LINKED ISSUE #633: Routes should unsubscribe from the page store before $page.params are updated
## Expected Behavior
A route autosubscribing to `$page.params` will be unsubscribed before `$page.params` is changed to a new route

## Actual Behavior
The autosubscription is being triggered late exactly once, i.e. if I navigate from `channels/[slug].svelte` to `games/[slug].svelte`, the Channel route's subscription will be triggered exactly once with the `slug` param for the Game route

pchynoweth:
- [x] Ideally, include a test that fails without this PR but passes with it. PRs will only be merged once they pass CI.
### Tests
-  [x] Run the tests with `npm test` or `yarn test`)

This fixes #633

The issue boils down to when the page store is updated.  It's important to issue the callback for the current page immediately when `page.subscribe()` is called.  However, existing subscribers should not receive callbacks until the root component has had the opportunity to remove the previous page.

I have achieved this by adding a custom store that will not notify subscribers immediately.  Instead, there are separate `set()` and `notify()` calls.  `notify()` can be called to notify existing subscribers some time after calling `set()`.  `notify()` gets called from the `after_update()` lifecycle hook on the root component.  I am not entirely sure how I feel about this part.  Setting the lifecycle hook in this way is not an officially supported part of the svelte API.  However, I am not sure how else to do this.  I thought since this is part of the sveltejs project it may be okay.  Let me know what you think.


Conduitry:
Thanks! I feel like a delayed-notification store should be implementable as a wrapper around a normal `writable` store, without duplicating its whole implementation in this repo.

Using a private Svelte API is a bigger problem though. and I wouldn't say that's acceptable to use as part of a solution. While it is unlikely this would change without the maintainers realizing it, we don't want to have to keep this internal note about a this usage, or be tied to avoiding refactoring this code in Svelte, or to have to worry about updating Sapper to work with both old and new versions of Svelte (those with this private API and those with whatever later API there ends up being).

There are other way (like using the `afterUpdate` hook within the root component) to add code to be called after each update. I poked around at this for a bit the other day after seeing this PR, hadn't gotten anywhere yet.

pchynoweth:
Thanks @Conduitry.  I've move the `afterUpdate()` call into the root component.  I'm not sure if that's entirely desirable given the comment a few lines up in that file.  It's definitely an improvement on what I had before.  I have also made pageStore a wrapper around a writeable store.  I'm not entirely happy with the subscribe function at the moment.  It seems a bit hacky to me.

pchynoweth:
Okay.  I've happier with the pageStore now.  Take a look and let me know what you think.  I'm probably missing a trick or two.

Issue #1113, opened 2020-03-09
Create a "who's using sapper" section


stephane-vanraes:
Should we make a section on the homepage for "who's using Sapper" ?
Similar to the one on https://svelte.dev/ ?

This would point people that are interested in using Sapper to actual real world examples.

antony:
I think we'd be open to this. Do you fancy creating a PR for said page?

kevmodrome:
You could make a copy of the who's using svelte file in the community repo to use with it, if that makes sense.

TheComputerM:
Can I add https://svelte-materialify.vercel.app/ to this section? This is not a company but a framework.

kevmodrome:
I don't think that is a good fit @TheComputerM. Imagine what would happen if it was merged. Any and all libraries/frameworks would need to be added.

TheComputerM:
I understand, but just to be clear I was focused on the website as it is made in sapper and is pretty large and complex to be included in a showcase.

PR #1169, opened 2020-04-24
Add "Who is using Sapper" section to the website.

stephane-vanraes:
This adds a _Who's using Sapper ?_ section to the website as proposed in https://github.com/sveltejs/sapper/issues/1113

The actual information is fetched from [svelte/community](https://github.com/sveltejs/community/tree/master/whos-using-sapper) which at the moment is empty, I think it would be better to add some sites there before merging this.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -9,7 +9,8 @@
     "predeploy": "npm run export",
     "cy:run": "cypress run",
     "cy:open": "cypress open",
-    "test": "run-p --race dev cy:run"
+    "test": "run-p --race dev cy:run",
+    "update": "node scripts/update_whos_using.js"
   },
   "dependencies": {
     "@polka/send": "^0.4.0",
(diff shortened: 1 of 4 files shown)


benmccann:
could we fetch this info via API instead?

otherwise I agree that a shell script seems cleaner than using `shelljs`

stephane-vanraes:
If we change this we shoudl probably change it for the svelte site itself as well

benmccann:
Thanks for putting together this PR! Sorry it sat for awhile. There's probably not much point to collecting a list of Sapper users now that SvelteKit is out. We've improved a bit the "Who is using Svelte?" on the svelte.dev homepage. We haven't talked much about whether we'd want to try to do the same for SvelteKit or not. In any case, I'm going to go ahead and close this since I doubt we'll be doing it for Sapper

Issue #567, opened 2019-02-16
Feature request: custom cache-control headers


nolanlawson:
I think I mentioned this in a previous issue, but it would be nice to be able to override Sapper's built-in cache-control headers. Sometimes you want to add `public`, or sometimes you want to increase or decrease the `max-age`, etc.

Currently I can work around this, but it's extremely hacky:

```js
function overrideSetHeader (req, res, next) {
  const origSetHeader = res.setHeader
  res.setHeader = function (key, value) {
    if (key === 'Cache-Control') {
      if (value === 'max-age=31536000, immutable') { // webpack assets
        return origSetHeader.apply(this, ['Cache-Control', 'public,max-age=31536000,immutable'])
      }
      if (value === 'max-age=600') { // HTML files
        return origSetHeader.apply(this, ['Cache-Control', 'public,max-age=3600'])
      }
    }

    return origSetHeader.apply(this, arguments)
  }
  return next()
}

app.use(overrideSetHeader, sapper.middleware())
```

If there's some Express/Polka-standard way for middleware to do this then I'd be happy with that; else maybe it could just be an option passed in to `sapper.middleware()`?

antony:
This has become mission critical for me I think. The problem is that whenever I deploy, cloudflare/browsers cache the old pages with old asset references for 10 minutes, which appears to mean that the site runs, but in some weird half-broken state for all my users, who become immensely irritated since buttons stop clicking, dropdowns stop loading etc.

I would assess therefore that this feature is very high priority, since it makes reliable continuous delivery of a sapper application nearly impossible right now.

The workaround looks good - I will implement this on my application asap, since right now I've just had to turn the CDN off which is far from ideal.

kylecordes:
This "serving a mix of old and new" is already well solved for the JavaScript served by Sapper: hashed URLs.

Maybe a good solution is to enhance Sapper, or the default configuration thereof, to also hash asset URLs? Many popular websites are served this way (with hashed asset URLs for images etc, not only JS) to prevent this kind of problem.

(As part of such a solution (I don’t know if there is a good Sapper answer for this part yet, even for JavaScript URLs) is that when deploying a new version, old asset URLs must also continue to serve their old contents for a while, to support pages already running in browsers, overly enthusiastic caching layers, etc.)

alexdilley:
The hard-set cache header stumped me when implementing the not-too-uncommon OAuth flow, à la:

Visit `/` (`Cache-Control: max-age=600`);
[...]

Issue #951, opened 2019-10-21
Images in static folder aren't cached


hgl:
**Is your feature request related to a problem? Please describe.**

Images in the static don't have a far future cache header attached. That's understandable, because they don't have a hash in their file names. But that means they are not cached. Tweaking sirv doesn't help much, because a far future cache is impossible with hashes in names.

**Describe the solution you'd like**
It'd be great if sapper could take assets like images under the wing and serve them the same as compiled js and css files, with names postfixed by hashes and a far future cache.

**How important is this feature to you?**
Not a show stopper.

hgl:
I tried out nuxt.js today, and when you write urls in certain way, e.g., `~/assets/image.png`, they go through webpack and hash names are automatically added. Wish sapper could offer a similar feature.

benmccann:
I think this is maybe asking for https://github.com/sveltejs/svelte-loader/issues/114 ?

hgl:
Maybe both issues actually belong to svelte. To be able to import assets, the svelte compiler should generate those import statements. svelte-loader simply makes the generated code compatible with weback/rollup.

benmccann:
You can get the hash in the file name by manually writing the import statements as described in https://github.com/sveltejs/svelte-loader/issues/114. It's a little tedious to do that and hopefully https://github.com/sveltejs/svelte-loader/issues/114 will get addressed at some point to make it better. But until then, while you're doing it manually, we should still be able to set the cache-control headers appropriately on static content with a hash in the file name.

The docs say:
>Sapper doesn't serve these files — you'd typically use  or  for that — but it will read the contents of the static folder so that you can easily generate a cache manifest for offline support (see ).

So I think there's not actually anything to implement in Sapper. The hashing is done by the loader/bundler and the serving is done by another middleware or server.

We can leave this ticket open for the time being though because I think what would be useful is documenting how to get the hash in the filename with the official bundler plugins (i.e. webpack and rollup) or linking to the appropriate section in those docs. https://github.com/sveltejs/svelte-loader/issues/114 shows how to do it in webpack. I'm not sure about rollup. I think you need an extra rollup plugin for that potentially?

benmccann:
[I raised an issue in the template](https://github.com/sveltejs/sapper-template/issues/229) to add hashes to the filenames and provided the code to do so there. If I don't hear any objections I'll send it as a PR.

PR #1173, opened 2020-04-28
remove cache-control header

gka:
This PR resolves the [outstanding `maxage` header issue](https://github.com/sveltejs/sapper/issues/567). A fixed `maxage` header is not practical. There should be either a way to customize the maxage, or, more simple, the Sapper middleware shouldn't even set it at all.

The solution is to remove the `res.setHeader('Cache-Control', dev ? 'no-cache' : 'max-age=600');` line and the corresponding test.

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -54,7 +54,6 @@ export function get_page_handler(
 		 } = get_build_info();
 
 		res.setHeader('Content-Type', 'text/html');
-		res.setHeader('Cache-Control', dev ? 'no-cache' : 'max-age=600');
 
 		// preload main.js and current route
 		// TODO detect other stuff we can preload? images, CSS, fonts?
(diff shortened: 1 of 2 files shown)


benmccann:
The current behavior seems extremely severely broken to me. I understand that folks will want to further customize this, but getting this PR merged should be high priority to unbreak things. We can then take our time figuring out how to let users provide customized caching strategies after that. I added some thoughts in https://github.com/sveltejs/sapper/issues/951 and agree very much with what @kylecordes said https://github.com/sveltejs/sapper/issues/567#issuecomment-487041326

There's a number of users reporting issues in https://github.com/sveltejs/sapper/issues/567 with well thought out and described issues. This includes Sapper regulars such as @antony who described this as "mission critical" and "very high priority" because the current behavior results in "weird half-broken state for all my users" https://github.com/sveltejs/sapper/issues/567#issuecomment-487024015

benmccann:
@antony is this a breaking change? It changes behavior, but I'm having a hard time envisioning a scenario where it's not backwards compatible or breaks something

antony:
@benmccann if you remove the cache-control header (which a number of CDNs use to decide how to cache your pages) it will certainly change the way your site operates, which I'd suggest is breaking.

PR #1191, opened 2020-05-13
Sapper FAQ

antony:
Identikit Sapper FAQ

### Tests

### Then
- [ ] Remove FAQ item from Wiki and redirect

### DIFF
--- a/site/src/routes/faq/index.svelte
+++ b/site/src/routes/faq/index.svelte
@@ -32,7 +32,7 @@
 			<p>{@html faq.answer}</p>
 		</article>
 	{/each}
-<p>See also the <a href="https://svelte.dev/faq" rel="external">Svelte FAQ</a> for questions relating to Svelte directly.</p>
+	<p>See also the <a href="https://svelte.dev/faq" rel="external">Svelte FAQ</a> for questions relating to Svelte directly.</p>
 </div>
 
 <style>
--- a/site/src/routes/faq/index.svelte
+++ b/site/src/routes/faq/index.svelte
@@ -1,0 +1,89 @@
+<script context="module">
+	export async function preload() {
+		const faqs = await this.fetch(`faq.json`).then(r => r.json());
+		return { faqs };
+	}
+</script>
+
+<script>
+	const description = "Frequently Asked Questions about Sapper"
+
+	export let faqs;
+</script>
+
+<svelte:head>
+	<title>Frequently Asked Questions Sapper</title>
+
+	<meta name="twitter:title" content="Sapper FAQ">
+	<meta name="twitter:description" content={description}>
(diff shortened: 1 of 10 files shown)


site/src/utils/slug.js
```diff
@@ -0,0 +1,78 @@
...
+	slugify(string, {
+		customReplacements: [	// runs before any other transformations
+			['$', 'DOLLAR'], // `$destroy` & co
+			['-', 'DASH'], // conflicts with `separator`
+		],
+		separator: SLUG_SEPARATOR,
+		decamelize: false,
+		lowercase: false
+	})
+		.replace(/DOLLAR/g, '$')
```
benmccann, line 16: nit: I might indent this line and the next one less

benmccann:
Maybe we should also have a pointer to the Svelte FAQs on this page? Most of them would be relevant for Sapper users. Before users would automatically see them because both FAQs were on one page

site/src/utils/slug.js
antony: I didn't actually touch that file at all, just copypasta from the Svelte site, so I'm going to leave it as-is :)

antony:
@benmccann and vice-versa perhaps

benmccann:
Doesn't necessarily hurt, but I'm not sure it's as important in the reverse direction. All users of sapper are users of svelte, but not vice versa. But it never hurts to add more documentation and links!

benmccann:
@antony just a quick reminder that it might be useful to add a link to the svelte faq here if you get a chance

site/src/routes/faq/index.svelte
```diff
@@ -0,0 +1,89 @@
...
+		<article class='faq'>
+			<h2>
+			<span id={faq.fragment} class="offset-anchor"></span>
+			<a class="anchor" rel='prefetch' href='faq#{faq.fragment}' title='{faq.question}'>&nbsp;</a>
+			{faq.metadata.question}
+			</h2>
+			<p>{@html faq.answer}</p>
+		</article>
+	{/each}
+</div>
```
antony, line 35:
```suggestion
<p>See also the <a href="https://svelte.dev/faq" rel="external">Svelte FAQ</a> for questions relating to Svelte directly.</p>
</div>
```

Conduitry:
Part of the filename is used in the hash portion of the URL for deeplinking to that question, correct? We should have something more correct here than `100-im-new-to-svelte.md`. I don't have a strong opinion about what this should be.

antony:
> Part of the filename is used in the hash portion of the URL for deeplinking to that question, correct? We should have something more correct here than `100-im-new-to-svelte.md`. I don't have a strong opinion about what this should be.

Oops! Fixed.

Issue #1142, opened 2020-04-02
Route with address "/client/" is not working


MrSrsen:
**Describe the bug**
When I use `/client/*` route in Sappers file-system routing I am getting plain text error `not found`.

**To Reproduce**

0. Launch Sapper in `dev` mode (I didn't test this with other then `dev`)
1. Create file in `/routes/client/new.svelte` with some content
2. Visit this route in browser as `localhost:3000/client/new`
3. You will see mentioned plain text error

**Expected behavior**
I see contents of `new.svelte` page.

**Information about your Sapper Installation:**
- Sapper: 0.27.9
- Svelte: 3.13
- Browser: Firefox 74.0
- OS: Kubuntu 18.04

**Severity**

Low - annoying.

**Additional context**

I think this is because of `/client` native Sappers files that are automatically generated by compilation but if this is the case i think some compile-time error or warning would be appropriate.

If this is intended behavior I think this needs to be documented in the docs. I didn't see any "reserved" route names there (https://sapper.svelte.dev/docs/).

antony:
Confirmed. And your supposition is correct!

Conduitry:
I believe this would be fixed by changing the exception handling [here](https://github.com/sveltejs/sapper/blob/facbd96e0bd2539acf32c5ac14cd9478c7c635e5/runtime/src/server/middleware/index.ts#L124) to call `next()` to continue with the next middleware rather than immediately returning a 404.

m-dary-nur:
I also confused before, when i use routes /client/input and /client/input/id, its work fine, but after reload browser on this route i got not found, its fine in otherroute case like /user/input or /user/input/id.

PR #1197, opened 2020-05-15
Pull request to resolve issue #1442

jervtub:
https://github.com/sveltejs/sapper/issues/1142
Route with address "/client/" is not working.

The first request to, for instance "http://localhost/client/new" would
pass the filter function, as [`req.start.startsWith(prefix)`](https://github.com/sveltejs/sapper/blob/facbd96e0bd2539acf32c5ac14cd9478c7c635e5/runtime/src/server/middleware/index.ts#L104) would match with "/client/".

I assume all the exported files by sapper have an extension, such as "chunk_slug.js" or "slug.svelte". Therefore, adding a new rule to the filter, to ensure each path of a request contains an extension, should resolve the issue.

I have add a test and in addition made sure it both fails _without_ the changes and succeeds _with_ the changes.


runtime/src/server/middleware/index.ts
```diff
@@ -99,9 +99,14 @@ export function serve({ prefix, pathname, cache_control }: {
 	pathname?: string,
 	cache_control: string
 }) {
+	// Filter requests based on req.path.
 	const filter = pathname
 		? (req: Req) => req.path === pathname
-		: (req: Req) => req.path.startsWith(prefix);
+		/*
+			* [#1442](https://github.com/sveltejs/sapper/issues/1142)
```
benmccann, line 106: nit: it looks weird that this line is indented more than the previous line that starts the comment. also, usually I would just use the `//` syntax for a two line comment in the middle of the code

jervtub: Thank you for your feedback, I have withdrawn the pull request to improve it. :)

benmccann: you can update existing requests. you just make the change, run `git commit`, and then `git push` and it will update the pull request

jervtub: didnt' work out yet, will continue this evening..

jervtub: > you can update existing requests. you just make the change, run `git commit`, and then `git push` and it will update the pull request

I'll keep that in mind for any future pull requests I want to update. For now, I took a large detour on updating the pull request..

PR #1206, opened 2020-05-17
docs: improve legacy flag description

jthegedus:
I had to ask in Discord what this flag actually did as I couldn't find it documented anywhere.


src/cli.ts
```diff
@@ -197,7 +197,7 @@ prog.command('export [dest]')
 	.option('--host', 'Host header to use when crawling site')
 	.option('--concurrent', 'Concurrent requests', 8)
 	.option('--timeout', 'Milliseconds to wait for a page (--no-timeout to disable)', 5000)
-	.option('--legacy', 'Create separate legacy build')
+	.option('--legacy', 'Creates a separate legacy build for older browsers')
```
pngwn, line 200: This still repeats the option's name to a degree, I'd opt for something even simpler: 

`"Creates an additional build compatible with older browsers"`

jthegedus: As I understand this additional build is only served to older browsers, I have used that language

pngwn: You used the word `legacy` again, isn't that the term that was confusing in the first place?

jthegedus: I wanted to know what was meant by legacy and how that would affect my app if I removed it. The key information is that it's an `additional` build and (potentially) served only to older browsers

benmccann:
This is very helpful as is. I think it'd be even better if we could define what "older browsers" are. Something like "Creates an additional JS bundle for older browsers which are not auto-updated. Primarily IE 10+" (I'm not familiar enough with this to know if that's right - just an example)

Conduitry:
What's currently considered a legacy browser for this purpose is whether `eval("async function x(){}");` throws. (This has its own problems, because old Edge supports async functions but not object spread, but that's a separate issue.) I'm not sure whether it makes sense to specifically tell people 'creates a separate build for browsers which do not support async/await' right in the CLI option description.

pngwn:
What even is a browser Ben. 

But yes, we do need to improve the Sapper documentation in general.

benmccann:
I'll also add that as a user, the first time I saw the flag, I thought it was doing some legacy version of the build process, which was really confusing to me. I wonder if we should name it something like `--legacyBrowsers` - though that would be a breaking change

pngwn:
I think good documentation can solve this.

benmccann:
I sent https://github.com/sveltejs/sapper/pull/1209 to add some more docs

jthegedus:
To be clear, is the CLI output considered as docs?

pngwn:
Kind of but not really.

jthegedus:
So I should close this PR?

pngwn:
No, I think is a good clarification. We should add both.

Issue #876, opened 2019-09-06
What are the Supported Browsers?


benwinding:
What's the intended browser compatibility of **sapper**? Can't find it written anywhere...

Would be nice if we could have a chart like this:

![01-svg-support-table](https://user-images.githubusercontent.com/11782590/64391882-91662c00-d089-11e9-8eb8-16b3fc0aeb69.jpg)

Wolfr:
You use Sapper to write code that has its own browser compatibility? As I understand it Sapper's browser compatibility is not relevant as it is the tool you use to help you write the code and itself it doesn't output any code that has browser compatibility issues.

kylecordes:
Browser support is definitely an important question. At some point a while back a fresh default Sapper app did not work in IE 11. There is been considerable discussion of the merits of such support.

(My opinion is split: it is both a moral abomination and a business necessity. ;-) )

Support would mean that if you use the swath of available features in Sapper and Svelte, the default tooling emits any necessary polyfills, compile target settings, etc. to work correctly across the set of supported browsers. Out of the box. Although this is on one hand a bunch of annoying work, it is the bar that is set by the current crop of popular frameworks.

Wolfr:
Right, thanks for the info @kylecordes , I am new to Sapper. I wanted to make some prototypes but some of my clients want to support IE11.

So it's good to know that the default bundle setup might not specifically work in IE11 and you shouldn't use Sapper if you want to support legacy browsers.

benwinding:
So I was using the `fetch()` api and was hoping it would be transpired to work on older platforms, but it wasn't...

Is there a way to target older browsers too? Would **babeljs** help here? Incorporated into Webpack or something?

fvsch:
@benwinding:

You need a bit more than transpiling to make fetch happen in IE. Fetch is not syntactic sugar, it's an API provided by the platform (a web browser). You can serve IE or all browsers a script that polyfills the Fetch API using XMLHttpRequest and Promises, such as https://www.npmjs.com/package/whatwg-fetch

As their docs say, to support IE you will also need a Promise polyfill. Not sure if Babel handles this.

> Would babeljs help here?

[...]

Issue #739, opened 2018-08-31
Document --legacy 


Rich-Harris:
https://github.com/sveltejs/sapper/issues/280

kylecordes:
Here is an alternative idea. Rather than document this option, consider removing the option. Instead a "production" build would automatically include differential bundling. In the future when there are almost no browsers left that need the old code, an option could be introduced to `--skip-legacy`.

Or maybe by then the ES5 output could be removed, and no option (thus no documentation other than "**of course** Sapper does differential bundling, you need not even ask") would ever be needed.

sw-yx:
@Conduitry  - may i ask why this was transferred over from the docs repo? 

i'm going to open up a WIP PR for this one and we can discuss documenting there

Conduitry:
This is the docs repo now as well. <https://sapper.svelte.dev/> lives in the site directory of this repo. Since Sapper was updated for Svelte 3, the other repo is for what's at <https://sapper-legacy.svelte.dev/>.

sw-yx:
oh.. lol ok

sw-yx:
ok i've moved over the PR!

diverent2:
Here's another idea: How about using `--support-legacy` instead?
IMO this clarifies that it's meant for browser support without getting too verbose or causing confusion.

Just an idea though :)

sw-yx:
please open a new issue, i gave up on my pr

PR #1209, opened 2020-05-17
Add docs for --legacy flag

benmccann:
Sending this based off the discussion in https://github.com/sveltejs/sapper/pull/1206

Closes https://github.com/sveltejs/sapper/issues/876
Closes https://github.com/sveltejs/sapper/issues/739

This isn't really the whole story for legacy browser support, but it's a good start. I think a lot of the docs for cross-browser support should go in the Svelte docs, which is why I didn't try to add more here yet.

Questions I have:
* Does Sapper do anything special or is this a straight pass-through to Svelte's `--legacy` option?
* Should we mention that a legacy browser is one that doesn't support `async`? (https://github.com/sveltejs/sapper/pull/1206#issuecomment-629801152). I think it'd be useful for people to know exactly which browsers are in each bundle. Where's the code that does this?

### DIFF
--- a/site/content/docs/09-building.md
+++ b/site/content/docs/09-building.md
@@ -16,9 +16,9 @@ node __sapper__/build
 
 ### Browser support
 
-By default, Sapper builds your site only for the latest versions of modern evergreen browsers.
+Your site is built only for the latest versions of modern evergreen browsers by default. If you are using Rollup, you can use the `--legacy`[1] flag to build a second bundle for legacy browsers like Internet Explorer. Sapper will then serve up the correct bundle at runtime[2].
 
-If you are using Rollup, you can use the `--legacy` flag to build two bundles: one for modern browsers and one for legacy browsers like Internet Explorer[1]. Sapper will then serve up the correct bundle[2] at runtime.
+When using `--legacy`, Sapper will pass an environment variable `SAPPER_LEGACY_BUILD` to your Rollup config. Sapper will then build your client-side bundle twice: once with `SAPPER_LEGACY_BUILD` set to `true` and once with it set to `false`. [sapper-template-rollup](https://github.com/sveltejs/sapper-template-rollup) provides an example of utilizing this configuration.[3]
 
 You may wish to add this flag to a script in your `package.json`:
 ```
@@ -27,5 +27,6 @@ You may wish to add this flag to a script in your `package.json`:
   },
 ```
 
-[1] You will also need to polyfill APIs that are not present in older browsers.
+[1] This option is unrelated to Svelte's `legacy` option
 [2] Browsers which do not support `async/await` syntax will be served the legacy bundle
+[3] You will also need to polyfill APIs that are not present in older browsers.
--- a/site/content/docs/09-building.md
+++ b/site/content/docs/09-building.md
@@ -18,7 +18,7 @@ node __sapper__/build
 
 By default, Sapper builds your site only for the latest versions of modern evergreen browsers.
 
-If you are using Rollup, you can use the `--legacy` flag to build two bundles: one for modern browsers and one for legacy browsers like Internet Explorer. Sapper will then serve up the correct bundle at runtime.
+If you are using Rollup, you can use the `--legacy` flag to build two bundles: one for modern browsers and one for legacy browsers like Internet Explorer[1]. Sapper will then serve up the correct bundle[2] at runtime.
 
 You may wish to add this flag to a script in your `package.json`:
 ```
--- a/site/content/docs/09-building.md
+++ b/site/content/docs/09-building.md
@@ -18,10 +18,13 @@ node __sapper__/build
 
 By default, Sapper builds your site only for the latest versions of modern evergreen browsers.
 
-For older browsers, like Internet Explorer, you will need to use the `--legacy` flag:
+If you are using Rollup, you can use the `--legacy` flag to build two bundles: one for modern browsers and one for legacy browsers like Internet Explorer. Sapper will then serve up the correct bundle at runtime.
 
+You may wish to add this flag to a script in your `package.json`:
 ```
-npx sapper build --legacy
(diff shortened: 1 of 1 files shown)


site/content/docs/09-building.md
```diff
@@ -12,4 +12,16 @@ The output is a Node app that you can run from the project root:
...
 node __sapper__/build
-```
\ No newline at end of file
+```
+
+### Browser support
+
+By default, Sapper builds your site only for the latest versions of modern evergreen browsers.
+
+For older browsers, like Internet Explorer, you will need to use the `--legacy` flag:
```
antony, line 21: I think we should be a bit more explicit about the fact that this will build *two* bundles and serve up the correct one at runtime. Otherwise people might think you need to build the app twice, one with the flag and one without, which won't work at all.

antony: Thinking about it we should also note that this only works for `Rollup` builds.

antony:
> Should we mention that a legacy browser is one that doesn't support async? (#1206 (comment)). I think it'd be useful for people to know exactly which browsers are in each bundle. Where's the code that does this?

Can't answer your first question, but there isn't a comprehensive list of browsers that are / aren't supported. The check does indeed just use an `async/await` test, which might not be ideal

site/content/docs/09-building.md
benmccann: That's an excellent call out I was not aware of! Is there anyway to support IE in Webpack?

antony: No, I think it was planned but other things took precidence.

To be honest I don't know how long we plan to support webpack for, it's kind-of a legacy decision which is no longer necessary.

site/content/docs/09-building.md
```diff
@@ -12,4 +12,19 @@ The output is a Node app that you can run from the project root:
...
 node __sapper__/build
-```
\ No newline at end of file
+```
+
+### Browser support
+
+By default, Sapper builds your site only for the latest versions of modern evergreen browsers.
+
+If you are using Rollup, you can use the `--legacy` flag to build two bundles: one for modern browsers and one for legacy browsers like Internet Explorer. Sapper will then serve up the correct bundle at runtime.
```
antony, line 21:
```suggestion
If you are using Rollup, you can use the `--legacy` flag to build two bundles: one for modern browsers and one for legacy browsers like Internet Explorer[1]. Sapper will then serve up the correct bundle[2] at runtime.
```

site/content/docs/09-building.md
```diff
@@ -12,4 +12,19 @@ The output is a Node app that you can run from the project root:
...
+
+You may wish to add this flag to a script in your `package.json`:
+```
+  "scripts": {
+    "build": "sapper build --legacy",
+  },
+```
+
+You will also need to polyfill APIs that are not present in older browsers.
```
antony, line 30:
```suggestion
[1] You will also need to polyfill APIs that are not present in older browsers.
[2] Browsers which do not support `async/await` syntax will be served the legacy bundle
```

Conduitry:
I'm trying to weigh the benefits of saying something more generally helpful here vs saying something more technically correct. What the `--legacy` flag actually does to your build is entirely in the user's hands by way of what they have in their `rollup.config.js`. The default sapper-template currently happens to swap in a given Babel configuration when building in legacy mode, but this is not set in stone and this is not truly Sapper's responsibility.

The more correct thing to say here would be that `--legacy` makes Sapper create two builds, with differing values of a given build-time environment variable, and serves one or the other depending on async/await support. Users can have this environment variable affect whatever they want in their Rollup config. I don't know how to convey all of this in a way that is correct and useful and unintimidating.

benmccann:
I think that's a really helpful callout and we should definitely mention something about the flag being passed to the bundler.

How does Sapper's `legacy` option interact with Svelte's `legacy` option? I assumed that if you specified `legacy` in Sapper that it would pass it to Svelte which was the main effect. But looking at the code I'm not sure it does get passed unless I'm missing where it happens?

Conduitry:
I don't believe `legacy` in Sapper is passed along to `legacy` in Svelte. Svelte's `legacy` flag is for avoiding generating code that doesn't transpile or polyfill well back to IE9-ish. Sapper's `legacy` build is (more or less) delivered to all versions of IE.

You'd probably have an awful time getting a Sapper app to run on versions of IE before 11 anyway, and `legacy` in Svelte isn't needed for IE11, so I don't think changing this default makes sense. (But, if someone wanted to, they could again do in in their bundler config, as this is not up to Sapper.)

benmccann:
Ok, I took a stab at incorporating this additional information

pngwn:
Lets add an additional note to this that the `legacy` build defined in the template's rollup config is just an example, and that users can tweak it to fit their requirements. The current wording suggest that this should 'just work' which may be true in some cases but is definitely false in others.

site/content/docs/09-building.md
```diff
@@ -12,4 +12,21 @@ The output is a Node app that you can run from the project root:
 
 ```bash
 node __sapper__/build
-```
\ No newline at end of file
+```
+
+### Browser support
+
+Your site is built only for the latest versions of modern evergreen browsers by default. If you are using Rollup, you can use the `--legacy`[1] flag to build a second bundle for legacy browsers like Internet Explorer. Sapper will then serve up the correct bundle at runtime[2].
```
pngwn, line 19: Maybe ins tead of:
```
If you are using Rollup, you can use the `--legacy`[1] flag to build a second bundle for legacy browsers like Internet Explorer.
```

We could try:

```
If you are using Rollup, you can use the `--legacy`[1] flag to build a second bundle that can be used to support legacy browsers like Internet Explorer.
```

Hinting at its purpose rather than suggesting it solves that particular problem.

benmccann: Sure. The suggested language sounds fine to me. Updated

PR #1212, opened 2020-05-19
Reduce indentation and rename variable

benmccann:
This breaks off part of https://github.com/sveltejs/sapper/pull/869 so that PR will be smaller and easier to understand

This does two things:
* Rename `r` to `response`
* Combine two `if` checks to reduce a level of indentation


Conduitry:
869 was merged all in one go, so closing this. Thanks.

Issue #356, opened 2018-08-13
Enable sapper to run on AWS Lambda


ispyinternet:
I've been able to setup sapper on AWS lambda without too much difficulty. Currently it can only be done using express and not polka as the server has to be wrapped using: https://github.com/awslabs/aws-serverless-express

However there would be a change required to sapper middleware to support this, but no sure exactly how to set this up.

The issue is that for the Lambda environment, the node server is setup to run on a file socket and not your traditional TCP port (e.g. 3000).

In sapper middleware this.fetch relies on node-fetch, and node-fetch makes local requests using the TCP port ( i.e. fetch('127.0.0.1:3000/page'). But in lambda we need to be able to fetch from the socket. node-fetch appear to be unwilling to support sockets, but got does (https://github.com/sindresorhus/got#comparison)

This would also need a mechanism to indicate to sapper whether to use or not, in my hack I use an environment variable.

https://github.com/sveltejs/sapper/blob/master/src/middleware.ts#L348-L382

```javascript
if( parsed.href.substr(7,9) == '127.0.0.1'
   && process.env.HTTP_SOCKET ) {
	let urlPath = parsed.pathname + parsed.hash;
	let socket = process.env.HTTP_SOCKET;;
	return got(`unix:${socket}:${urlPath}`, opts).then( res => ({ json: () => JSON.parse(res.body)}));
}
```

If the authors / maintainers could work out how best to integrate and support this that would be great

ispyinternet:
I have a working solution, that probably needs some more experienced eyes to look, although @Rich-Harris has indicated he's not sure got is the route he would want to take. However it does give some insight to the issue and what the solution needs to address:

https://github.com/sveltejs/sapper/compare/master...ispyinternet:master

ispyinternet:
I've worked on this issue a little more. My initial fix resorted to replacing node-fetch with got as got supports fetch over socket files (otherwise known as IPC connection https://nodejs.org/api/net.html#net_identifying_paths_for_ipc_connections), but it was pointed out to me that fetch needs to fully align on browser and server, which is the goal of node-fetch. However node-fetch will not implement IPC support.

It seems the only option, if lambda support is to be provided (and I for one am keen to push this through), is to fork node-fetch and add IPC support.

[...]

PR #1215, opened 2020-05-20
Add originAliases middleware option

benmccann:
Closes https://github.com/sveltejs/sapper/issues/489 and somewhat solves https://github.com/sveltejs/sapper/issues/356

Allows you to use sapper with an API service residing in another server. This is especially useful if your API server is written in another language.

This solution also allows you to use private internal hostnames when fetching data for better latency (since the request doesn't have to hit the externally-facing load balancer). It also works if fetching data from multiple domains (though I don't imagine that's the common case it seemed better not to prevent it)

I was questioning the best API for this. An alternate API might be two options: a function which rewrites URLs to fetch and a function that decides whether to include credentials. I'm open to other ideas as well and happy to implement anything that makes sense

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -11,7 +11,8 @@ import App from '@sapper/internal/App.svelte';
 
 export function get_page_handler(
 	manifest: Manifest,
-	session_getter: (req: Req, res: Res) => any
+	session_getter: (req: Req, res: Res) => any,
+	originAliases?: Map<String, String>
 ) {
 	const get_build_info = dev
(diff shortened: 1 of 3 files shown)


stalkerg:
First, I want to say a big thank you. Second, can you provide an example? If it final version I want to start fighting for that.
I will check it soon on my project. (maybe I can finally use sapper)

benmccann:
@stalkerg the PR has documentation that explains how to use it. You can find me on Discord with the same username if you need some additional help

hgl:
@benmccann Any reason this PR was closed?

stalkerg:
I have no such issue with SvelteKit anymore...

hgl:
I'll migrate to SvelteKit as soon as it's recommended for production use.

Issue #437, opened 2018-09-13
Provide an option to tell Sapper to serve its assets from a CDN


ispyinternet:
If we had an option to instruct sapper to serve the css and js from a different domain we could then offload those assets onto a CDN.

ispyinternet:
Could we use an environment variable to signal our intention to sapper, something like SAPPER_CDN_DOMAIN and (perhaps also SAPPER_CDN_BASE_PATH)?

The server generated assets would be easy for sapper to rewrite.

Template based assets, what could there be src and href (for example direct link to download asset), any others?

In both cases could we use a directive?

‘’’
<a cdn:href=‘/download.pdf’>download</a>
<img cdn:src=‘/logo.png’ />
‘’’

vladejs:
I have a similar situation. I want to load assets via CDN or local depending on `NODE_ENV` env variable, but don't know how to `if` that in the `template.html` file (I know how to put process.env.NODE_ENV inside via webpack though)

aubergene:
I like this idea. When you say assets are you exclusively talking about files within the `assets/` folder?

ispyinternet:
and the svelte generated javascript chunks, and any webpack generated css you extract too I guess.

ispyinternet:
@Rich-Harris  you probably know about this, but in case not: https://webpack.js.org/concepts/output/#advanced

Mention of CDN configuration. Not sure if this can be leveraged in anyway?

ispyinternet:
Wonder if this tool can be used?

https://www.npmjs.com/package/cdnizer

digitalfuel:
I also have a project that I would love to start with sapper but need generated chunks, css etc served from S3 CDN different to the apps domain.

waspeer:
+1

sw-yx:
this doesnt look like its sapper’s responsibility? you can configure webpack externals for this.

aubergene:
It would be needed or at least nice when the site is statically generated

ispyinternet:
as I see the problem, you want to dev to use local assets but you want production to use CDN assets and while it can be argued that the developer could use a bundler to replace the links to user-land assets, i'm not sure how easy / possible it would be to get a bundler to edit the code generated by sapper to 'CDN' the JS / CSS chunks?

Other options as suggested on discord just place a CDN in front of sapper and let it cache / passthrough accordingly

IlyaSemenov:
[...]

Issue #1216, opened 2020-05-21
Handling configuration in Sapper


benmccann:
## Background

There are a number of PRs or open issues that would require providing a configuration option. We should handle configuration in a well-thought out and consistent manner.

[language-tools already uses a `svelte.config.js`](https://github.com/sveltejs/language-tools/blob/3139ccd5b4306a914be4ba64b0c036fcd8888f78/packages/svelte-vscode/README.md#generic-setup) and [Svelte was considering adopting it more widely](https://github.com/sveltejs/svelte/issues/1101).

I'm assuming that we would use a `.config.js` because it's inline with the existing `svelte.config.js` and also because it's more flexible and allows providing additional types (e.g. `function`)

## RFC

Please see [the RFC](https://github.com/sveltejs/rfcs/pull/23) for a proposed implementation and list of Sapper PRs that might be affected

antony:
Dare I say this sounds a bit like an RFC?

benmccann:
RFC submitted!

benmccann:
SvelteKit has a `svelte.config.cjs` file to address this

PR #1217, opened 2020-05-21
Support custom assets prefix for webpack

YYCoder:
Relate to an old issue [437](https://github.com/sveltejs/sapper/issues/437).

Support use custom assets prefix to serve static files like js and css, but currently only for webpack users since I'm not familiar with rollup. Still figuring how to solve this using rollup.

Personally, I think it's an important feature because there is a lot of companies use a different domain to serve their static files in order to bypass the browser request restriction of the same domain and to reduce meaningless bandwidth consumption generated by cookie.

For me, I like sapper and want to use it for our company's project, but this problem is completely preventing me from using it. I can't find any workaround to avoid it. So, I would like to add this feature, hopefully, it could help others like me.

What I do is add a placeholder(%sapper.assetsPrefix%) to get_page_hanlder.ts where generate js/css and HTTP links header, then when user run build or dev, it will read publicPath from webpack.config.js and replace it. So, the finally generated server runtime(server.mjs) could have a user-specified assets prefix. Because it needs to run build for every different publicPath, I add three subfolders in assets-prefix-webpack folder to test these three different situations. Looks weird, I know 😓.

So, for webpack users, all need to do is add a publicPath in their webpack.config.js like below.

```
output: {
	...config.client.output(),
	publicPath: 'https://some.domain.com/client/'
},
```
Note that `client/` is not omitable.


benmccann:
> a lot of companies use a different domain to serve their static files in order to bypass the browser request restriction of the same domain and to reduce meaningless bandwidth consumption generated by cookie

I saw an interesting article the other day that it's actually worse performance: https://blog.theodo.com/2019/09/cookieless-domain-http2-world/

YYCoder:
> > a lot of companies use a different domain to serve their static files in order to bypass the browser request restriction of the same domain and to reduce meaningless bandwidth consumption generated by cookie
> 
> I saw an interesting article the other day that it's actually worse performance: https://blog.theodo.com/2019/09/cookieless-domain-http2-world/

Brilliant article! And thanks for your advice.

It's true that using several domains to serve assets just to save bandwidth caused by cookies might be negligible these days, so normally we limit the number of domains in case of it becomes a performance overhead. But I think, for sapper, it needs more fine-grained control for serve static files because there might be still a lot of users who need to use different domains.

benmccann:
Hmm, so in looking at this, it seems the configuration is being put in the webpack config and then it's being read from there in a few different places. That introduces lots of checks for `.bundler` throughout the code and I'm not sure that Sapper should have to be so aware of the bundler that you are using.

[We are thinking about introducing a Sapper configuration file](https://github.com/sveltejs/sapper/issues/1216). It might be better to put any configuration in that file and then the bundlers can utilize it if necessary. I think that might be more in line with what is currently being done elsewhere

YYCoder:
Totally agree, I think sapper configuration file is necessary, like other frameworks like next or nuxt. I' ll look into it.

benmccann:
Thanks for putting together this PR! Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit by adding a `svelte.config.cjs` file . I'm going to go ahead and close this PR since we won't be adding the feature to Sapper. You might like to check out SvelteKit instead. We'll be reviewing PRs for missing features there

Issue #760, opened 2019-06-20
TypeScript support updates (Sapper + Svelte 3)


pnmcosta:
Hi,

Just thought I'd share my experience with Sapper + TypeScript.

I have a monorepo with various "packages", and one of them is a Sapper app that I've mostly converted to TS with the help of some repos and issues from here. Thank you all for your efforts! I plan to share this monorepo boilerplate once it's more presentable than it is now :)

Only a few things stand out for me, that I'll summarize here for others who may want to adopt it too:

- the new `mjs` generated modules into `./src/node_modules/@sapper` are impossible to import with TS for the moment, see https://github.com/Microsoft/TypeScript/issues/18442, TLDR: As far as I can tell at the moment it is not certain how Node will proceed with `mjs` therefore TS is still unsure how to proceed too. Workrounds involve other dependencies that i'd like to avoid.
- sapper build/dev config expects entry points in `.js` so you have to keep `client.js`, `server.js` and `service-worker.js`, it would be nice if we could configure this, or sapper accept one or the other.
- you need to configure rollup/webback to include a `typescript` plugin, e.g. `rollup-plugin-typescript2`
- you can't use `<script lang="ts">` on `*.svelte` components, for the moment, mostly because of point 1 here.

So far so good, and I'm really enjoying the sapper/svelte world, IMO is much more readable/understandable overall than most other frameworks out there.

Thanks,

Teoxoy:
> the new mjs generated modules into ./src/node_modules/@sapper are impossible to import with TS for the moment, see microsoft/TypeScript#18442, TLDR: As far as I can tell at the moment it is not certain how Node will proceed with mjs therefore TS is still unsure how to proceed too. Workrounds involve other dependencies that i'd like to avoid.

You can import the `mjs` files by passing the `mjs` extension to the options of `rollup-plugin-node-resolve` (i.e. `resolve({ extensions: ['.mjs', '.js', '.ts', '.json'] })`).

I also have a file with the type info if you are interested:
```ts
declare module '@sapper/app' {
    // from sapper/runtime/src/app/types.ts
    // sapper doesn't export its types yet
    interface Redirect {
        statusCode: number
        location: string
    }
    // end

    function goto(href: string, opts = { replaceState: false }): Promise<unknown>
    function prefetch(href: string): Promise<{ redirect?: Redirect; data?: unknown }>
    function prefetchRoutes(pathnames: string[]): Promise<unknown>
[...]

PR #1222, opened 2020-05-22
TypeScript support: don't try to compile templates

benmccann:
Enables TypeScript support https://github.com/sveltejs/sapper/issues/760

TypeScript won't compile with Svelte compiler and first needs to be preprocessed. However, that happens after this check because `create_app` outputs JS which imports these `preload` exports before we try to bundle the whole app

This also has to be much faster to build the mainfest!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5260,6 +5260,11 @@
       "integrity": "sha1-IzTBjpx1n3vdVv3vfprj1YjmjtM=",
       "dev": true
     },
+    "strip-comments": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/strip-comments/-/strip-comments-2.0.1.tgz",
+      "integrity": "sha512-ZprKx+bBLXv067WTCALv8SSz5l2+XhpYCsVtSqlMnkAXMWDq+/ekVbl1ghqP9rUHTzv6sm/DwCOiYutU/yp1fw=="
+    },
     "strip-eof": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/strip-eof/-/strip-eof-1.0.0.tgz",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5260,11 +5260,6 @@
       "integrity": "sha1-IzTBjpx1n3vdVv3vfprj1YjmjtM=",
       "dev": true
     },
-    "strip-comments": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/strip-comments/-/strip-comments-2.0.1.tgz",
-      "integrity": "sha512-ZprKx+bBLXv067WTCALv8SSz5l2+XhpYCsVtSqlMnkAXMWDq+/ekVbl1ghqP9rUHTzv6sm/DwCOiYutU/yp1fw=="
-    },
     "strip-eof": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/strip-eof/-/strip-eof-1.0.0.tgz",
@@ -5443,6 +5438,11 @@
       "requires": {
         "setimmediate": "^1.0.4"
       }
+    },
+    "tippex": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/tippex/-/tippex-3.0.0.tgz",
+      "integrity": "sha1-sXYJonzq/+B5ezhzk/2Zh45Nfqk="
     },
     "tmp": {
       "version": "0.0.33",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "http-link-header": "^1.0.2",
     "shimport": "^1.0.1",
     "sourcemap-codec": "^1.4.6",
-    "string-hash": "^1.1.3"
+    "string-hash": "^1.1.3",
+    "strip-comments": "^2.0.1"
   },
   "devDependencies": {
     "@types/mocha": "^5.2.7",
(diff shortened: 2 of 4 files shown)


thgh:
Maybe add an extra check for the file extension and only disable for `.ts`?

There is a usecase where you may want to compose the preload function:
`export const preload = verifyUser(catchErrors(function () { ... }))`

Also reusable preload functions:
`export { preload } from './userData'`

benmccann:
Thanks @thgh !  I've updated the regex to cover both the cases you mentioned. Take another look and let me know what you think

> Maybe add an extra check for the file extension and only disable for .ts?

The code is typically in a `.svelte` file, so we couldn't do that unfortunately. But I think this will be a really nice speed up for building apps in JS anyway.

benmccann:
I've been using this in my fork for the past few weeks and the only issue I ran into was if `preload` was commented out. I've updated this PR to address that case and added several new unit tests to cover all the cases that have been suggested so far.

Conduitry:
Another possibility might be https://github.com/Rich-Harris/tippex

benmccann:
Cool. I'm happy to switch it out if you prefer. `strip-comments` has been working well though and is a heavily used library

Conduitry:
Well tippex seems to match the case we have here better because it also blanks out strings from its input, which strip-comments does not but which is needed to help avoid false positives.

benmccann:
Ah, yeah that make sense. I tried to use Tippex, but unfortunately it seems to have a bug which prevents it from being used here https://github.com/Rich-Harris/tippex/issues/23. So it's probably best to stick with `strip-comments` for now and if the bug is fixed we can switch out `strip-comments` for `tippex` then

benmccann:
Using Tippex is substantially more complicated. I can run `strip-comments` on the whole file. With Tippex, I need to first identify the location of each `script` block in the file and then extract its contents. Is there some existing piece of code in the Svelte compiler or preprocessor that can be reused for that? Or are we worried enough about encountering something like `let func: string = 'export async function preload()';` that it's worth it?

p.s. the test failures are unrelated to this PR and are coming from https://github.com/sveltejs/sapper/pull/743 having been merged

PatrickG:
In the meantime, maybe we can only compile the module script block, until this is resolved?

Like this:
```js
let matches;
if (/preload/.test(source) && (matches = source.match(/<script context="module">[^]*?<\/script>/))) {
  try {
    const { vars } = svelte.compile(matches[0], { generate: false });
    return vars.some((variable: any) => variable.module && variable.export_name === 'preload');
  } catch (err) {}
}
```

The benefit would be, to use typescript in the other script block.

benmccann:
I've gone ahead and switched this to use https://github.com/Rich-Harris/tippex

benmccann:
@adamb pointed out to me over Discord that there were some problems using `tippex`, so I've switched this back to using `strip-comments`. The issue with `tippex` is that it requires just the `script` contents to be passed in. However, it's basically impossible to get the `script` contents in a robust manner without writing a full-blown parser. E.g. consider the following:

```
<script context="module">
const x = `<script>...</script>`
export function preload() { }
</script>
```

We have a few options here:
- Sometimes hit a false positive on detecting `preload` with a simple regex. That's this PR
  - This seems acceptable to me because I think it's rare enough. The only case where I can realistically see it happening is if someone is writing about Sapper in `.svelte` files in Sapper. I checked the Sapper documentation and it would not be affected because it's primarily written in `.md`. It also wouldn't affect any users who are loading contents from any dynamic source like a database
- Sometimes hit a false negative on detecting `preload` by trying to extract the `script` contents and then use Tippex like in the above example
  - I think we could make the logic for extracting the script contents more robust so that failures were very unlikely, but we'd still have to accept that it would not always be perfect. I think that we could make those cases rare enough that it'd be a fine way forward though. E.g. we'd probably have to accept not handling this case:
```
<script context="module">
let x = y ? `<script context="module">var foobar="y";` : `<script context="module">var foobar="x";`;
x += '</script>';
export function preload() { }
</script>
```
- Try to first detect if the project is using TypeScript. Use the Svelte compiler to check if it's valid JavaScript just as we were doing before if the project is not a TypeScript project. Use the above logic only for TypeScript projects
  - Probably the best way to do this is just have the user specify if they are using TypeScript in the Sapper config file. The other ways I can think of also seem error-prone. I'd really like not to block on the config RFC though, so hope that we could either use one of the unperfect ways of detecting TypeScript in the meantime or leave this as an enhancement to implement later
- Don't support TypeScript
  - My least favorite option. TypeScript is hugely popular
- Massively refactor the entire Sapper project to be a series of Rollup plugins so that the preprocessor can executed before `create_manifest_data` and `create_app`
  - This would be so challenging that it's basically equivalent to the previous bullet. I'm not entirely sure it's even feasible, but if it were it would be a huge amount of work
  - I think it's somewhat unnecessary to be this robust because I can also come up with cases where the code currently checked into `master` does not work and it hasn't been an issue. I think it's acceptable to accept some edge case if it's rare enough. E.g. the following would break today:
```
<script context="module">
const x = `<style>`
export function preload() { }
</script>

Hello world
<!--
</style>
-->
```

benmccann:
The main thing I was keeping my eyes on with this PR is the fact that it uses a regex to extract the `<script context="module">` tags. It turns out that svelte itself uses a regex to extract `script` tags, so I think it's highly unlikely that this would cause any regression. If Svelte comes up with any other way of doing this in the future then we can always update Sapper to match

https://github.com/sveltejs/svelte/blob/40987b7780ef1865fb94e89b4d2f5a53a71056f9/src/compiler/preprocess/index.ts#L97

Conduitry:
This was a good idea, but closing in favor of #1344, which is a better idea.

PR #1225, opened 2020-05-24
Docs: improve docs for skipping server-side script execution

benmccann:
The old title "Making a component SSR compatible" didn't seem quite accurate to me. We're not making the component SSR compatible in the sense that it renders on the server. Rather we're just rendering the component on the client

This also cleans up some of the text and makes it clearer that you can use the same technique for any module you wish to render on the client-side only and not just ones that would fail on the server-side

### DIFF
--- a/site/content/docs/06-server-side-rendering.md
+++ b/site/content/docs/06-server-side-rendering.md
@@ -4,11 +4,11 @@ title: Server-side rendering
 
 Sapper, by default, renders server-side first (SSR), and then re-mounts any dynamic elements on the client. Svelte provides [excellent support for this](https://svelte.dev/docs#Server-side_component_API). This has benefits in performance and search engine indexing, among others, but comes with its own set of complexities.
 
-### Making a component SSR compatible
+### Skipping SSR for a component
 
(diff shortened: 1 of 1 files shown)


antony:
This section of the docs relates more to creating a component in your application which will render on both the client and server-side, but might only run it's javascript on the client side, so I believe the old wording is more accurate.

I think the what you have suggested here is a lot more wordy without saying anything more than it already does.

I see this part of the docs referred to a lot in chat, and it's never failed to help with the problem it was designed to, so I'm not in favour of this change.

pngwn:
The old title is also more appropriate because that is the symptom that people will notice, some kind of incompatibility with the server build. The new title forces people to understand the implementation details of both the client specific module they are using and Sapper’s SSR. This error is almost always due to the fact that the module references the window on import but that is very rarely advertised by the modules in question.

benmccann:
Ok. I'll close this for now

I think we could do a bit more to describe how hydration works. I've read both the Sapper and Svelte docs and it's not quite clear to me from the docs what happens. I actually now think that Svelte is recreating the entire DOM client-side, which is much different than what I expected and not described in the docs. I thought it'd just attach event listeners

PR #1227, opened 2020-05-24
site: Add documentation about rel=external

pgjones:
This is a very handy feature that took me an age to find (by reading
the code). This should make it more visible to others.

Ideally I'd like this feature to be documented, so feel free to drop or edit this if it doesn't fit the documentation as you'd like.

### DIFF
--- a/site/content/docs/08-link-options.md
+++ b/site/content/docs/08-link-options.md
@@ -22,7 +22,7 @@ We can mitigate that by *prefetching* the data. Adding a `rel=prefetch` attribut
 
 ### rel=external
 
-The sapper router listens for clicks on `<a>` elements and bypasses the normal browser navigation for relative (same origin) paths. We sometimes need links that are targeted at routes that are not part of the sapper app and hence should be reached by a normal browser navigation.
+By default, the Sapper runtime intercepts clicks on `<a>` elements and bypasses the normal browser navigation for relative (same-origin) URLs that match one of your page routes. We sometimes need to tell Sapper that certain links need to be be handled by normal browser navigation.
 
 Adding a `rel=external` attribute to a link...
 
--- a/site/content/docs/08-link-options.md
+++ b/site/content/docs/08-link-options.md
@@ -1,13 +1,12 @@
 ---
-title: Prefetching
+title: Link options
 ---
+
+### rel=prefetch
 
 Sapper uses code splitting to break your app into small chunks (one per route), ensuring fast startup times.
 
 For *dynamic* routes, such as our `src/routes/blog/[slug].svelte` example, that's not enough. In order to render the blog post, we need to fetch the data for it, and we can't do that until we know what `slug` is. In the worst case, that could cause lag as the browser waits for the data to come back from the server.
-
-
-### rel=prefetch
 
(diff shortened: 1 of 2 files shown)


site/content/docs/08-link-options.md
```diff
@@ -19,4 +18,16 @@ We can mitigate that by *prefetching* the data. Adding a `rel=prefetch` attribut
...
+<!-- TODO add a function to prefetch programmatically -->
+
+### rel=external
+
+The sapper router listens for clicks on `<a>` elements and bypasses the normal browser navigation for relative paths. We sometimes need links that are targetted at routes that are not part of the sapper app and hence should be reached by a normal browser navigation.
+
+Adding a `rel=external` attribute to a link...
+
+```html
+<a rel=prefetch href='path'>Path</a>
```
benmccann, line 30: Should this be `external` instead of `prefetch`?

pgjones: Yes, thank you.

site/content/docs/08-link-options.md
```diff
@@ -19,4 +18,16 @@ We can mitigate that by *prefetching* the data. Adding a `rel=prefetch` attribut
...
+
+The sapper router listens for clicks on `<a>` elements and bypasses the normal browser navigation for relative paths. We sometimes need links that are targetted at routes that are not part of the sapper app and hence should be reached by a normal browser navigation.
+
+Adding a `rel=external` attribute to a link...
+
+```html
+<a rel=external href='path'>Path</a>
+```
+
+...will trigger a browser navigation.
```
benmccann, line 33: How about making this: "will trigger a browser navigation when the link is clicked."

pgjones: Done, thanks.

benmccann:
Just thinking about this some more... It seems odd to me that you'd have to manually tag every external URL with this. I would think that we should make Sapper check if the hostname is different than the current URL

pgjones:
@benmccann I've updated to try make this clearer. This only applies for relative (same origin) paths, different hostnames already work as you'd expect.

site/content/docs/08-link-options.md
```diff
@@ -19,4 +18,16 @@ We can mitigate that by *prefetching* the data. Adding a `rel=prefetch` attribut
 
 > `rel=prefetch` is a Sapper idiom, not a standard attribute for `<a>` elements
 
-<!-- TODO add a function to prefetch programmatically -->
\ No newline at end of file
+<!-- TODO add a function to prefetch programmatically -->
+
+### rel=external
+
+The sapper router listens for clicks on `<a>` elements and bypasses the normal browser navigation for relative (same origin) paths. We sometimes need links that are targetted at routes that are not part of the sapper app and hence should be reached by a normal browser navigation.
```
benmccann, line 25: nit: I would use "targeted" instead of "targetted". I'm not sure the latter is necessarily incorrect, but it's very uncommon: https://english.stackexchange.com/questions/205815/is-targetted-a-standard-british-english-spelling

pgjones: Done, thanks.

pgjones:
Thanks.

PR #1237, opened 2020-05-26
ignore html files with missing base element

akvadrako:
I have some html files in my static folder. This caused `sapper export` to die because it expects all html files to have a `<base>` element and they do not. This is a simple fix.

### DIFF
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -171,6 +171,7 @@ async function _export({
 
 		let tasks = [];
 
+		handle_html:
 		if (range === 2) {
 			if (type === 'text/html') {
 				// parse link rel=preload headers and embed them in the HTML
@@ -187,6 +188,12 @@ async function _export({
 
 					const base_match = /<base ([\s\S]+?)>/m.exec(cleaned);
 					const base_href = base_match && get_href(base_match[1]);
+					if(base_href === null) {
+                                            oninfo({
+                                                    message: `ignoring html file with missing base element`
+                                            });
+                                            break handle_html
+                                        }
 					const base = resolve(url.href, base_href);
 
 					let match;
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -188,12 +188,12 @@ async function _export({
 
 					const base_match = /<base ([\s\S]+?)>/m.exec(cleaned);
 					const base_href = base_match && get_href(base_match[1]);
-					if(base_href === null) {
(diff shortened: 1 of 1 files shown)


benmccann:
This seems like a duplicate of https://github.com/sveltejs/sapper/pull/984

akvadrako:
I don't think so. This just fixes a bug with export, it doesn't change normal handling of the base element.

src/api/export.ts
```diff
@@ -187,6 +188,12 @@ async function _export({
 
 					const base_match = /<base ([\s\S]+?)>/m.exec(cleaned);
 					const base_href = base_match && get_href(base_match[1]);
+					if(base_href === null) {
+                                            oninfo({
```
benmccann, line 192: you should replace the spaces with tabs for consistency

src/api/export.ts
```diff
@@ -187,6 +188,12 @@ async function _export({
 
 					const base_match = /<base ([\s\S]+?)>/m.exec(cleaned);
 					const base_href = base_match && get_href(base_match[1]);
+					if(base_href === null) {
```
benmccann, line 191: you need a space after `if`. I have a PR pending that will setup `eslint`

akvadrako:
It does seem like an alternative solution of #1208 though. That one uses a fake base value to continue crawling. My solution just skips the file if no base is present, which I thought was safer.

akvadrako:
> oh, I see. thanks for clarifying. I didn't look too closely at first. wouldn't you want the file exported though? otherwise, why have it present in your static directory in the first place?

It is exported, it just skips the link crawling.

benmccann:
It sounds to me like [`base` will probably be made optional](https://github.com/sveltejs/sapper/issues/904). If that's the case and `base` is not present on any page where `type === 'text/html'` then I don't think it'd work very well to skip link crawling on those pages. On the one hand, I understand that this solution would help you immediately. But if we're going to have to come back and change this same section of code I'd rather come up with a long-term solution the first time around especially given that reviewer cycles are so scarce.

akvadrako:
In this case it would make most sense to use normal base handling like a browser. So on page /abc/def, base defaults to /abc. Sound good?

benmccann:
Yeah, that makes sense to me

akvadrako:
Anyway I think it's correct that I close this in favor of 1208 since that one has a test case too.

Issue #920, opened 2019-09-30
Dev server fails to reload with "address already in use :::3000"


jakubsacha:
**Describe the bug**
Dev server fails to reload with error: "address already in use :::3000"

**Logs**
```
src/components/molecules/QueryInputSearchSuggestions.svelte changed. rebuilding...
Attempt...
✔ server (1.8s)
✔ client (1.8s)
events.js:186
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE: address already in use :::3000
    at Server.setupListenHandle [as _listen2] (net.js:1298:14)
    at listenInCluster (net.js:1346:12)
    at Server.listen (net.js:1434:7)
    at Polka.listen (/Users/jsacha/src/feather/node_modules/polka/index.js:53:22)
    at eval (webpack:///./src/server.js?:23:3)
    at Module../src/server.js (/Users/jsacha/src/feather/__sapper__/dev/server/server.js:550:1)
    at __webpack_require__ (/Users/jsacha/src/feather/__sapper__/dev/server/server.js:27:30)
    at /Users/jsacha/src/feather/__sapper__/dev/server/server.js:118:18
    at Object.<anonymous> (/Users/jsacha/src/feather/__sapper__/dev/server/server.js:121:10)
    at Module._compile (internal/modules/cjs/loader.js:936:30)
Emitted 'error' event on Server instance at:
    at emitErrorNT (net.js:1325:8)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  code: 'EADDRINUSE',
  errno: 'EADDRINUSE',
  syscall: 'listen',
  address: '::',
  port: 3000
}
> Server crashed
```

**Configs**

```
const webpack = require("webpack");
const path = require("path");
const config = require("sapper/config/webpack.js");
const pkg = require("./package.json");

const MiniCssExtractPlugin = require("mini-css-extract-plugin");

const mode = process.env.NODE_ENV;
const dev = mode === "development";

const alias = { svelte: path.resolve("node_modules", "svelte") };
const extensions = [".mjs", ".js", ".json", ".svelte", ".html"];
const mainFields = ["svelte", "module", "browser", "main"];
const { sass } = require("svelte-preprocess-sass");
const BundleAnalyzerPlugin = require("webpack-bundle-analyzer")
  .BundleAnalyzerPlugin;

const purgecss = require("@fullhuman/postcss-purgecss")({
  // Specify the paths to all of the template files in your project
  content: [
[...]

PR #1239, opened 2020-05-27
Prevent server restarting while restart in progress

DveMac:
Resolves https://github.com/sveltejs/sapper/issues/920

The issue was caused by the server file watcher process calling `handle_result` in rapid succession. The source of this could be IDEs making many file saves (I only saw the issue in VSCode not vim), or just caused by people with an itchy cmd+S finger.

Initially considered debouncing the `handle_result` method, but a simple flag seems to be sufficient.

Flag was added inside the callback code for server changes and it prevents the server restart process if the flag is true.

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -310,14 +310,14 @@ class Watcher extends EventEmitter {
 					};
 
 					if (this.proc) {
-                        if (this.restarting) return;
-                        this.restarting = true;
+						if (this.restarting) return;
+						this.restarting = true;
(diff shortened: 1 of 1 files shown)


src/api/dev.ts
```diff
@@ -310,11 +310,14 @@ class Watcher extends EventEmitter {
 					};
 
 					if (this.proc) {
+                        if (this.restarting) return;
```
benmccann, line 313: these spaces should be tabs for consistency with the existing code

PR #1244, opened 2020-05-28
Prefix file paths in SW manifest with a slash

### LINKED ISSUE #1210: SW not serving precached files
**Describe the bug**
The serviceworker does not serve cached files because of a missing "/" in the filenames of the `cached`-variable

**Logs**
image

**To Reproduce**
Run `npx degit "sveltejs/sapper-template#rollup" my-app` and set a breakpoint in the serviceworker.

**Expected behavior**
The serviceworker should find the cached files and return them.

- Your browser and the version: Chromium Version 81.0.4044.138 (Developer Build) (64-bit)

- Your operating system: Voidlinux Kernel 5.4.40_1

- Sapper version 0.27.13

- Svelte version 3.22.3

- Whether your application uses Webpack or Rollup: Happens with both templates.

**Severity**
annoying

**Workaround**
My current workaround is to prepend a "/" for the filenames in the `cached` variable:
```javascript
// `shell` is an array of all the files generated by the bundler,
// `files` is an array of everything in the `static` directory
const to_cache = shell.concat(files);
//const cached = new Set(to_cache);
const cached = new Set(to_cache.map(filename => `/${filename}`));
```

kaitsh:
This PR adds a "/" to the file paths in the SW manifest file during compile time.
Fixes #1210


benmccann:
@antony should we apply the "breaking change" label to this one?

antony:
@benmccann why do you consider this breaking?

benmccann:
For folks that already updated their service worker to match the existing paths, this change would break their caching. I think it's rare that folks have fixed their apps because I see broken apps in the wild all the time, so this change would fix the majority of apps and probably only a small minority would need to revert their service workers to match this change. But it's worth calling out in the changelog as a breaking change so folks can check and be aware if they previously updated their service worker to deal with this issue that they'll need to update it again

lukeed:
Breaking something by fixing it is probably the best kind of break, haha. Would also mean any workaround for anything has to be considered before patching.

PR #1250, opened 2020-06-03
separate <Layout> from <Error> in generate_app( )

thebells1111:
_error.svelte will now be able to be styled independently of _layout.svelte to allow for unique error pages. This is in response to the comment by @intrikate in PR #948: _error.svelte should not inherit the common layout by default.

### Tests

### DIFF
--- a/src/core/create_app.ts
+++ b/src/core/create_app.ts
@@ -283,13 +283,13 @@ function generate_app(manifest_data: ManifestData, path_to_routes: string) {
 			setContext(CONTEXT_KEY, stores);
 		</script>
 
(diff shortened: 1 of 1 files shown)


Conduitry:
Thanks - but while we do want to have a way for pages to know whether they are an error page, I don't think we want to do it this way - which would be a breaking change in any case.

ArekBartnik:
@Conduitry it would be a breaking change but easily fixable (move some components from `_layout` to `_error`)

I think the reasoning should be this - if `_layout` encapsulates correct pages, `_error` should do the same for error pages.
Now there's an inconsistent pattern.

PR #1258, opened 2020-06-07
Add section with GitHub Actions

koddr:
- Add section with GitHub Actions to README
- Add link to Sapper Action & short example
- Add links to more complex examples

### Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,27 +72,6 @@ You should be good to test changes locally.
 npm run test
 ```
 
-## GitHub Action for Sapper
-
-Create a `.github/workflows/sapper_build.yml` file in your _Sapper-based_ website repository and use this GitHub action to build your static website with Sapper:
-
-- [actions/sapper-action](https://github.com/marketplace/actions/sapper-action)
-
-For example:
-
-```yml
-- name: Build Sapper
-  uses: truewebartisans/actions-sapper@master
-  with:
-    build_mode: "export"  # or "build"
-    args: "--entry about" # or nothing, by default "--legacy"
-```
-
-More complex examples:
-
-- [Build Sapper & Deploy to GitHub Pages](https://github.com/marketplace/actions/sapper-action#-deploy-to-github-pages)
-- [Build Sapper & Deploy to remote virtual server (VDS/VPS/Droplet) via SSH](https://github.com/marketplace/actions/sapper-action#-deploy-to-remote-virtual-server-vdsvpsdroplet-via-ssh)
-
 ## License
 
 [MIT](LICENSE)
--- a/README.md
+++ b/README.md
@@ -72,6 +72,27 @@ You should be good to test changes locally.
 npm run test
 ```
 
+## GitHub Action for Sapper
+
+Create a `.github/workflows/sapper_build.yml` file in your _Sapper-based_ website repository and use this GitHub action to build your static website with Sapper:
+
+- [actions/sapper-action](https://github.com/marketplace/actions/sapper-action)
+
+For example:
+
+```yml
+- name: Build Sapper
+  uses: truewebartisans/actions-sapper@master
+  with:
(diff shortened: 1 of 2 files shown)


antony:
Hi @koddr 

This is a nice idea - however there are a couple of issues with it:

1. It depends on a community maintained github action, which we don't want to end up endorsing if we have no control over it. I don't think it's strictly necessary - simply running export in the action will work, no action needed.
2. Readme isn't the correct place for this - it's not relevant for the majority of users. It would be better contributed as a deployment option in the documentation, where other methods such as Vercel are documented.

antony:
@koddr Thanks - I think the title should be something along the lines of `Deploying with github actions`, and the instructions shouldn't include the third party dependency.

What does the dependency actually do anyway? All you need to do is:

`sapper export`

what's the purpose of it?

koddr:
Hello! 👋 

> @koddr Thanks - I think the title should be something along the lines of `Deploying with github actions`, and the instructions shouldn't include the third party dependency.

Changed.

> What does the dependency actually do anyway? All you need to do is:
> 
> `sapper export`
> 
> what's the purpose of it?

Have you had any experience with [GitHub Actions](https://github.com/features/actions)? Because it's strange question for me, sorry.

This is not _third-party dependency_. It's a _step_ (or _container_ or _image_ or _pod_) to prepare a static website created on Sapper for further deployment to any platform via `GitHub Actions` CI/CD.

`Sapper Action` is a process of automating the deployment: you commit changes to your repository and GitHub (automatically) makes a `sapper export` for you (with the right flags) and deploy this build to `GitHub Pages` or `Netlify` or anywhere you want.

> For more understanding, if you know Eleventy ([11ty](https://www.11ty.dev/)) static website generator, they have community-driven GitHub action: [TartanLlama/actions-eleventy](https://github.com/TartanLlama/actions-eleventy), which saves my (and not so) personal time every project. So, `Sapper Action` it's exactly the same automation!

By the way, I am ready to give this GitHub action to you, if third-party developers doing something for `Sapper` – is a problem for you. I consider it my duty to support the project I love.

Thanks.

antony:
>> This is not third-party dependency. It's a step (or container or image or pod) to prepare a static website created on Sapper for further deployment to any platform via GitHub Actions CI/CD.

I see - I hadn't realised that you were an author. It's a third party dependency because it's not something the Svelte organisation controls, which means that if we were to include it in the official documentation, and you were then to change it to do something nefarious, then it would be bad for us - do you see?

Not to suggest that you would, but we have a certain responsibility to the community.

>> Have you had any experience with GitHub Actions? Because it's strange question for me, sorry.

Absolutely - I use them extensively. In my own projects, I have an action which runs:

```
  run: sapper export
```

and a series of pages are generated which can then be deployed. It looks like this is essentially the same thing that your action does.

In order to keep things simple for the documentation - we feel that it is better if the actual sapper commands that are run are included in the documentation, rather than hiding them behind a third-party action.

>> By the way, I am ready to give this GitHub action to you, if third-party developers doing something for Sapper – is a problem for you. I consider it my duty to support the project I love.

And this too is something we'd absolutely love - but we have to be careful because of two things - one, adopting a third-party project puts the burden of maintenance on to us, and two - because adding somebody new to a project under the svelte org is something that has to be done carefully, because we don't know you, yet.

So essentially, everything you say is valid and true, but ideally we keep this documentation section pure (and free of these unknowns), and then we add your project to the https://github.com/sveltejs/integrations project.

After a certain amount of time, we can decide that if enough people are using the integration, then we adopt it as official, add it to the organisation and the documentation, and invite you to become a collaborator to oversee maintenance and development of it.

But it has to happen in that order - to maintain the quality, and security, of the Svelte organisation. It's our duty to the community.

koddr:
Okay. Thanks for reply 👍 

I understand, because Svelte Community is more important, than my wish.

Issue #919, opened 2019-09-30
<link rel="preload"> is missing when exporting


khrome83:
**Describe the bug**
Issue #402 and PR #568 added `<link rel="preload">` to the output. Currently I am unable to see this anywhere. 

#402 was adding it to the `sapper export` command. But it is not showing on export, or when running develop, or using build/start...

It appears that the code in #568 is now failing. This does appear to be a bug that was not caught with regression. 

**Logs**
Please include browser console and server logs around the time this bug occurred.

**To Reproduce**
Run export on any site and notice that the expected `<link rel="preload">` is missing in `sapper 0.2.7.8`. 

You can view source, use inspector, or check Google page Speed Insights.

#### You can see this on a live site here - 
https://khromedev-72myozpsp.now.sh/

#### GitHub Repo - 
https://github.com/khrome83/khrome.dev/tree/svelte-poc/www

You can see issue brought up in Google Page Speed Insights - https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fkhromedev-72myozpsp.now.sh%2F

**Expected behavior**
I expect the following to be in the `<head>` based on #402 
```html
  <link rel="preload" href="styles.css" as="style">
  <link rel="preload" href="/client/chunk.3065f5a7.js" as="script">
  <link rel="preload" href=/client/index.eb6a9ec4.js" as="script">
```

**Information about your Sapper Installation:**
- Your browser and the version: Chrome 56

- Your operating system:  OS X 10

- Your hosting environment: Local, Now v2

- Sapper version 0.27.8

- Svelte version 3.9.1

- Exported App

- Using Rollup

**Severity**
Medium Severity. Can drastically reduce page load times.

andreinhugo:
@s0kil fix evaluates if link tag has a `rel="preload"` attribute but link tag gets created with `rel="modulepreload"`:

console.log(link):
`Link {
refs: [ { uri: '/client/client.f27e5f31.js', rel: 'modulepreload' } ]
}`

Sorry, can't test in a bigger project that requires a bundle.js file but I believe this file should be added to the document's head.

benmccann:
The issue is here: https://github.com/sveltejs/sapper/blob/cfa5a426be030d9dda5f9b89281c3e70ff43585f/src/api/export.ts#L230

We need to support both `preload` and `modulepreload`. Sapper typically uses `modulepreload`, so the code to support this isn't actually going to do anything today for the most part

[...]

PR #1259, opened 2020-06-09
Add modulepreload Header For JS Client

s0kil:
Fix https://github.com/sveltejs/sapper/issues/919 by supporting `modulepreload` header.


s0kil:
@Conduitry Anything else required to be merged?

PR #1262, opened 2020-06-11
chore: bump sirv version

lukeed:
### Summary

Updates `sirv` dependency to 1.0. There are no (relevant) breaking changes since Sapper's tests and site never made use of `single` (because it didn't exist outside of the `@next` prereleases).

### Tests

-  [x] Ran the tests tests with `npm test` or `yarn test`)

---

I don't know what's changed between the Actions run yesterday and today, but the `master` branch is failing for me locally. This branch (after latest commit) and master now fail in the same way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5745,6 +5745,12 @@
         "kleur": "^3.0.0"
       }
     },
+    "webpack-modules": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/webpack-modules/-/webpack-modules-1.0.0.tgz",
+      "integrity": "sha512-2SvHR8a9jigRPxh8SoiVG4D4HBU1gZ0N6s8dk9WuTiyVm+ASCkeZSnKEKC3W0CdZqwZchYhaJQqxAEl/oMAklA==",
+      "dev": true
+    },
     "webpack-sources": {
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/webpack-sources/-/webpack-sources-1.3.0.tgz",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2073,8 +2073,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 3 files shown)


lukeed:
Ok, sometimes fails...? lol

benmccann:
The `.mjs` thing has been causing issues for TypeScript and Snowpack as well. I was suggesting maybe we consider changing it to just be `.js` https://github.com/sveltejs/sapper/issues/1204#issuecomment-642146965. Glad to see you got it figured out for WebPack at least

Conduitry:
Can you update package-lock.json please?

lukeed:
Back to failing – as expected.

My guess is that one of the other lockfile updates pulled in a change that doesn't comply with the layout test expectant. Would explain why it also always fails for me locally (on master).

Conduitry:
> Would explain why it also always fails for me locally (on master)

Are you not installing with npm when you've been doing these local tests? The actual released builds use npm, and so you probably should be as well.

IMO this PR should only include package-lock.json updates related specifically to updating sirv anyway.

lukeed:
No, never.

Force-pushed the correct sequence of events. Should be working now.

This does highlight that there's something in the dependency chain that's not acting as expected (anymore). This is the catch-22 with relying on lock files. It works here, but the dependencies are defined as ranges. Users can be getting different versions which produces unexpected/faulty behavior, but we'd never see it because our development environment is locked in a glass jar, frozen at some point in time.

benmccann:
It seems that if you fully replace the package-lock with all updated dependencies that Pupeteer is now returning `'y: bar 1 z: baz 1' goto foo/bar/qux' goto foo/abc/def child segment: baz'` instead of `'y: bar 1\nz: baz 1'\ngoto foo/bar/qux'\ngoto foo/abc/def\nchild segment: baz'` when getting the `textContent` of a node (i.e. returning spaces instead of new lines). We will need to figure out how to upgrade it at some point, but I'm glad we have the lock-file in this case. It means the tests didn't start just randomly failing and we can lock it to a working version until we figure out what's wrong

lukeed:
Nice, thanks for tracking it down :)

benmccann:
It looks like the behavior change was introduced in Svelte 3.10.0. I'm not exactly sure which commit it was. I don't see anything obvious in the changelog. Anyway, it seems like we should just upgrade Svelte and fix the test

benmccann:
It took me quite awhile to figure out what was going on. I filed a separate issue to track and sent a PR to fix it though

lukeed:
Because new `sirv` uses `.mjs` file for the "module" entry. 
And even though webpack will resolve that entry, it's stubborn and refuses to treat `mjs` files correctly when coming from `.js` (and vice versa). Talk a little bit about it [on its readme](https://github.com/lukeed/webpack-modules#webpack-modules). Hopefully will be fixed in v5

benmccann:
I'll go ahead and merge this one since it only affects the tests and since the template PR was merged and this is the same exact change

PR #1265, opened 2020-06-12
Document routes.parts

benmccann:
I was always seeing `parts` as being `null` and couldn't figure out what it held. It probably took me a couple hours of backtracking through the code to see where it got populated. I figured I'd document it for the next person that looks at this code


benmccann:
Updated

Issue #1260, opened 2020-06-10
Reducing size of JS request chain


benmccann:
https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Frealworld.svelte.dev%2F

![Screenshot from 2020-06-09 20-28-09](https://user-images.githubusercontent.com/322311/84223884-f70b9c80-aa8f-11ea-8f35-1c783591c1df.png)

It looks to me like the request chain gets deeper each time a component imports another component. I wonder how scalable this is. On the one hand, it means downloads for subsequent page navigations will be smaller because those components may have been downloaded on previous pages. But on the other hand, I imagine you could easily end up with some pretty long download chains. E.g. `client.js` only has a few lines in it and does very little except request two other files, which adds latency for little benefit

PR #712, opened 2019-05-24
Preload js assets - fixes #408

greensk:
Alternative to #440, fixes #408.

Preloading scripts content is important to supply good Google site speed rate. The #440 PR looks outdated so I implemented an own solution.

Issue #449, opened 2018-09-24
[Rollup] Remove `client.js` from Link header


lukeed:
On the Rollup version _only_, the `bundle_info.assets.main` JS bundle should not be included in the `Link` response header. 

While that would technically be the correct action to take (and continues to be with the Webpack template), it results in a double-download of the main file on _any_ application produced with Rollup.

The reason is because the main JS file is passed thru Shimport in every scenario, and so the document cannot infer that the preloaded asset can be reused. This can be verified by removing the `Link` header and instead using `link rel=preload as=script` tags instead. In this case, there's a JS console warning about the `client.*.js` not being used within a few seconds, despite the fact that _was_... it was just that a second request was made instead.

Contrast this with the Webpack template, which _does_ have a `script src=*` tag directly on the page. Here, the browser can correctly infer that the preloaded asset goes `<here>`.

The easiest solution is to remove the `client.*.js` file from the `preloaded_chunks` array, and do a "raw" request or the file thru Shimport. Since the page is SSR'd & isn't interactive until all chunks have finished loading anyway (thru Shimport), this wouldn't delay the page's timeline by much.

^ I have a local fix for this & can push it up if desired.

The other option is to do the browser-compat checks _in the server middleware_ instead of waiting to begin on `DOMContentLoaded`. You'd sniff the UA string that comes in & serve the appropriate bundle-set immediately.

This is something I've seen @kristoferbaxter do well with his `preact-hn` and is something I've considered working into `sirv` – or perhaps another module that builds from it. 

And finally, here's a network timeline for the base Rollup template, running on a simulated 3G.

<img width="1034" alt="screen shot 2018-09-24 at 2 44 07 pm" src="https://user-images.githubusercontent.com/5855893/45981610-821dd800-c00a-11e8-887e-98c6829e5d5e.png">

arxpoetica:
@lukeed what's your local fix? mind sharing?

mrkishi:
This seems to happen regardless of whether `shimport` is being used or not.

I don't know what the best course of action is, but the browser (Chrome, at least) won't match the preloaded scripts with scripts of `type="module"`. When using `shimport`, it won't match them because they should be `fetch` instead of `script`.

More info [in this article](https://developers.google.com/web/updates/2017/12/modulepreload), with an even more experimental feature `modulepreload`. I guess it would be safe to swap the current `preload` for a `modulepreload`, since the latter will only work on browsers that also do not load `shimport`.

[...]

PR #440, opened 2018-09-13
Preload assets in Rollup apps - fixes #408

nsivertsen:
I'm not sure if the way I'm testing whether a chunk is a page is 100% right, see `RollupResult.ts`, lines 47-48.

Fixes #408

PR #1269, opened 2020-06-13
Preload CSS and JS dependencies

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1260
Closes https://github.com/sveltejs/sapper/issues/408
Closes https://github.com/sveltejs/sapper/pull/440
Closes https://github.com/sveltejs/sapper/pull/712
Closes https://github.com/sveltejs/sapper/issues/449
Closes https://github.com/sveltejs/sapper/issues/405
Closes https://github.com/sveltejs/sapper/issues/70

I made a small library to extract the dependency tree from the Rollup `OutputChunk[]`. If you're okay with it being added a dependency here then I can go ahead and publish it to `npm`. By having it as an external project other projects can use it as well. Also, it made it easier to test because I wrote a test that operates on a JSON dump of a Rollup compilation result. Importing JSON in a test was kind of tricky and I had more control over changing settings for tsconfig, mocha, etc. in an external project to make that work

I ran this with `npm link` in [the realworld project](https://github.com/sveltejs/realworld) with Chrome set to emulate a slow 3g connection and it makes a big difference. You can see the JS files for the current page being loaded in parallel in the network tab instead of sequentially.

**This PR**

Screenshot from 2020-06-12 22-12-50

**Current Deployed Real World Sample**

Screenshot from 2020-06-12 22-20-50

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "puppeteer": "^5.0.0",
     "require-relative": "^0.8.7",
     "rollup": "^2.21.0",
+    "rollup-dependency-tree": "^0.0.1",
     "rollup-plugin-svelte": "^5.1.0",
     "sade": "^1.6.1",
     "sirv": "^1.0.0",
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -50,42 +50,54 @@ export function get_page_handler(
 			bundler: 'rollup' | 'webpack',
 			shimport: string | null,
 			assets: Record<string, string | string[]>,
+			dependencies: Record<string, string[]>,
+			css: {
+				main: string | null,
+				chunks: Record<string, string[]>
+			},
 			legacy_assets?: Record<string, string>
-		 } = get_build_info();
+		} = get_build_info();
 
 		res.setHeader('Content-Type', 'text/html');
 
 		// preload main.js and current route
-		// TODO detect other stuff we can preload? images, CSS, fonts?
-		let preloaded_chunks = Array.isArray(build_info.assets.main) ? build_info.assets.main : [build_info.assets.main];
+		// TODO detect other stuff we can preload like fonts?
+		let preload_files = Array.isArray(build_info.assets.main) ? build_info.assets.main : [build_info.assets.main];
 		if (!error && !is_service_worker_index) {
 			page.parts.forEach(part => {
 				if (!part) return;
 
 				// using concat because it could be a string or an array. thanks webpack!
-				preloaded_chunks = preloaded_chunks.concat(build_info.assets[part.name]);
+				preload_files = preload_files.concat(build_info.assets[part.name]);
 			});
 		}
 
+		let es6_preload = false;
 		if (build_info.bundler === 'rollup') {
-			// TODO add dependencies and CSS
-			const link = preloaded_chunks
-				.filter(file => file && !file.match(/\.map$/))
-				.map(file => `<${req.baseUrl}/client/${file}>;rel="modulepreload"`)
-				.join(', ');
 
-			res.setHeader('Link', link);
-		} else {
-			const link = preloaded_chunks
-				.filter(file => file && !file.match(/\.map$/))
-				.map((file) => {
-					const as = /\.css$/.test(file) ? 'style' : 'script';
-					return `<${req.baseUrl}/client/${file}>;rel="preload";as="${as}"`;
-				})
-				.join(', ');
+			es6_preload = true;
 
-			res.setHeader('Link', link);
+			const route = page.parts[page.parts.length - 1].file;
+
+			// JS
+			preload_files = preload_files.concat(build_info.dependencies[route]);
+
+			// CSS
+			preload_files = preload_files.concat(build_info.css.main);
+			preload_files = preload_files.concat(build_info.css.chunks[route]);
 		}
+
+		const link = preload_files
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -56,6 +57,7 @@
     "sucrase": "^3.10.1",
     "svelte": "^3.6.9",
     "svelte-loader": "^2.13.6",
+    "typescript": "^3.9.5",
```
antony, line 60: why does typescript get installed here?

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -69,9 +70,18 @@ export function get_page_handler(
...
-				.filter(file => file && !file.match(/\.map$/))
+
+			// TODO add CSS
+			let jsFiles = preloaded_chunks.filter(file => file && !file.match(/\.map$/));
+			const route = page.parts[page.parts.length - 1].file;
+			const deps = build_info.dependencies[route];
+			if (deps) {
+				jsFiles = jsFiles.concat(deps);
+			} else {
+				console.warn('Could not find dependencies for ' + route);
```
antony, line 81: console.* methods support multiple arguments which are concatenated into a single string.

```
console.warn('Could not find dependencies for', route);
```

package.json
```diff
@@ -66,7 +68,7 @@
     "test": "mocha --opts mocha.opts",
     "pretest": "npm run build",
     "build": "rm -rf dist && rollup -c",
-    "prepare": "npm run build",
+    "prepare": "cd ./node_modules/rollup-dependency-tree && npm install && npm run build",
```
antony, line 71: I'm assuming this is because it's an unpublished package

benmccann: yeah, I'll remove after publishing the package

package.json
benmccann: I've removed it. I think it was left over from an earlier iteration of working on this. I'm kind of surprised it's not there though because I think it'd be needed to specify which version is being used

runtime/src/server/middleware/get_page_handler.ts
benmccann: done

package.json
```diff
@@ -45,6 +45,7 @@
     "puppeteer": "^1.19.0",
     "require-relative": "^0.8.7",
     "rollup": "^1.17.0",
+    "rollup-dependency-tree": "https://github.com/benmccann/rollup-dependency-tree.git",
```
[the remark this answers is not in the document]
benmccann: Yeah, I'll go ahead and publish if we're okay with it. Should I check with @Conduitry too or would your review be enough?

I'm not sure about adding `plugin` to the name because it's not actually a Rollup plugin in the sense that it's added to the `plugins` in `rollup.config.js`. It's really just a library

habibrosyad:
Hoping that this will get merged soon 👍

package.json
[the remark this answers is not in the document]
benmccann: published to npm and updated!

bgjehu:
@benmccann does this preloading dependencies feature works for `sapper export`?

benmccann:
If you run the Sapper server, then `preload` will be added as a header. Since you can't easily control headers on exported pages, I believe that there's some code in `export` that tries to turn the headers into `<link rel="preload">` tags in the exported pages. I haven't tested the export functionality. That was around before this PR which wouldn't affect it and I haven't used `export` myself, but yes it should work - just a little differently

riccardolardi:
Also interested in whether the preloading will work in `sapper export`. @benmccann any hint on this?

benmccann:
I answered the question with all information I have about it just above: https://github.com/sveltejs/sapper/pull/1269#issuecomment-675727545

benmccann:
I think I know why it's not working for `export`: https://github.com/sveltejs/sapper/issues/919#issuecomment-693609251

I don't use `export`, so I'm not going to have time to fix it myself, but I'll review a PR if anyone wants to send one

riccardolardi:
@benmccann is it just about adding that line eg changing it to modulepreload? I‘d love to contribute if I‘m able

benmccann:
You'd want to support both, so I'd imagine the fix looks like this:
```
				if (ref.rel === 'preload' || ref.rel === 'modulepreload') {
					body = (body as string).replace('</head>',
						`<link rel="${ref.rel}" as=${JSON.stringify(ref.as)} href=${JSON.stringify(ref.uri)}></head>`);
				}
```

The main thing here is testing out the fix and making sure it works

cortopy:
just to give a bit of context about metrics, I run a quick lighthouse audit on [routify.dev](https://routify.dev/). It scores rather pourly on Performance (38 out of 100). I think the biggest issue is not having critical css, but that's another issue.

In this screenshot I'm pointing at the resources which could be loaded in a more performant manner using 'preload'

image

This is the [link](https://web.dev/uses-rel-preload/?utm_source=lighthouse&utm_medium=devtools) about how lighthouse comes to this conclussion

benmccann:
I'm not sure what you're suggesting. routify.dev is built with Routify and not Sapper, so I'm not really sure how it's related. For comparison, Sapper sites score very highly on Google's Page Speed score. E.g. https://realworld.svelte.dev/ which is built with Sapper scores in the 90s.

cortopy:
@benmccann sorry, I posted in the wrong thread. I moved the comment but notifications must've gone through anyway. My apologies. You should have the other one too I believe

riccardolardi:
@benmccann I've submitted a PR https://github.com/sveltejs/sapper/pull/1539 when you find some time I'd be thankful if you could review. Thanks!

PR #1275, opened 2020-06-15
Add option to disable SSR

benmccann:
Implements https://github.com/sveltejs/sapper/issues/383

Allows `ssrEnabled` to be a `function` in order to support [dynamic rendering](https://developers.google.com/search/docs/guides/dynamic-rendering) and per-route SSR

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -9,9 +9,29 @@ import { Manifest, Page, Req, Res } from './types';
 import { build_dir, dev, src_dir } from '@sapper/internal/manifest-server';
 import App from '@sapper/internal/App.svelte';
 
+/**
+ * A general purpose options-handling function based off Chart.js options handling
+ * @param {Array} inputs - An array of values, falling back to the last value.
+ * @param {object} [context] - If defined and the current value is a function, the value
+ * is called with `context` as first argument.
+ */
+function resolveOption(inputs: Array, context?: object) {
+	for (let i = 0; i < inputs.length; ++i) {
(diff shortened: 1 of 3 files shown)


Conduitry:
afaict, disabling SSR is only particularly interesting if it also plays nicely with exporting. More flexibility in making a pure SPA is the context I've generally seen this come up in. Have you tried to see how that goes?

benmccann:
I just tested exporting the real world app locally with this flag on and all the public pages on the exported site worked (I obviously can't register a user or use parts behind a login wall on an exported site)

I do think disabling SSR is very useful in other scenarios as well though such as testing, reducing server load, or improving time-to-interactive. I will be using it on my own site for improved performance in a non-exported scenario.

pjebs:
Would I be able to use a go backend with this PR (

antony:
@pjebs you can use a go backend with or without this PR. Come and [chat](https://svelte.dev/chat) to us if you want to understand more.

pjebs:
I meant Go backend and no node at all

benmccann:
I expect you would be able to do that today as well though I haven't tried. You would be missing some features that I think are valuable like support for asset `preload`, but wouldn't prevent you from doing it. But I do think that discussion is largely orthogonal from this one and would be better had on Discord. I don't want to go too far off-topic here

MichielLeyman:
What's required for this to be merged? Can I help?

riccardolardi:
I'm too looking forward for this merge and happy to help if needed.

lukechu10:
What's the status of this?

benmccann:
We won't be adding it to Sapper, but will add it to SvelteKit instead

PR #1276, opened 2020-06-15
Use same serve settings in dev and prod

benmccann:
It's unnecessary to disable the cache in development mode because all files under `client/` contain hashes and so will not be cached if their contents change

Removing this has two benefits:
* Faster reloads in development
* Makes dev more like prod


Conduitry:
It's probably indeed safe to cache the files under `/client/` but I don't think it is to cache the other ones this serves.

benmccann:
Oops. yeah I didn't realize it was serving multiple directories. https://github.com/sveltejs/sapper/pull/1277 is a better alternative

PR #1277, opened 2020-06-16
Cache client files in dev mode too

benmccann:
It's unnecessary to disable the cache in development mode because all files under `client/` contain hashes and so will not be cached if their contents change

Removing this has two benefits:
* Faster reloads in development
* Makes dev more like prod

### DIFF
--- a/runtime/src/server/middleware/index.ts
+++ b/runtime/src/server/middleware/index.ts
@@ -58,7 +58,7 @@ export default function middleware(opts: {
 
 		serve({
 			prefix: '/client/',
-			cache_control: dev ? 'no-cache' : 'max-age=31536000, immutable'
+			cache_control: 'max-age=31536000, immutable'
 		}),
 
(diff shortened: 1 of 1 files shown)


antony:
I can see the benefit and reasoning for this, but anything to do with caching and local just sounds like it will drive me nuts. Even if it has almost no impact.

What I would like to see is maybe this merged with the ideas in https://github.com/sveltejs/sapper/pull/1173 to provide more control over caching in Sapper. That way people could disabled/enable local caching if they wanted to, but also have more control over caching generally.

benmccann:
If it causes issues with caching locally then it will in prod as well and I think it'd be better to discover that locally. That being said, this isn't very important to me, so feel free to close if you're still nervous about it

Conduitry:
I've definitely had a few occasions when debugging locally something where I manually edited files in `/client/`. That sort of thing would be harder with this. I agree with Antony, and don't think doing caching during local dev is a good idea.

Issue #1279, opened 2020-06-17
Dynamic blog entry 404 on build


ldelossa:
**Describe the bug**
I am using this script:
https://github.com/Conduitry/sapper-gh-pages

and generating a static blog. You can see my code here:
https://github.com/ldelossa/ldelossa-blog

Almost everything works, but the static generation of the dynamic [slug].svelte blog entry 404s.
> Built in 3.2s
> Crawling http://localhost:3000/ldelossa-blog/
   6.51 kB   index.html
   7.09 kB   resume/index.html
   8.29 kB   projects/index.html
   8.41 kB   about/index.html
     220 B   blog.json
   6.83 kB   blog/index.html
       9 B   (404) blog/94297015-acc4-4089-b438-eeb566ce9c17
   5.71 kB   service-worker-index.html

You can see there by going to:
https://ldelossa.github.io/ldelossa-blog/blog 
and attempting to click on the first blog entry.

**Logs**
```
❯ npm run deploy

> TODO@0.0.1 deploy /home/louis/git/javascript/ldelossa-blog
> ./scripts/deploy.sh

rm '.github/workflows/sapper_deploy.yml'
rm '.gitignore'
rm 'README.md'
rm 'cypress.json'
rm 'cypress/fixtures/example.json'
rm 'cypress/integration/spec.js'
rm 'cypress/plugins/index.js'
rm 'cypress/support/commands.js'
rm 'cypress/support/index.js'
rm 'package-lock.json'
rm 'package.json'
rm 'scripts/deploy.js'
rm 'scripts/deploy.sh'
rm 'src/client.js'
rm 'src/components/Nav.svelte'
rm 'src/components/card/Card.svelte'
rm 'src/components/card/ProfileCard.svelte'
rm 'src/components/card/ProjectDeck.svelte'
rm 'src/components/sidebar/Console.svelte'
rm 'src/components/sidebar/Icons.svelte'
rm 'src/components/sidebar/Sidebar.svelte'
rm 'src/components/sidebar/SidebarItem.svelte'
rm 'src/routes/_error.svelte'
rm 'src/routes/_layout.svelte'
rm 'src/routes/about.svelte'
rm 'src/routes/blog/[slug].json.js'
rm 'src/routes/blog/[slug].svelte'
rm 'src/routes/blog/index.json.js'
rm 'src/routes/blog/index.svelte'
rm 'src/routes/blog/posts/_home-network-pt1.md'
rm 'src/routes/index.svelte'
[...]

PR #1280, opened 2020-06-17
Handle shell escaping

ldelossa:
this commit will handle file names with '$' characters correctly
when using the deploy script

Signed-off-by: louis <louis@localhost.localdomain>

actually closes: https://github.com/sveltejs/sapper/issues/1279

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -401,7 +401,7 @@ Also:
 
 ## 0.15.0
 
-* Nested routes (consult [migration guide](https://sapper.svelte.technology/guide#0-14-to-0-15) and docs on [layouts](https://sapper.svelte.technology/guide#layouts)) ([#262](https://github.com/sveltejs/sapper/issues/262))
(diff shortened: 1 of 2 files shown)


ldelossa:
@nuno1982 can this get merged?

benmccann:
It looks like Rich merged a commit https://github.com/sveltejs/sapper/commit/f3e9fc48d281ff990458a4a537a50d59db105e37 touching the same section of code, so this will need to be rebased

benmccann:
@ldelossa just a quick reminder that this PR will need to be rebased. do you still want to pursue it?

ldelossa:
Whoops sorry about that. AFK till tomorrow but will pick it up first thing then.

Conduitry:
There was the problem with the `find` command on macOS that has been since fixed, but I'm still not entirely clear on what sorts of filenames it was that were causing a problem and would have been fixed by this PR.

ldelossa:
It seems sapper will generate files with a "$" symbol in them. The way the script was written today they are not probably shell escaped, and the script expands them to empty variables.

benmccann:
I'm going to go ahead and close this since the Sapper repo will eventually be deprecated in favor of SvelteKit, so it doesn't matter much whether we update this deployment script or not

Issue #1284, opened 2020-06-19
Upgrade Svelte in tests


benmccann:
`test/apps/layout/test.ts` is failing on line 33 with newer versions of Svelte as pointed out by @lukeed in https://github.com/sveltejs/sapper/pull/1262

Starting with Svelte 3.10.0, it looks like the server is returning `\n\n\n\t\n\t\ty: bar 1\n\t\n\t\t\t\tz: baz 1\n\tgoto foo/bar/qux\n\tgoto foo/abc/def\n\t\t\t\n\n\tchild segment: baz\n\t` but then the client-side rendering is returning `\n\t\t\t\n\n\t\n\ty: bar 1 z: baz 1 goto foo/bar/qux goto foo/abc/def child segment: baz`

Previously it was returning `'\n\t\t\t\n\n\t\n\ty: bar 1\nz: baz 1\ngoto foo/bar/qux\ngoto foo/abc/def\n\nchild segment: baz'`

PR #1285, opened 2020-06-19
Upgrade Svelte in tests

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1284

I'm not quite sure why Svelte is behaving differently starting with 3.10.0, but I'm not sure it matters. Both behaviors are correct and plenty of people are using the newer versions without issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "sade": "^1.6.1",
     "sirv": "^0.4.2",
     "sucrase": "^3.10.1",
-    "svelte": "^3.23.2",
+    "svelte": "^3.24.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "sade": "^1.6.1",
     "sirv": "^0.4.2",
     "sucrase": "^3.10.1",
-    "svelte": "^3.6.9",
+    "svelte": "^3.23.2",
```
antony, line 57:
```suggestion
    "svelte": "^3.24.0",
```

PR #1303, opened 2020-07-04
Add --crawlstaticroutes flag to sapper export to add static routes as entrypoints to export

### LINKED ISSUE #1014: Export all pages, not just those linked to with <a> tags
A clear and concise description of what the problem is. For example: I'm always frustrated when [...]
When I export, I get the main index page and all the other pages linked to with <a> tags. This is explained in the [docs](https://sapper.svelte.dev/docs#How_it_works). However, my site has multiple pages that aren't directly linked to: 404 pages, pages shown after form submission etc. and these do not get exported. This means that parts of my site do not work when I host the static site on Netlify.

I'd like the 'npm run export' call to export all the files possible, not just those that are crawl-able from the main entry point. Ideally this should be inferred automatically, but I'd be willing to provide a sitemap file or equivalent to specify.

Quite important. This is blocking a straight-forward process of serving this from Netlify (both Sapper and Netlify have been excellent otherwise). I'm now looking to whether I should change my host and serve a node app, or whether there's an alternative way to produce the static files I need.

logannc:
Fixes #1014 when limited to static (i.e., non-dynamic) routes.

This add `--crawlstaticroutes` to `sapper export` so that static routes in `src/routes/` are added as entrypoints for the export process. This is

1. what people expect this feature to do (i.e., less surprising)
2. a major ergonomic benefit for those that have static routes not reachable through an SSR crawl

The current implementation is backwards compatible, though I do believe per (1) that in a future major version update this would be a good candidate for a breaking change to make this flag default true.

### DIFF
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -12,6 +12,7 @@ import Deferred from './utils/Deferred';
 import { noop } from './utils/noop';
 import { parse as parseLinkHeader } from 'http-link-header';
 import { rimraf, copy, mkdirp } from './utils/fs_utils';
+import { create_manifest_data } from '../core';
 
 const writeFile = promisify(fs.writeFile)
 
@@ -26,7 +27,10 @@ type Opts = {
 	concurrent?: number,
 	oninfo?: ({ message }: { message: string }) => void;
 	onfile?: ({ file, size, status }: { file: string, size: number, status: number }) => void;
+	routes?: string;
 	entry?: string;
+	ext?: string;
+	crawlstaticroutes: boolean;
 };
 
 type Ref = {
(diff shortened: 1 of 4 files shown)


logannc:
There was some pushback in #1014 but I still believe this has value. If this seems like something the maintainers will merge, i will add a test suite for this flag.

pngwn:
Two maintainers raised concerns about this and the issue was closed. We don't think the feature is particularly necessary given there is already a way to do this (and we don't like having multiple ways of doing the same thing), but we definitely aren't going to be adding another API to achieve something you can already do.

logannc:
@Rich-Harris @Conduitry can I get a second opinion?

benmccann:
You might prefer to use http://junglejs.org

Issue #843, opened 2019-08-12
Preload re-run upon hydration if server preload returned falsy


Conduitry:
**Describe the bug**
If the server preload returns something falsy, the client-side preload is run upon hydration.

**Expected behavior**
If the server preload returns something serializable (even if it's falsy), the client-side preload should not be run upon hydration.

I imagine this could be solved by doing a `|| {}` thing earlier in the process, so that the client actually gets a serialized `{}` in the initial rendered page, and knows not to run preload.

**Severity**
Probably low. If a preload is returning no data, it's probably not making any wasteful repeated ajax requests on the client.

Conduitry:
To document what happened with this this morning: Here's what I arrived at to not re-run client-side preload when server-side preload returned falsy:

```diff
diff --git a/runtime/src/server/middleware/get_page_handler.ts b/runtime/src/server/middleware/get_page_handler.ts
index 53f4136..cb94fac 100644
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -148,33 +148,33 @@ export function get_page_handler(
 
 		try {
 			const root_preloaded = manifest.root_preload
-				? manifest.root_preload.call(preload_context, {
+				&& (await manifest.root_preload.call(preload_context, {
 					host: req.headers.host,
 					path: req.path,
 					query: req.query,
 					params: {}
-				}, session)
-				: {};
+				}, session))
+				|| {};
 
 			match = error ? null : page.pattern.exec(req.path);
 
 
 			let toPreload = [root_preloaded];
 			if (!is_service_worker_index) {
-				toPreload = toPreload.concat(page.parts.map(part => {
+				toPreload = toPreload.concat(page.parts.map(async part => {
 					if (!part) return null;
 
 					// the deepest level is used below, to initialise the store
 					params = part.params ? part.params(match) : {};
 
 					return part.preload
-						? part.preload.call(preload_context, {
+						&& (await part.preload.call(preload_context, {
 							host: req.headers.host,
 							path: req.path,
 							query: req.query,
 							params
-						}, session)
-						: {};
+						}, session))
+						|| {};
 				}))
 			}
 
@@ -251,7 +251,7 @@ export function get_page_handler(
 
 					props[`level${l++}`] = {
 						component: part.component,
-						props: preloaded[i + 1] || {},
+						props: preloaded[i + 1],
 						segment: segments[i]
 					};
 				}
```

[...]

Issue #710, opened 2019-05-23
Exception in preload leads to flash of content


IlyaSemenov:
### Steps to reproduce

```
npx degit sveltejs/sapper-template#webpack my-app
```

create `src/routes/test.svelte`:

```html
<script context="module">
export async function preload(page, session) {
	await this.fetch('http://nosuchhost')
}
</script>

<script>
export let user
</script>

<div>user = {user}</div>
```

### Actual result

Server-side generated HTML contains error message:

```html
<h1 class="svelte-8od9u6">500</h1>

	<p class="svelte-8od9u6">request to http://nosuchhost/ failed, reason: getaddrinfo ENOTFOUND nosuchhost nosuchhost:80</p>

	<pre>FetchError: request to http://nosuchhost/ failed, reason: getaddrinfo ENOTFOUND nosuchhost nosuchhost:80
    at ClientRequest.eval (webpack:///./src/node_modules/@sapper/server.mjs?:1955:11)
    at ClientRequest.emit (events.js:182:13)
    at Socket.socketErrorListener (_http_client.js:391:9)
    at Socket.emit (events.js:182:13)
    at emitErrorNT (internal/streams/destroy.js:82:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:50:3)
    at process._tickCallback (internal/process/next_tick.js:63:19)</pre>
```

It is displayed for a moment in the browser, but then replaced with `user = undefined`.

In javascript console there is a message: `<Test> was created without expected prop 'user'`

### Expected result

The error message 500 stays there. No errors in javascript console.

nicobrinkkemper:
I'm still having this issue with Sapper

bzin:
I just bump into this issue while fetching data from a GraphQL api endpoint. 

It only occurs when i open the route the first time, refreshing the page and  data is shown just fine. 

```
<script context="module">
    export async function preload (page, session) {
        // Fetch GraphQL query with homepage data
        const queryHomepage = await require('./../queries/homepage.graphql');

        // Url from GraphQL endpoint
        const url = process.env.GRAPHQL_API
        const opts = {
            method: 'POST',
            headers: { 'Content-Type': 'application/graphql' },
            body: queryHomepage
        };
        
        try {
            const res = await this.fetch(url, opts);
            const data = await res.json();
            return { post: data };
        } catch(error) {
            console.log(error)
        }
    }
</script>
```

Banjerr:
[...]

Issue #447, opened 2018-09-23
Provide warning if store data cannot be serialized by devalue 


ispyinternet:
When you set store data on the server, if it cannot be serialized by devalue (objects containing functions are one such case), your client side will not receive a copy of the store in the init runtime. There are no warnings and for newbies this could be a source of confusion. Suggest a warning message to console server side and possible even set an error message in the store to warn the client side?

ansarizafar:
We should be able to pass a context object (like Graphql) to page/server routes and components both on server and client. The context object should be available as this.context like this.store. In my opinion its a must have feature.

neoel:
Here :)

https://github.com/sveltejs/sapper/blob/cca417a85a0515e052e035c6762d4671c2b034fe/templates/src/server/middleware/get_page_handler.ts#L320-L326

PatrickG:
I just stumbled across this issue.
This is fixed.

PR #1304, opened 2020-07-05
better error: show a warning/error when server-preloaded data can't be transmitted to the client

babichjacob:
- [x] *It's really useful if your PR relates to an outstanding issue, so please reference it in your PR, or create an explanatory one for discussion. In many cases features are absent for a reason.*: There are no directly related issues for this because it is already semi-documented behavior. That being said, there is no runtime error/warning when this happens. Tangentially related issues: https://github.com/sveltejs/sapper/issues/843, https://github.com/sveltejs/sapper/issues/710 (another instance where re-rendering was unwanted from messing up `preload`), https://github.com/sveltejs/sapper/issues/447 (resolved issue that never got closed by the way; there is already an error for the same situation with the `session` store) 
- [x] *Ideally, include a test that fails without this PR but passes with it. PRs will only be merged once they pass CI. (Remember to `npm run lint`!)*: There is no way to mock/assert console output in `mocha`/`assert` without introducing another dependency like [`sinon`](https://sinonjs.org/releases/v9.0.2/spies/). By the way, there is no longer a `lint` package script.
### Tests
-  [x] *Run the tests tests with `npm test` or `yarn test`*: 1 test was already failing and this doesn't change that

Like was said in #447, if there is data in `preload`'s return that cannot be serialized by `devalue` (like a function in my case), the client will not receive a copy. There are no warnings for this (i.e. this is a completely silent failure), which was a source of confusion for me, despite this behavior being documented. (Actually, another PR to specify what it means [when serialization fails](https://sapper.svelte.dev/docs#Return_value) could be needed since it is not explicitly stated that `preload` runs again client-side rather than displaying an error page). 

This pull request introduces an error (where serialization failed and why) + warning (consequences and more information) message combo in the server side console while remaining silent in the client side. This approach was taken to remain backwards compatible and not bother the client with the server's mistakes. 

Possible problems or improvements still needed with this:
* There are two console calls (`error` then `warn`) rather than one
* The warning could be restricted to `dev`
* Being silent on the client is different behavior than what happens when the `session` store's serialization fails, which displays an error page (see 2 lines later in the same file)

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -267,7 +267,10 @@ export function get_page_handler(
 			const { html, head, css } = App.render(props);
 
 			const serialized = {
-				preloaded: `[${preloaded.map(data => try_serialize(data)).join(',')}]`,
+				preloaded: `[${preloaded.map(data => try_serialize(data, err => {
+					console.error(`Failed to serialize preloaded data to transmit to the client at the /${segments.join('/')} route: ${err.message}`);
+					console.warn('The client will re-render over the server-rendered page fresh instead of continuing where it left off. See https://sapper.svelte.dev/docs#Return_value for more information');
+				})).join(',')}]`,
 				session: session && try_serialize(session, err => {
 					throw new Error(`Failed to serialize session data: ${err.message}`);
 				}),


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -267,7 +267,10 @@ export function get_page_handler(
 			const { html, head, css } = App.render(props);
 
 			const serialized = {
-				preloaded: `[${preloaded.map(data => try_serialize(data)).join(',')}]`,
+				preloaded: `[${preloaded.map(data => try_serialize(data, err => {
```
benmccann, line 270: should we use `serialize_error` in the error case?

https://github.com/sveltejs/sapper/blob/427a6bca9a0dbca7fcf3efe0841f63fc50387e84/runtime/src/server/middleware/get_page_handler.ts#L382

babichjacob: Like the comment suggests, that (breaking) changes the current behavior from *server failing -> client trying again* to *server failing -> showing persistent error page*. I think an important contributor / maintainer / core developer would have to decide that.

benmccann: I believe we already do that for any other type of failure, so it would be the most consistent behavior. I'm not sure why a serialization failure would be treated any differently

babichjacob: Switching to that behavior in c391c9a makes [the `preloading/bails on custom classes returned from preload` test](https://github.com/sveltejs/sapper/blob/master/test/apps/preloading/test.ts#L47-L54) fail.

benmccann: My vote would be to update the test. Perhaps it would be considered a breaking change and we have to wait until 0.28.0 to merge it. Thoughts @antony @Conduitry ?

antony: This is certainly a behaviour I would prefer, yes. It is indeed breaking, but in the best possible way.

antony:
This looks good. Is there a good way to add a test for this?

babichjacob:
> This looks good. Is there a good way to add a test for this?

Sure; see the latest commit. 
One outstanding problem I have with it is that it's dependent on `devalue`'s particular error message. Would it be better to replace the `assert.equal` call with `assert.ok((await r.text('p')).startsWith('Failed to serialize preloaded data: '))`?

Conduitry:
I believe the original intention of this behavior was that, if there were no preload data sent from the server to the client (for example, if it could not be serialized), the client-side preload would be also be run to try to re-create it. I don't think this should be an outright error.

Also sort of adjacent: #843.

babichjacob:
Does that mean the most preferred behavior was in commit https://github.com/sveltejs/sapper/pull/1304/commits/427a6bca9a0dbca7fcf3efe0841f63fc50387e84 and the ones added since then should be dropped from this PR?

benmccann:
How about if we error out with a hard fail if we're in development mode, but allow the client to re-render over the top of it in production mode? That would make it easier to notice something's wrong in development, but not error unless necessary in production

benmccann:
Also, we made a very similar change in https://github.com/sveltejs/sapper/pull/719. Perhaps we'd want it to do the same?

Conduitry:
Having this be an error, even in dev mode, seems to indicate that we're telling the user they're doing something objectively wrong that we don't want to support, but that, for some technical/performance/whatever reason, we don't want to enforce it in production builds. I'm okay with making this a warning, but, again, making this be an error at all (even in dev mode) marks a conscious change in how we're saying this feature should work, and I don't think that change is warranted. It's perfectly possible to have a working (although non-optimal) app that re-runs preload on the client.

benmccann:
If you wanted to run something only on the client, I'd expect you'd put it in `onMount` instead of `preload`? Or put a check in `preload` for `window` or something to see if you're running on the client or server. I wouldn't expect to be able to write code that always fails on the server and then have the failure be ignored

pngwn:
I don't think this should be an error, this is a pretty significant breaking change and I don't think any of the arguments in favour are particularly convincing. Certainly not convincing enough to make this change without further discussion on what the behaviour _should_ be. This is the kind of breaking change that could require significant investment from users to fix and offers very little value in exchange.

I'd suggest going with a warning for now, if anything. At the very least we need a simple way for users to log these errors to their error reporting system before issuing this change.

Conduitry:
`preload` prevents the component from loading at all (and delays navigation) until the promises resolve, which `onMouint` cannot do. IMO you're not writing code that fails on the server, you're writing code that returns a value that we can't _serialize_ on the server.

Making this be an error feels to me like a frivolous, puritanical thing for Sapper to do.

benmccann:
I'm okay with the warning behavior @babichjacob initially proposed if everyone else is. I made a backup of this branch and then rewound it to the initial commit. Let me know if it looks okay now

ajbouh:
As a user I've been very surprised in a bad way by the existing behavior.
It's only after I add a lot more logging that I realize that I'm preloading
twice because I'm accidentally trying to serialize something that can't be
serialized.

By swallowing the failure we put people that are trying to build fast,
efficient services at a meaningful disadvantage. :/

PR #1309, opened 2020-07-09
[WIP, please help improve] Fix #699 when importing css using postcss

### LINKED ISSUE #699: rollup-plugin-postcss not working
I can't seem to get postcss to work properly and extract my css like it does in pure svelte.
With css extract option on it generates: 
```js
export default undefined;
export default undefined;
/* sourceMappingURL=./chunk.e3537e20.css.map */
```
but with it off, it generates:
```js
var css = "a.link.svelte-131civd{color:#d3d3d3}";
export default css;
import styleInject from 'C:/Users/admin/Code/domainhax/node_modules/style-inject/dist/style-inject.es.js';
styleInject(css);
var css = "main.svelte-1joomdn{padding-right:15px;padding-left:15px;margin-right:auto;margin-left:auto;width:64em}input.svelte-1joomdn{width:100%}#description.svelte-1joomdn>div.svelte-1joomdn{width:50%;padding:0;float:left}header.svelte-1joomdn{text-align:center}p.svelte-1joomdn{margin-top:0}a.svelte-1joomdn{color:#959595}h1.svelte-1joomdn a.svelte-1joomdn{color:#dfdfdf}";
export default css;
import styleInject from 'C:/Users/admin/Code/domainhax/node_modules/style-inject/dist/style-inject.es.js';
styleInject(css);
/* sourceMappingURL=./chunk.e3537e20.css.map */
```

In normal svelte when I do `import './global.css'`, it generates a inlined bit of js with it off, and the processed file with it on, and does the same for svelte's generated CSS.
I can't figure out what's going wrong, but I would like my css minified and prefixed.

Evertt:
All the bug info can be found in #699, but here's a short summary.

When you use `import './styles.css'` in your Svelte or JS code and you use `rollup-plugin-postcss` to process that css file first, then your css output looks like javascript. Depending on your configuration it could look like:

```js
export default undefined;
export default undefined;
/* sourceMappingURL=./chunk.e3537e20.css.map */
```

or

```js
var css = "a.link.svelte-131civd{color:#d3d3d3}";
export default css;
import styleInject from 'C:/Users/admin/Code/domainhax/node_modules/style-inject/dist/style-inject.es.js';
styleInject(css);
var css = "main.svelte-1joomdn{padding-right:15px;padding-left:15px;margin-right:auto;margin-left:auto;width:64em}input.svelte-1joomdn{width:100%}#description.svelte-1joomdn>div.svelte-1joomdn{width:50%;padding:0;float:left}header.svelte-1joomdn{text-align:center}p.svelte-1joomdn{margin-top:0}a.svelte-1joomdn{color:#959595}h1.svelte-1joomdn a.svelte-1joomdn{color:#dfdfdf}";
export default css;
import styleInject from 'C:/Users/admin/Code/domainhax/node_modules/style-inject/dist/style-inject.es.js';
styleInject(css);
/* sourceMappingURL=./chunk.e3537e20.css.map */
```

I saw that the bug has gone unfixed for a long time. Now I don't understand Sapper well enough to understand all the intricacies that I need to be mindful of when fixing a bug. So I'm aware that this PR is probably relatively bad. I didn't run the tests, but when I run `npm run dev` I get the following warning repeatedly:

    Sourcemap is likely to be incorrect: a plugin (sapper-internal) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help

But other than that, it now does output the css correctly, finally. So is anyone willing to take a look at my PR and improve whatever needs to be improved before it can be merged?

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -39,12 +39,14 @@ export default class RollupCompiler {
 				this.chunks.push(chunk);
 			},
 			transform: (code: string, id: string) => {
-				if (/\.css$/.test(id) && !/styleInject/.test(code)) {
-					this.css_files.push({ id, code });
-					return ``;
-				} else if (/styleInject/.test(code)) {
+				if (!/\.css$/.test(id)) return;
+				
+				if (/styleInject/.test(code)) {
 					return code;
 				}
+				
+				this.css_files.push({ id, code });
+				return ``;
 			}
 		});
 
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -39,9 +39,11 @@ export default class RollupCompiler {
 				this.chunks.push(chunk);
 			},
 			transform: (code: string, id: string) => {
-				if (/\.css$/.test(id)) {
+				if (/\.css$/.test(id) && !/styleInject/.test(code)) {
 					this.css_files.push({ id, code });
 					return ``;
+				} else if (/styleInject/.test(code)) {
+					return code;
 				}
 			}
 		});


src/core/create_compilers/RollupCompiler.ts
```diff
@@ -39,10 +39,14 @@ export default class RollupCompiler {
 				this.chunks.push(chunk);
 			},
 			transform: (code: string, id: string) => {
-				if (/\.css$/.test(id)) {
-					this.css_files.push({ id, code });
-					return ``;
+				if (!/\.css$/.test(id)) return;
+				
+				if (/styleInject/.test(code)) {
```
benmccann, line 44: There's a lot to comprehend regarding this bug and I'm not sure I totally understand it yet because I'm not that familiar with PostCSS, etc. Looking for `styleInject` seems too specific to this usecase though and I think it'd be better handled in a more generic way. The first line of this function is checking to make sure that the code is CSS. Perhaps the issue here is that you actually have JavaScript in a file with a `.css` extension? It seems weird to me that you would have that, but perhaps the fix should be to check if the file contents are JavaScript instead of CSS? But I'm not quite sure a good way of doing that. Looking for something like `export` seems a bit better, but also seems like it could cause problems. E.g. what if the user has some data export functionality on their site and has a CSS class named `.export`?

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -39,10 +39,14 @@ export default class RollupCompiler {
...
+					return code;
```
benmccann, line 45: What's the difference between `return;` like on the first line and `return code;` that you have here?

benmccann:
Can you explain why you went with this approach instead of the one suggested by @krzysztof-grzybek in https://github.com/sveltejs/sapper/issues/699#issuecomment-642460603?

We should also test that any solution would work with code splitting. Code splitting currently only works with Rollup 1. We should probably merge https://github.com/sveltejs/sapper/pull/1306, which I sent yesterday and touches this same block of code, to fix code splitting support for Rollup 2 to make it easier to test since a lot of the sample apps use Rollup 2

src/core/create_compilers/RollupCompiler.ts
Evertt: In this case I'm just mimicking the code that was there previously. That also either returned nothing (if the id did not test positive for `/\.css$/`) or it returned an empty string.

Evertt:
> Can you explain why you went with this approach instead of the one suggested by @krzysztof-grzybek in #699 (comment)?

Simply because I don't understand it enough to be able to implement it. I tried to play a little bit with the `writeBundle` hook, but I just got more errors that I didn't understand.

> We should also test that any solution would work with code splitting. Code splitting currently only works with Rollup 1. We should probably merge #1306, which I sent yesterday and touches this same block of code, to fix code splitting support for Rollup 2 to make it easier to test since a lot of the sample apps use Rollup 2

👍

benmccann:
I appreciate you sharing your findings. I actually think it'd be more helpful to close this PR though. We have a very long review queue and it makes it harder to know which PRs need review when we have PRs that aren't in a mergable state in the review queue

Evertt:
👍

Issue #584, opened 2019-02-28
Prevent scroll to top when calling goto from method


cbenz:
I'd like to prevent scroll to top when I click on a `<button on:click=search()>`, and the `search` method calls `goto("?q=foo")`.

Prevent scroll seems important in order to keep the user focused on the search form which is not at the top of the page.

I would have the same need with a `<select>` element: changing its value would call `goto("?selected=foo")`.

Is it possible?

Cf #376

cbenz:
here's a quick codesandbox showing the issue: https://codesandbox.io/s/16zxv8ovj

thgh:
This could be implemented as an option of goto:
```js
goto('?q=foo', { scroll: false })
```

Implementation
```diff
- return navigate(target, null).then(() => {}); 
+ return navigate(target, null, opts.scroll === false).then(() => {}); 
```
https://github.com/sveltejs/sapper/blob/411e2594af6f46ed73c45c1f0db0746adf20bdb5/runtime/src/app/goto/index.ts#L8

flayks:
+1

PR #1320, opened 2020-07-15
Add noscroll option to goto, document sapper-noscroll on links

Jayphen:
This adds an option to `goto` - `noscroll: boolean` - which allows preservation of scroll position when navigating programmatically.

`noscroll` already exists as an option on `navigate`, and this just passes the option through to that underlying method.

I've added tests that show the default behaviour when using `goto` (scrolling to the top), and the preservation behaviour when the option is passed.

This implements https://github.com/sveltejs/sapper/issues/584

**Note**: In working on this, I discovered a bug in `goto`/`navigate` - the scroll position is not preserved when using `noscroll: true` when only the search params are updated (not the pathname). I will add a separate issue for this.

edit: [added issue with test cases](https://github.com/sveltejs/sapper/issues/1321)

- [-] It's really useful if your PR relates to an outstanding issue, so please reference it in your PR, or create an explanatory one for discussion. In many cases features are absent for a reason.
- [-] This message body should clearly illustrate what problems it solves. If there are related issues, remember to reference them.
- [-] Ideally, include a test that fails without this PR but passes with it. PRs will only be merged once they pass CI. (Remember to `npm run lint`!)
### Tests
-  [-] Run the tests tests with `npm test` or `yarn test`)

### DIFF
--- a/site/content/docs/03-client-api.md
+++ b/site/content/docs/03-client-api.md
@@ -27,7 +27,9 @@ sapper.start({
 ### goto(href, options?)
 
 * `href` — the page to go to
-* `options` — can include a `replaceState` property, which determines whether to use `history.pushState` (the default) or `history.replaceState`. Not required
+* `options` - not required
+  * `replaceState` (`boolean`, default `false`) — determines whether to use `history.pushState` (the default) or `history.replaceState`.
+  * `noscroll` (`boolean`, default `false`) — prevent scroll to top after navigation.
 
 Programmatically navigates to the given `href`. If the destination is a Sapper route, Sapper will handle the navigation, otherwise the page will be reloaded with the new `href`. In other words, the behaviour is as though the user clicked on a link with this `href`.
 
--- a/site/content/docs/03-client-api.md
+++ b/site/content/docs/03-client-api.md
@@ -27,7 +27,7 @@ sapper.start({
 ### goto(href, options?)
 
 * `href` — the page to go to
-* `options` - not required
+* `options` — not required
   * `replaceState` (`boolean`, default `false`) — determines whether to use `history.pushState` (the default) or `history.replaceState`.
   * `noscroll` (`boolean`, default `false`) — prevent scroll to top after navigation.
 
--- a/site/content/docs/08-link-options.md
+++ b/site/content/docs/08-link-options.md
@@ -31,3 +31,15 @@ Adding a `rel=external` attribute to a link...
 ```
 
 ...will trigger a browser navigation when the link is clicked.
+
+### sapper-noscroll
+
+When navigating to internal links, Sapper will change the scroll position to 0,0 so that the user is at the very top left of the page. When a hash is defined, it will scroll to the element with a matching ID.
+
+In certain cases, you may wish to disable this behaviour. Adding a `sapper-noscroll` attribute to a link...
+
+```html
+<a href='path' sapper-noscroll>Path</a>
+```
+
+...will prevent scrolling after the link is clicked.
--- a/site/content/docs/08-link-options.md
+++ b/site/content/docs/08-link-options.md
@@ -32,14 +32,14 @@ Adding a `rel=external` attribute to a link...
 
 ...will trigger a browser navigation when the link is clicked.
 
-### sapper-noscroll
+### sapper:noscroll
 
 When navigating to internal links, Sapper will change the scroll position to 0,0 so that the user is at the very top left of the page. When a hash is defined, it will scroll to the element with a matching ID.
 
-In certain cases, you may wish to disable this behaviour. Adding a `sapper-noscroll` attribute to a link...
+In certain cases, you may wish to disable this behaviour. Adding a `sapper:noscroll` attribute to a link...
 
 ```html
-<a href='path' sapper-noscroll>Path</a>
+<a href='path' sapper:noscroll>Path</a>
 ```
 
 ...will prevent scrolling after the link is clicked.
--- a/test/apps/scroll/test.ts
+++ b/test/apps/scroll/test.ts
@@ -1,133 +1,133 @@
 import * as assert from 'assert';
-import {build} from '../../../api';
-import {AppRunner} from '../AppRunner';
+import { build } from '../../../api';
+import { AppRunner } from '../AppRunner';
 
-describe('scroll', function () {
-  this.timeout(10000);
+describe('scroll', function() {
+	this.timeout(10000);
 
-  let r: AppRunner;
+	let r: AppRunner;
 
-  // hooks
-  before('build app', () => build({cwd: __dirname}));
-  before('start runner', async () => {
-    r = await new AppRunner().start(__dirname);
-  });
+	// hooks
+	before('build app', () => build({ cwd: __dirname }));
+	before('start runner', async () => {
+		r = await new AppRunner().start(__dirname);
+	});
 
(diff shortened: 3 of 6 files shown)


antony:
@Jayphen does this supercede the other PR?

test/apps/scroll/test.ts
```diff
@@ -1,107 +1,133 @@
...
-		await r.wait();
-		const secondScrollY = await r.page.evaluate(() => window.scrollY);
-
-		assert.equal(firstScrollY, secondScrollY);
-	});
-
-	it('survives the tests with no server errors', () => {
-		assert.deepEqual(r.errors, []);
-	});
+describe('scroll', function () {
```
antony, line 5: Please don't reformat all the code as part of a PR. Our linting rules aren't really enforced right now, but the existing style is the style we want to keep.

Jayphen:
@antony Yes, the other can be closed. It did not actually implement the change correctly.

test/apps/scroll/test.ts
Jayphen: Yeah, I am trying to fix this now. I guess my editor used global default Prettier settings since this project has no config.

Jayphen: Alright, I reverted the spacing changes. Sorry about that.

benmccann:
Can you add this to the documentation as well? https://github.com/sveltejs/sapper/tree/master/site/content/docs

Jayphen:
Sure, should I also document the currently undocumented noscroll attribute that can be used on anchors?

benmccann:
Yes, that'd be great. Thanks!

Jayphen:
@benmccann Okay, I've added some docs. I copied the style of the Svelte docs for describing the type and default for each option: https://svelte.dev/docs#flip

site/content/docs/03-client-api.md
```diff
@@ -27,7 +27,9 @@ sapper.start({
 ### goto(href, options?)
 
 * `href` — the page to go to
-* `options` — can include a `replaceState` property, which determines whether to use `history.pushState` (the default) or `history.replaceState`. Not required
+* `options` - not required
```
benmccann, line 30: the dash on this line was changed so that it no longer matches the others on the page

test/apps/scroll/test.ts
```diff
@@ -99,7 +99,33 @@ describe('scroll', function() {
 		const secondScrollY = await r.page.evaluate(() => window.scrollY);
 
 		assert.equal(firstScrollY, secondScrollY);
-	});
+  });
```
benmccann, line 102: the newly added lines use spaces whereas the existing lines use tabs. should switch to tabs for consistency

Jayphen: I've converted all the lines I touched to tabs.

site/content/docs/08-link-options.md
```diff
@@ -31,3 +31,15 @@ Adding a `rel=external` attribute to a link...
...
 ...will trigger a browser navigation when the link is clicked.
+
+### sapper-noscroll
+
+When navigating to internal links, Sapper will change the scroll position to 0,0 so that the user is at the very top left of the page. When a hash is defined, it will scroll to the element with a matching ID.
+
+In certain cases, you may wish to disable this behaviour. Adding a `sapper-noscroll` attribute to a link...
+
+```html
+<a href='path' sapper-noscroll>Path</a>
```
lukeed, line 42: Personally I'd prefer `noscroll` attribute. Maybe others can weigh in?
```suggestion
<a href='path' noscroll>Path</a>
```

arxpoetica: @lukeed agree

pngwn: `stickywhenclicky` imo

lukeed: 😂

Jayphen: This change is not relevant to this PR — I only documented why currently exists

Jayphen: Prefixing with sapper is obviously the safer option, and likely was done to avoid any risk of colliding with future spec changes. I think this discussion should be taken to an issue

lukeed: Correct, I'm asking for input before we officially document something

Jayphen: Fair enough.

Personally I am wary of adding arbitrary attributes to DOM elements without some kind of namespace, as it could lead to collisions in the future and with little benefit (except perhaps looking a little cleaner). For this reason I’d have actually used a data attribute (which also would make this valid markup)

benmccann: Some options:
* `sapper-noscroll` (verbose. but would avoid breaking change)
* `noscroll` (nice looking, but not namespaced)
*  `s:noscroll` (angular-like. perhaps svelte would want s namespace in the future though?)
*  `s-noscroll` (vue-like. perhaps svelte would want s namespace in the future though?)
*  `data-noscroll` (valid markup)
*  `data-s-noscroll` (valid markup and namespaced, but verbose)
*  `data-s:noscroll` (valid markup and namespaced, but verbose)

Probably `s-noscroll` and `data-s:noscroll` would be my favorites. I like that the latter is valid markup. It's a bit more verbose though. Right now that's probably fine because I don't know how much this option will be used, but maybe that'd be more annoying if we add more frequently used options with the same syntax?

ignatiusmb: What about `sapper:noscroll`?
https://github.com/sveltejs/sapper/blob/ca3645768597035b08e9d0e025cabd062180184a/runtime/src/app/start/index.ts#L113
Since it's not accessed through `.dataset`, I think having a `data-` prefix (even though valid HTML) is redundant

Jayphen: The fact that it *could* be accessed via dataset & also via a data attribute selector in css (and is valid HTML) are all strong points for. Are there any points against other than aesthetics/verbosity?

Jayphen: As a side note to the actual decision — to change this will be a breaking change, despite it being previously undocumented. I am not sure that it is reasonable to wrap a breaking change up with my PR here. Perhaps I should remove the documentation of `sapper-noscroll` from this PR, and make a new one where this discussion can continue?

antony: I think I prefer `sapper-noscroll` which makes it clear that it's a Sapper specific thing. 'noscroll' feels like an built-in attribute.

benmccann: I took a poll of the maintainers offline and there was a preference for keeping noscroll namespaced. However, folks preferred `sapper:noscroll` to `sapper-noscroll`, so I updated this PR to rename the attribute.

I'll go ahead and merge this now. Thanks for the patience and input everyone!

jhwheeler: I also prefer `sapper:noscroll`, but I'm curious, are there other namespaced attributes for html elements?

Conduitry: Not currently, but that's where things are going to be moving in the future. `rel="prefetch"` (which, when used on an anchor element, is a Sapper-only thing, which is confusingly named the same as a real HTML thing) is going to be renamed to some other namespaced `sapper:` attribute.

jhwheeler: Great, I like this idea. Much clearer!

Issue #1243, opened 2020-05-28
Application initialization - making fetch available outside preload


benmccann:
**Is your feature request related to a problem? Please describe.**

Apollo client needs to be given the `fetch` function during setup of the client while setting up the application. The `fetch` method still isn't actually called until executing the client during `preload`

Today there's no way to do that and so people are coming up with workarounds https://github.com/timhall/svelte-apollo/issues/9:
* Using [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch), [isomorphic-unfetch](https://www.npmjs.com/package/isomorphic-unfetch), or [cross-fetch](https://www.npmjs.com/package/cross-fetch) in order to provide their own `fetch` function bypassing Sapper's provided functionality. But this won't pass headers (i.e. cookies and auth headers from the original request) and so requires a fair amount of work to make that actually work
* Just using `onMount` and skipping SSR altogether 

You can't create the Apollo client in `preload` where `this.fetch` is available because then it will only get created on the server during initial page load and when you navigate to another page it won't have been created.

**Describe the solution you'd like**
I'm thinking it might be good to have some kind of `init` method that gets called at the beginning of the application on both the client and server in order to setup things like data fetchers and takes an object containing `fetch` as a parameter

**Describe alternatives you've considered**
I will probably try to use `cross-fetch` and not use Sapper's `fetch`. It's a bit of a Rube Goldberg making it all fit together though. I haven't figured out how I'll pass along cookies and auth headers on the server from my original request to the API server

**How important is this feature to you?**
There are probably workarounds. I've really been tearing my hair out trying to figure out how to get this wired up though.

GraphQL is extremely popular. It'd be nice if Sapper had better support for utilizing GraphQL clients.

**Additional context**
In addition to this issue, users also need to create per-request context on the server to hold the client. It took me quite a lot of investigation to find a solution seems doable. (https://github.com/sveltejs/sapper/issues/751#issuecomment-516882258)

Drevoed:
Any news on this?

Issue #908, opened 2019-09-20
Export and cross-fetch don't play nice together


arxpoetica:
**[Is your feature request related to a problem? Please describe.]**
I spent a day and a half trying to figure out why export wasn't crawling any of my `*.json.js` files. The files were *generated* on export, just not crawled. These are from fetch API calls in preload functions/scripts.

I'm tempted to call this a bug, but I discovered that when I started using `this.fetch` instead of [cross-fetch](https://www.npmjs.com/package/cross-fetch), it started working correctly. So technically this isn't a bug if you use the preloader `this.fetch` API correctly.

**[Describe the solution you'd like]**
But that makes creating fetch API wrappers hard. I either have to explicitly pass `this.fetch` as an argument to a wrapper function. (I create reusable `GET` and `POST` utility libraries/functions). Or I have to write out the same boilerplate again and again.

I don't have a strict recommendation—maybe one would be to make `this.fetch` universally available—that's problematic too though. I'm tempted to say, let's actually strip out the `this.fetch` feature entirely, recommend using something like cross-fetch in its place, and then just fix that "bug."

**[Describe alternatives you've considered]**
None.

**[How important is this feature to you?]**
Dunno. Maybe this is a "just me" issue, and it should be closed—maybe no one else will run into this. I thought it was worth discussing though.

sw-yx:
i think this neglects the reason this.fetch exists in the first place. it argues for better documentation on why sapper needs you to use this.fetch.

Conduitry:
Something else that's come up as a possibility is allowing users to attach their own stuff to the `this` object that preloads are called with, so you can have your own helpers that still use `this.fetch` internally. The question then would be where this ought to live, since we'd need to attach this stuff both on the server and the client.

Conduitry:
> I'm tempted to say, let's actually strip out the `this.fetch` feature entirely, recommend using something like cross-fetch in its place, and then just fix that "bug."

This isn't particularly viable as far as I can tell, as the `this.fetch` behavior needs to be rather different on the server and on the client in order for it to be called identically in both places and just work. On the client, it's just a regular `fetch`, but on the server, you can see all the extra stuff we have to do in the `get_page_handler.ts` file.

arxpoetica:
> as the this.fetch behavior needs to be rather different on the server and on the client

Agreed. Arguably, `this.fetch` is doing it's job perfectly reasonably, without too much boilerplate, which is actually a good API.

[...]

PR #1336, opened 2020-07-23
Preload context configuration

benmccann:
This makes a lot of things easier:
* Distributed tracing - add a header to every request to trace requests from frontend all the way to datastore (requested by @adamb on Discord. difficulty setting up honeycomb.io also mentioned in https://github.com/sveltejs/sapper/issues/751)
* Using an internal hostname on the server-side (that's what I want this for. my api server is written in another language)
* Using `fetch` in setup - e.g. to setup Apollo client (https://github.com/sveltejs/sapper/issues/1243 https://github.com/sveltejs/sapper/issues/751)
* Specifying a base URL (https://github.com/sveltejs/sapper/issues/489)
* Using a UNIX socket for AWS Lambda (https://github.com/sveltejs/sapper/issues/356)
* Ability to wrap `fetch` (https://github.com/sveltejs/sapper/issues/908)
* It currently requests only over HTTP. A user on Discord reported this wouldn't work with his server than listens only on port 443
* Accessing Polka request on server (https://github.com/sveltejs/sapper/issues/1436)

This was also somewhat suggested by @Conduitry in https://github.com/sveltejs/sapper/issues/908#issuecomment-533636138

I exported the default fetch method so that users can do their own logic first (modifying the URL, logging the request, setting a header, etc.) and then call the default fetch method.

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -20,6 +20,8 @@ import {
 import goto from './goto';
 import { page_store } from './stores';
 
+export type ClientPreloadConfigurator = (args: {context: object}) => void;
+
 declare const __SAPPER__;
 export const initial_data = typeof __SAPPER__ !== 'undefined' && __SAPPER__;
 
@@ -67,8 +69,10 @@ export function set_prefetching(href, promise) {
 }
 
 export let target: Element;
-export function set_target(element) {
-	target = element;
+let preload_configurator: ClientPreloadConfigurator;
+export function set_options(opts) {
+	target = opts.target;
+	preload_configurator = opts.preloadConfigurator || (() => {});
 }
 
 export let uid = 1;
@@ -294,6 +298,8 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 			props.status = status;
 		}
 	};
+
+	preload_configurator({context: preload_context, ...stores});
 
 	if (!root_preloaded) {
 		const root_preload = root_comp.preload || (() => {});
--- a/runtime/src/app/start/index.ts
+++ b/runtime/src/app/start/index.ts
@@ -7,15 +7,17 @@ import {
 	scroll_state,
 	select_target,
 	handle_error,
-	set_target,
+	set_options,
 	uid,
 	set_uid,
-	set_cid
+	set_cid,
+	ClientPreloadConfigurator
 } from '../app';
 import prefetch from '../prefetch/index';
 
 export default function start(opts: {
-	target: Node
+	target: Node,
+	preloadConfigurator: ClientPreloadConfigurator
 }): Promise<void> {
 	if ('scrollRestoration' in history) {
 		history.scrollRestoration = 'manual';
@@ -34,7 +36,7 @@ export default function start(opts: {
 		history.scrollRestoration = 'manual';
 	});
 
-	set_target(opts.target);
+	set_options(opts);
 
 	addEventListener('click', handle_click);
 	addEventListener('popstate', handle_popstate);
--- a/runtime/src/server/index.ts
+++ b/runtime/src/server/index.ts
@@ -1,1 +1,2 @@
+export { server_fetch as fetch } from './middleware/get_page_handler';
 export { default as middleware } from './middleware/index';
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -3,15 +3,55 @@ import fs from 'fs';
 import path from 'path';
 import cookie from 'cookie';
 import devalue from 'devalue';
-import fetch from 'node-fetch';
+import node_fetch from 'node-fetch';
 import URL from 'url';
 import { sourcemap_stacktrace } from './sourcemap_stacktrace';
 import { Manifest, ManifestPage, Req, Res, build_dir, dev, src_dir } from '@sapper/internal/manifest-server';
 import App from '@sapper/internal/App.svelte';
 
+export function server_fetch(req: Req, res: Res, url: string, opts?: any) {
+	const parsed = new URL.URL(url, `http://127.0.0.1:${process.env.PORT}${req.baseUrl ? req.baseUrl + '/' :''}`);
+
+	opts = Object.assign({}, opts);
+
+	const include_credentials = (
+		opts.credentials === 'include' ||
+		opts.credentials !== 'omit' && parsed.origin === `http://127.0.0.1:${process.env.PORT}`
+	);
+
+	if (include_credentials) {
+		opts.headers = Object.assign({}, opts.headers);
+
+		const cookies = Object.assign(
+			{},
+			cookie.parse(req.headers.cookie || ''),
+			cookie.parse(opts.headers.cookie || '')
+		);
+
+		const set_cookie = res.getHeader('Set-Cookie');
+		(Array.isArray(set_cookie) ? set_cookie : [set_cookie]).forEach(str => {
+			const match = /([^=]+)=([^;]+)/.exec(<string>str);
+			if (match) cookies[match[1]] = match[2];
+		});
+
+		const str = Object.keys(cookies)
+			.map(key => `${key}=${cookies[key]}`)
+			.join('; ');
+
+		opts.headers.cookie = str;
+
+		if (!opts.headers.authorization && req.headers.authorization) {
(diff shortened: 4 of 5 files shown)


Conduitry:
This looks like it only replaces the server-side `this.fetch`, and leaves the client-side `this.fetch` to be the same as the browser-provided `fetch`. What I've run into before - both in questions in chat and in stuff I've run into in real life - is a common 'wrapper' around fetch that I wanted to use in both the server and the client.

I don't know what this API would look like, but one thing that was suggested some time ago (I don't remember whether it ever made its way into an issue) was a function that would be called on the preload `this` context before each preload and can do whatever it wants with that - like adding brand new methods, etc. If this were something that's to be called on the preload context on both the server and the client, I don't yet think we have a good place for this to live.

antony:
Just to add my 2 cents to this. I agree with the feature also - but I prefer the implementation discussed by @Conduitry.

I currently eschew use of the built in `fetch` directly by wrapping it in https://github.com/beyonk-adventures/sapper-httpclient which has handlers for errors, and things like auth, easier destructuring, and base urls, to name a few things.

Right now I just wrap `this.fetch` or `fetch` with it, as I use it. However I'd love to be able to push it into a context and have it available in the same way that `fetch` is. So absolutely in favour of that solution.

benmccann:
That seems reasonable. I agree it'd be useful on the client side as well. I think users would probably want to be able to set different values on client and server (e.g. maybe on the server you wrap fetch to log calls to the file system, which wouldn't make sense on the client).

benmccann:
I updated this to be inline with the proposed implementation. It could use some documentation and probably tests, but I wanted to get feedback on the approach before taking that step

ajbouh:
My experiments have suggested that on the server we need to be able to set certain values as a function of the request. These are:
- fetch
- preload execution
- session
- code to make initial calls to setContext (parameterized by app stores)

On the client we need to set the following, as a function of the initial session
- fetch
- preload execution
- code to make initial calls to setContext (parameterized by app stores)

My Typescript changes add most of these

pngwn:
Should this be an RFC? I don’t know that there is consensus about how we should do this, what the API should look like and whether or not that API will make sense moving  forwards. 

Feels like there is some discussion required before implementing.

ajbouh:
How do we prevent things from stalling while we wait for consensus?

I'm personally a fan of working code to illustrate proposals and provide a default "ready to merge" option for people to use in their own projects today

pngwn:
An RFC does the opposite. PRs that we can’t agree on sit around to rot. RFCs encourage the higher level discussion we to make sure it is the right choice. Those RFCs can be accompanied by a POC of preferred.

benmccann:
Discussion with @ajbouh over Discord:
* `node-fetch` provides a different interface than `window.fetch`. E.g. first parameter can be a request object. Do we want to expose that?
* he's instrumented `preload` in his project to provide timing information, etc. Also wants to know if `preload` had an error or succeeded
  * tracing tools like honeycomb and opentelemetry (which is merging of opencensus and opentelemetry) provide ways to do inter-service and intra-service tracing

RetWolf:
Curious where we stand with this? This adds quite a bit of nice to have functionality that I currently have a few hacky workarounds to achieve but it would be great to have some of this stuff supported by default. My specific use cases are:
-   Configuring base URL
-   Using fetch in setup (using a GraphQL client)

Is there somewhere I should be following for updates on this?

benmccann:
> Curious where we stand with this? 

This PR reflects the current state. I will probably try to put forth an updated design and/or RFC at some point, but haven't had the time yet since I've been focusing on other areas of sapper.

> Is there somewhere I should be following for updates on this?

Feel free to subscribe to this PR to get updates.

ajbouh:
I'm also experimenting with this on my fork

ehrencrona:
First out: It seems like all the use cases revolve around configuring or replacing `fetch`. Might it be sensible to just have a "FetchConfigurator" instead?

If we want to allow extending the context, I'd propose using the following type definitions for the configurators:

```
type PreloadContextDecorator<T extends PreloadContext> = (
	context: PreloadContext,
	clientContext: { session: any, page: Page }
) => T;
type ServerPreloadContextDecorator<T extends PreloadContext> = (
	context: PreloadContext,
	serverContext: { session: any; page: Page; req: Req; res: Res }
) => T;
```

Reasons:
 * It's a decorator pattern in pattern terminology so I'd advocate "decorator" rather than "configurator". Not sure how many people are familiar with pattern terminology anymore, though. An instance would be called `decoratePreloadContext`.
 * I think it's nicer if it's functional; i.e. returns a new context instance (internally, it could still just be modifying the one passed and returning it), also for type safety
 * I think there should be explicit types stating what is passed in
 * I added `session` and `page` on the server, so it's easier to write isomorphic code (it's a pity `session` is `any`).
 * Using generics enables type safety for the project code. It would define a `class MyPreloadContext extends PreloadContext` and then have `const decoratePreloadContext: PreloadContextDecorator<MyPreloadContext>`.
 * `PreloadContextDecorator` could be used either client- or server-side, `ServerPreloadContextDecorator` could only be used on the server side.

(I'm assuming `PreloadContext` as per #1468)

benmccann:
My first attempts just allowed configuring `fetch`, but I got the feedback that I should allow handling the full context, which I think can make sense since it's more flexible

I like the name decorator much more than configurator. But I may need to rethink if this is the correct approach. [Adam suggested that he'd like before and after calls or an "around" call](https://github.com/sveltejs/sapper/pull/1336#issuecomment-671503339)

ehrencrona:
If you want to allow maximum flexibility you could configure a preload function instead of a configurator/decorator, e.g.:

```
type PreloadFn = (
	this: PreloadOperations,
	page: Page,
	session: any,
	context: {
		componentPreload: PreloadFn,
		// only on server of course
		req: Req,
		res: Res
	}
) => Object | Promise<Object>;
```

This function would be responsible for doing the call to the component's actual preload function (passed as `componentPreload`). It could decide whether to modify/extend the context, to do anything before or after, or modify the result of preload (e.g. by always injecting a certain property, such as the time preload took).

(another idea I built in there is calling the object containing fetch/redirect and other functions `operations` and leaving `context` for variables; I left `page` and `session` out of it for backwards compatibility)

ajbouh:
My current approach is to use a function that returns these customized callbacks. 

On the client, this function is parameterized by the initial server-provided session.

On the server this function is parameterized by the incoming request and outgoing response. On the server this callback looks like:

```typescript
sapper.middleware({
  seed: (req, res) => ({
    fetch,
    preload: async function (
      ctx: sapper.PreloadContext<Fetch>,
      preload: sapper.Preload<Fetch, any, sapper.Session>,
      page: sapper.Page,
      session: sapper.Session,
    ) {
      const start = Date.now()
      try {
        return await preload.call(ctx, page, session)
      } finally {
        console.log("preload took", Date.now() - start, "ms")
      }
    },
    layout: ({setContext, stores}) => {
      // this sets initial values seen in .svelte files that use getContext(...)
      setContext('foo', ...);
    }
  }),
```

The backing types for this look something like:

```typescript
export type PreloadContext<Fetch> = {
  fetch: Fetch;
  redirect: (statusCode: number, location: string) => void;
  error: (status: number, error: Error | string) => void;
};

export type Preload<Fetch, Props, S extends Session = Session> = (
  this: PreloadContext<Fetch>,
  page: Page,
  session: S,
) => Props | Promise<Props>

export type Preloader<Fetch, Props> = (
  ctx: PreloadContext<Fetch>,
  preload: Preload<Fetch, Props>,
  page: Page,
  session: Session,
) => (Props | Promise <Props>)

export type Stores<Fetch> = {
  page: Readable<Page | {}>
  preloading: Readable<unknown>
  session: Readable<Session>
  fetch: Readable<Fetch>
}

export type ContextInitInput<Stores> = {
  setContext<T>(key: ContextKey<T>, value: T): void
  stores: Stores
}

export type ContextInit<Stores> = (input: ContextInitInput<Stores>) => void

export type BaseContext<Fetch, Props> = {
  fetch: Fetch;
  preload: Preloader<Fetch, Props>
  layout?: ContextInit<Stores<Fetch>>
}

export type BaseContextSeed<Req, Res> = (req: Req, res: Res) => BaseContext<Fetch, unknown>
```

benmccann:
I've filed a feature request for this in SvelteKit (https://github.com/sveltejs/kit/issues/327), so will go ahead and close this PR

Issue #1333, opened 2020-07-22
Tutorial does not work


SteffRainville:
**Describe the bug**
Tried the LEARN sapper path found at https://sapper.svelte.dev/docs/

npx degit "sveltejs/sapper-template#rollup" my-app
or: npx degit "sveltejs/sapper-template#webpack" my-app
cd my-app
npm install
npm run dev

Went to localhost:3000 and got

500
Cannot read property 'map' of undefined

TypeError: Cannot read property 'map' of undefined
    at load_component (http://localhost:3000/client/client.6eb8927e.js:2093:75)
    at http://localhost:3000/client/client.6eb8927e.js:2044:50
    at Array.map (<anonymous>)
    at hydrate_target (http://localhost:3000/client/client.6eb8927e.js:2028:42)
    at navigate (http://localhost:3000/client/client.6eb8927e.js:1894:3)
    at handle_click (http://localhost:3000/client/client.6eb8927e.js:2211:3)

The OR path of the tutorial seem to work
npx degit "sveltejs/sapper-template#webpack" my-app

Windows 10 64 bit
Node v14.5.0

benmccann:
I just tried and it worked fine for me on Ubuntu and Node 12.

SteffRainville:
Just tested on Mac 10.15.6 with node 14.5.0 and it works so the problem seems to be limited to Windows

nickolasgregory:
Working OK for me on
* Windows 10 64 bit
* Node v12.18.3

antony:
Working ok for me on Ubuntu 20.04
Node v14

SteffRainville:
Different message this morning on a different Windows 10 computer using node 14.6.0

c:\TMP>npx degit "sveltejs/sapper-template#rollup" my-app
npx: installed 1 in 2.407s
! could not fetch remote https://github.com/sveltejs/sapper-template
! could not find commit hash for rollup

c:\TMP>npx degit "sveltejs/sapper-template#webpack" my-app
npx: installed 1 in 2.602s
! could not fetch remote https://github.com/sveltejs/sapper-template
! could not find commit hash for webpack

antony:
I feel like that computr is a degit issue when the computer is not authenticated (Even though it's a public repo). See https://github.com/Rich-Harris/degit/issues/37

SteffRainville:
Confirmed ! The could not fetch issue was git related.

SteffRainville:
[...]

Issue #1256, opened 2020-06-06
"component.css is undefined" on fresh install


Pierstoval:
**Describe the bug**

After a fresh install, running the server outputs this in the browser:

![image](https://user-images.githubusercontent.com/3369266/83947974-ad664c00-a81a-11ea-9802-01259cba77c5.png)

**Logs**

<details>
  <summary>List of commands to reproduce</summary>

```
$ node --version
v14.4.0
$ npm --version
6.14.5
$ npx degit "sveltejs/sapper-template#rollup" sapper_test
npx : 1 installé(s) en 1.633s
> cloned sveltejs/sapper-template#rollup to sapper_test
$ cd sapper_test
$ npm install
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@~2.1.2 (node_modules\rollup\node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})
npm WARN TODO@0.0.1 No repository field.
npm WARN TODO@0.0.1 No license field.

added 272 packages from 182 contributors and audited 274 packages in 26.177s

19 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
$ npm run dev

> TODO@0.0.1 dev e:\dev\www\sapper_test
> sapper dev

✔ server (1.2s)
✔ client (1.2s)
✔ service worker (26ms)
> Listening on http://localhost:3000
```

</details>

I tested this on Node 12, installed Node 14 to check if I was not outdated just in case, and it doesn't work anymore.

**Stacktraces**

<details>
  <summary>Stack trace in the browser</summary>

```
load_component@http://localhost:3000/client/client.06dea69b.js:2088:20
hydrate_target/branch<@http://localhost:3000/client/client.06dea69b.js:2039:50
hydrate_target@http://localhost:3000/client/client.06dea69b.js:2023:42
navigate@http://localhost:3000/client/client.06dea69b.js:1889:3
start/<@http://localhost:3000/client/client.06dea69b.js:2148:22
```
</details>

**Information about your Sapper Installation:**

- Firefox 77.0.1 (64 bits)
- Windows 10
- Sapper version 0.27.13
- Svelte version 3.23.0
- With Rollup

**Severity**

High (project unusable)

[...]

PR #1338, opened 2020-07-25
Fix Sapper on Windows

benmccann:
Fixes https://github.com/sveltejs/sapper/issues/1333
Fixes https://github.com/sveltejs/sapper/issues/1256

It might only affect some versions of Windows since the tests were passing on GitHub Actions without this change

Thanks to @shackless for remote debugging this with me and testing!

### DIFF
--- a/test/unit/utils.test.ts
+++ b/test/unit/utils.test.ts
@@ -1,0 +1,9 @@
+import * as assert from 'assert';
+import { normalize_path } from '../../src/utils';
+
+describe('normalize_path', () => {
+	it('lowercases the first letter', () => {
+		assert.equal('c:\Users\simon\Source\my-app\src\routes\index.svelte',
+			normalize_path('C:\Users\simon\Source\my-app\src\routes\index.svelte'));
+	});
+});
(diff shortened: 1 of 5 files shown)


benmccann:
Added a test

test/unit/utils.test.ts
```diff
@@ -0,0 +1,9 @@
+import * as assert from 'assert';
+import { normalize_path } from '../../src/utils';
+
+describe('normalize_path', () => {
+	it('lowercases the first letter', () => {
```
antony, line 5: Awesome!

I think the `assert` signature is `assert.equal(actual, expected)` though, this is backwards.

benmccann: Ah, thanks. That's my Java/JUnit muscle memory. Updated!

Conduitry: The way it is now looks like the proper actual, expected ordering to me. However, the strings themselves in the tests need to have the backslashes escaped.

benmccann: good catch! updated

PR #1339, opened 2020-07-26
Fix for situations where you want to execute a popevent handler before page transitions start

monkeythedev:
When trying to execute code on popstate event, components transitions start before the popevent handler is executed whenever the component is already loaded.

What i'm trying to do is reversing the direction of my transition when the user clicks the back button:

**My popevent handler**
```
  function onPopState(e: PopStateEvent) {
    console.log("my pop event handler");
    if (window.location.href.includes("login")) $isHistoryBack$ = true;
  }

  onMount(() => {
    window.addEventListener("popstate", onPopState);

    ready = true;
  });
```

**My component**
```
  let reverse = false;
  let isHistoryBack$ = getPageContext().isHistoryBack$;
  $: reverse = $isHistoryBack$;
```
```
<div out:leaveEnter={{ direction: reverse ? 'enter' : 'leave' }} in:leaveEnter={{ direction: reverse ? 'leave' : 'enter' }}>
  ...
</div>
```

**Before fix rendering logs:**
before-fix
before-fix

**After fix rendering logs:**
after-fix
after-fix

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -377,11 +377,17 @@ export function load_component(component: ComponentLoader): Promise<{
 	default: ComponentConstructor,
 	preload?: (input: any) => any
 }> {
-	// TODO this is temporary — once placeholders are
-	// always rewritten, scratch the ternary
-	const promises: Array<Promise<any>> = (typeof component.css === 'string' ? [] : component.css.map(load_css));
-	promises.unshift(component.js());
-	return Promise.all(promises).then(values => values[0]);
+  // TODO this is temporary — once placeholders are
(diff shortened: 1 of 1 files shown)


benmccann:
@monkeythedev if we implemented https://github.com/sveltejs/sapper/issues/1278 would that solve the issue you're trying to address here?

monkeythedev:
@benmccann yes, mine i knda of workaround but I need a deeper knowledge of sapper to do a cleaner fix

benmccann:
Ok, I'll close this in favor of implementing a solution for #1278 then to make sure that we're implementing a clean fix for this issue

PR #1341, opened 2020-07-26
Allow use of history.state

raythurnevoid:
new goto api: `goto("my/page", { foo: "bar" })`


runtime/src/app/goto/index.ts
```diff
@@ -1,10 +1,10 @@
 import { history, select_target, navigate, cid } from '../app';
 
-export default function goto(href: string, opts = { replaceState: false }) {
+export default function goto(href: string, state = {}, opts = { replaceState: false }) {
```
benmccann, line 3: this would be a breaking change, which you could avoid by putting `state` in `opts`

monkeythedev: Yeah i know.. i didn't liked the idea of putting data inside a parameter called "options" and neither adding a third parameter knowing that most of the time it's likely you are gonne pushState and not replace it, resulting in to write `goto("path", null, myState)`.

But since you are right, at the moment we can move it under opts. Maybe in a next release the API could be enhanced.

benmccann: I'm not familiar with what the breaking change policy is. There's a few PRs that are flagged as breaking changes that will be merged when 0.28.0 is released. @antony or @Conduitry might have more details on that. But it would seem weird to me to have an `opts` parameter and not have the other parameters inside it because otherwise why not just have a boolean parameter for `replaceState` directly?

runtime/src/app/goto/index.ts
```diff
@@ -1,13 +1,14 @@
 import { history, select_target, navigate, cid } from '../app';
 
-export default function goto(href: string, opts = { replaceState: false }) {
+export default function goto(href: string, opts = { replaceState: false, state: {} }) {
 	const target = select_target(new URL(href, document.baseURI));
 
 	if (target) {
-		history[opts.replaceState ? 'replaceState' : 'pushState']({ id: cid }, '', href);
+                const state = opts.state || {};
```
benmccann, line 7: these spaces should be tabs for consistency with the existing code

runtime/src/app/goto/index.ts
monkeythedev: I know, i fact i would likely created two different methods, pushState and replaceState as the standard API since i can't realize which other options we will ever see there.

site/content/docs/03-client-api.md
```diff
@@ -24,9 +24,10 @@ sapper.start({
 ```
 
 
-### goto(href, options?)
+### goto(href, state?, options?)
```
benmccann, line 27: it looks like the docs need to be updated to match the recent changes

jimmywarting:
Just came here looking for state management, only wondering how you would than later get the state in your route...

benmccann:
@monkeythedev I haven't heard from you in awhile. Do you still want to pursue this PR?

site/content/docs/03-client-api.md
monkeythedev: Should be ok now, i've merged with new sapper version.

benmccann:
The build appears to be failing

monkeythedev:
> Just came here looking for state management, only wondering how you would than later get the state in your route...

@jimmywarting  `history.state`

> The build appears to be failing

@benmccann Fixed

benmccann:
We did a bit of cleanup and moved all the navigation-related APIs into a single file. It looks like this would need to be rebased to live in the new file: https://github.com/sveltejs/sapper/blob/master/runtime/src/app/router/index.ts

monkeythedev (commented):
master merged in my branch

monkeythedev (commented):
Sorry, i forced this push to transform my merge to a rebase to get a cleaner history.

DboyRedd (approved):

runtime/src/app/goto/index.ts
```diff
@@ -1,13 +1,14 @@
 import { history, select_target, navigate, cid } from '../app';
 
-export default function goto(href: string, state = {}, opts = { replaceState: false }) {
+export default function goto(href: string, opts = { replaceState: false, state: {} }) {
 	const target = select_target(new URL(href, document.baseURI));
 
 	if (target) {
+                const state = opts.state || {};
```
DboyRedd: How

DboyRedd (approved):

benmccann:
Sorry this had gone unreviewed for so long. It looks like there's a merge conflict here. Maybe it's better to try to make sure SvelteKit works as desired than trying to get it in here at this point? I wouldn't want to add new functionality to Sapper that's not in SvelteKit and have folks experience a step backwards when migrating

raythurnevoid:
> Sorry this had gone unreviewed for so long. It looks like there's a merge conflict here. Maybe it's better to try to make sure SvelteKit works as desired than trying to get it in here at this point? I wouldn't want to add new functionality to Sapper that's not in SvelteKit and have folks experience a step backwards when migrating

I agree

benmccann:
Thanks

PR #1344, opened 2020-07-27
TypeScript support: preload detection at runtime

benmccann:
Alternative to https://github.com/sveltejs/sapper/pull/1222

Detects `preload` at runtime instead of compile time

It produces compilation warnings like `'preload' is not exported by 'src/routes/_layout.svelte'` which we will have to silence in the template, but otherwise works and avoids any edge cases that we'd hit by trying to detect whether `preload` is present at compile time


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -154,8 +154,9 @@ export function get_page_handler(
 		let params;
 
 		try {
-			const root_preloaded = manifest.root_preload
-				? manifest.root_preload.call(preload_context, {
+			const root_preload = manifest.root_comp.preload || (() => {});
```
Conduitry, line 157: Either we should not make `root_preload` fallback to a no-op, or we should not have a case in `root_preloaded` for when `root_preload` is falsy.

benmccann: Where do we check that `root_preload` is falsy?

Conduitry: Just below on the next line.

benmccann: I don't think we do? We just use `root_preload` directly without doing any additional check on it

Conduitry: Either I'm going crazy, or the next statement is `const root_preloaded = root_preload ? root_preload.call(...) : {};`.

benmccann: Thanks. I was the one going crazy

I've updated it now

PR #1357, opened 2020-08-04
Support better stack trace with sourcemaps

### LINKED ISSUE #117: Emit and Consume Sourcemaps

habibrosyad:
This PR is a continuity of #773 by adding test and fix found issues.

Closes #117 
Closes #773 

### Tests

### DIFF
--- a/runtime/src/server/middleware/sourcemap_stacktrace.ts
+++ b/runtime/src/server/middleware/sourcemap_stacktrace.ts
@@ -1,0 +1,95 @@
+import fs from 'fs';
+import path from 'path';
+import { SourceMapConsumer, RawSourceMap } from 'source-map';
+
+function retrieveSourceMapURL(contents: string) {
+    const reversed = contents
+        .split('\n')
+        .reverse()
+        .join('\n');
+
+    const match = /\/[/*]#[ \t]+sourceMappingURL=([^\s'"]+?)(?:[ \t]+|$)/gm.exec(reversed);
+    if (match) return match[1];
+
+    return undefined;
+}
+
+const fileCache = new Map<string, string>();
+
+function getFileContents(path: string) {
+    if (fileCache.has(path)) {
+        return fileCache.get(path);
+    }
+    if (fs.existsSync(path)) {
+        try {
+            const data = fs.readFileSync(path, 'utf8');
+            fileCache.set(path, data);
+            return data;
+        } catch {
+            return undefined;
+        }
+    }
+}
+
+function sourcemapStacktrace(stack: string) {
+    const replaceFn = (line: string) =>
+        line.replace(
+            /^ {4}at (?:(.+?)\s+\()?(?:(.+?):(\d+)(?::(\d+))?)\)?/,
+            (input, varName, filePath, line, column) => {
+                if (!filePath) return input;
+
+                const contents = getFileContents(filePath);
+                if (!contents) return input;
+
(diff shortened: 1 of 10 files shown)


runtime/src/server/middleware/sourcemap_stacktrace.ts
```diff
@@ -0,0 +1,97 @@
...
+	return undefined;
+}
+
+const file_cache = new Map<string, string>();
+
+function get_file_contents(path: string) {
+	if (file_cache.has(path)) {
+		return file_cache.get(path);
+	}
+	if (fs.existsSync(path)) {
```
benmccann, line 23: you can remove this `if` check. if the file doesn't exist then we'll hit the `catch` block and return `undefined` just the same

runtime/src/server/middleware/sourcemap_stacktrace.ts
```diff
@@ -0,0 +1,97 @@
...
+
+				const abs_source_path = path.resolve(dir, pos.source);
+				const urlPart = `${abs_source_path}:${pos.line || 0}:${pos.column || 0}`;
+
+				if (!var_name) return `    at ${urlPart}`;
+				return `    at ${var_name} (${urlPart})`;
+			}
+		);
+
+	file_cache.clear();
```
benmccann, line 91: I'm not really sure what the point of the `file_cache` is if it's cleared everytime we try to source map an exception. perhaps we should just remove it?

runtime/src/server/middleware/sourcemap_stacktrace.ts
```diff
@@ -0,0 +1,97 @@
...
+				const pos = consumer.originalPositionFor({
+					line: Number(line),
+					column: Number(column),
+					bias: SourceMapConsumer.LEAST_UPPER_BOUND
+				});
+
+				if (!pos.source) return input;
+
+				const abs_source_path = path.resolve(dir, pos.source);
+				const urlPart = `${abs_source_path}:${pos.line || 0}:${pos.column || 0}`;
```
benmccann, line 84: it'd be more in keeping with the naming convention to make this be `url_part`

habibrosyad: forgot to commit that, i remember that I've made the changes in local 😕

runtime/src/server/middleware/sourcemap_stacktrace.ts
habibrosyad: I noticed that in a stack trace, sometimes you can hit the same file with different lines and cols. The `file_cache` there is just a means to avoid rereading that file multiple times in a single call to `sourcemap_stacktrace`.

runtime/src/server/middleware/sourcemap_stacktrace.ts
habibrosyad: done

runtime/src/server/middleware/sourcemap_stacktrace.ts
habibrosyad: done

benmccann:
@habibrosyad this PR will need to be rebased

habibrosyad:
@benmccann done...

PR #1358, opened 2020-08-04
Removed http coupling from this.fetch

vthukral94:
https://github.com/sveltejs/sapper/pull/1336 (last issue in the list)
`this.fetch` in preload function has a tight coupling with `http`. This creates problem when server is listening via `https` as  `this.fetch` continues to route requests to `http` in this case because of coupling.
### Tests


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sapper",
-  "version": "0.27.16",
+  "version": "0.27.17",
```
benmccann, line 3: leave the version number at `0.27.16`. we will change it when we cut a release

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -109,13 +109,14 @@ export function get_page_handler(
 				preload_error = { statusCode, message };
 			},
 			fetch: (url: string, opts?: any) => {
-				const parsed = new URL.URL(url, `http://127.0.0.1:${process.env.PORT}${req.baseUrl ? req.baseUrl + '/' :''}`);
+				const protocol = req.socket.encrypted ? 'https' : 'http'; // Let's not assume server is listening via http
```
benmccann, line 112: I would remove the comment here. it doesn't add anything that's not obvious from the code

vthukral94:
A functional test for this change will be complicated as * trusted certificates* will be needed to make `https` request. Hence skipping those.

benmccann:
Can you drop the changes to `package-lock.json`? I pushed that change to `master` since it was showing up in all the PRs, so if you rebase against `master` you shouldn't need it anymore

runtime/src/server/middleware/types.ts
```diff
@@ -38,6 +39,10 @@ export type Props = {
 	[key: string]: any;
 };
 
+export interface Socket extends NetSocket {
+	encrypted?: boolean;
```
benmccann, line 43: I'm not sure if this is available on all versions of Node. I can't find it in the Node documentation and StackOverflow suggests it's not present in more recent versions (look at the comments): https://stackoverflow.com/a/10353248

It would be better to get any types that aren't Sapper-specific from DefinitelyTyped or the npm package itself. That would also have the benefit of helping us make sure it's available on the targeted platform

vthukral94: @benmccann This is right that `req.connection` is getting deprecated in recent version of node. But, `req.socket` is available on all the versions of node > v0.3.0.
We are using `req.socket.encrypted` for the test of `https`. This is available of node v14 as well.
There are better ways to get protocol on most recent version of node like `req.protocol`. But, it's not available to earlier versions.
Let me know if you still have any concerns.

benmccann: Ok. Maybe we should install `@types/node` and use the `TLSSocket` type instead of making our own type though?

vthukral94: Let me look into it.

PR #1360, opened 2020-08-06
Upgrade Mocha and Puppeteer

benmccann:


benmccann:
yeah, it's required with the new version and they got rid of `mocha.opts`

PR #1362, opened 2020-08-06
CSP fix for <script> & <style>

vthukral94:
Currently below CSP
```
default-src: 'self' nonce-{token}
script-src: 'self' nonce-{token},
style-src: 'self' nonce-{token}
```
breaks the sapper project because the nonce attribute is not appropriately injected to appropriate resources.
### Tests


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -109,13 +109,14 @@ export function get_page_handler(
 				preload_error = { statusCode, message };
 			},
 			fetch: (url: string, opts?: any) => {
-				const parsed = new URL.URL(url, `http://127.0.0.1:${process.env.PORT}${req.baseUrl ? req.baseUrl + '/' :''}`);
+				const protocol = req.socket.encrypted ? 'https' : 'http';
```
benmccann, line 112: please drop the changes from https://github.com/sveltejs/sapper/pull/1358 from this PR

vthukral94: Yes, I will do that

PR #1363, opened 2020-08-06
Fix typing issues in core files

benmccann:
These changes are necessary to turn on basic type checking. It turns out that despite using TypeScript we don't actually get any of its benefits in terms of safety today.

I'm splitting this PR into multiple because otherwise it's just too big and unwieldy to review. There's a larger set of changes I'll send later fix issues in the `runtime` directory and turn on the checks after working through some final issues. I stole most of these changes from @ajbouh who went much further than this in turning on many additional checks, but I'm trying to do the bare minimum to use TypeScript with all the checks turned off to keep the PR small

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -120,7 +120,7 @@ class Watcher extends EventEmitter {
 		// remove this in a future version
 		const template = read_template(src);
 		if (template.indexOf('%sapper.base%') === -1) {
-			const error = new Error(`As of Sapper v0.10, your template.html file must include %sapper.base% in the <head>`);
+			const error: NodeJS.ErrnoException = new Error(`As of Sapper v0.10, your template.html file must include %sapper.base% in the <head>`);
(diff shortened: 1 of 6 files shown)


src/api/dev.ts
```diff
@@ -120,7 +120,7 @@ class Watcher extends EventEmitter {
 		// remove this in a future version
 		const template = read_template(src);
 		if (template.indexOf('%sapper.base%') === -1) {
-			const error = new Error(`As of Sapper v0.10, your template.html file must include %sapper.base% in the <head>`);
+			const error: NodeJS.ErrnoException = new Error(`As of Sapper v0.10, your template.html file must include %sapper.base% in the <head>`);
```
antony, line 123: At what point do we just remove this error entirely?

It's been a long, long time since 0.10 and we don't have these sorts of protections around other breaking changes.

benmccann: yeah, good point. I've gone ahead and removed it

PR #1366, opened 2020-08-07
Add deploying footnote for sapper build

### LINKED ISSUE #1010: Code generated from `npm run build` is not runnable within that directory. It is missing static directory, package.json, and relative build_dir
# Steps
Run the following commands to clone, install, build, and run:
```
npx degit "sveltejs/sapper-template#rollup" my-app
cd my-app
npm install
npm run build
cd __sapper__/build
node ./index.js
```

# Expected Results
`__sapper__/build` folder is runnable and deployable.

# Actual Results
1. Missing package.json in `__sapper__/build` directory.
2. Missing static folder in `__sapper__/build` directory.
3. build_dir constant in `__sapper__/build/server.js` is relative to root path. #481

# Environment / Dependencies
- macOS Catalina (10.15.1)
- Sapper 0.27.9
- Svelte 3.15.0
- Rollup 1.27.5

# Workaround
1. Create a post-build.js script in the root directory.
2. Add `"postbuild": "node post-build"` to the root package.json
3. Run the following commands:
```
npm run build
cd __sapper__/build
npm i
npm start
```
*post-build.js*
Note: I'm on Node 13.2.0, so using ES modules here. If you're in the same boat, tweak your package.json so the type is module as shown below. Otherwise refactor the `import` statements to `require` statements below and use the global `__dirname` variable.

```"type": "module"```

```
import fs from 'fs'
import path, { dirname } from 'path'
import { fileURLToPath } from 'url'

const __filename = fileURLToPath(import.meta.url)
const __dirname = dirname(__filename)
const BUILD_DIR = path.join(__dirname, '__sapper__/build')

function copyFolderSync(from, to) {
  if (!fs.existsSync(to)) {
    fs.mkdirSync(to)
  }
  fs.readdirSync(from).forEach(element => {
    if (fs.lstatSync(path.join(from, element)).isFile()) {
      fs.copyFileSync(path.join(from, element), path.join(to, element))
    } else {
      copyFolderSync(path.join(from, element), path.join(to, element))
    }
  })
}

function buildPackageJson() {
  const packageJson = JSON.parse(fs.readFileSync(path.join(__dirname, 'package.json')).toString())
  const buildPackageJson = {
    name: packageJson.name,
    dependencies: packageJson.dependencies,
    scripts: {
      start: 'node .'
    }
  }
  fs.writeFileSync(path.join(BUILD_DIR, 'package.json'), JSON.stringify(buildPackageJson, null, 2))
}

function copyStatic() {
  copyFolderSync(path.join(__dirname, 'static'), path.join(BUILD_DIR, 'static'))
}

function hackBuildDir() {
  const serverPath = path.join(BUILD_DIR, 'server/server.js')
  let serverText = fs.readFileSync(serverPath).toString()
  serverText = serverText.replace('__sapper__/build', '.')
  fs.writeFileSync(serverPath, serverText)
}

buildPackageJson()
copyStatic()
hackBuildDir()
```

# Incidentally...
I'm a long time React developer, but really digging Svelte. I recently decided to rewrite my blog which was originally written using create-react-app circa 2 years ago. I experimented with updating the existing code with Context and hooks, or using Gatsby or Next. While I found a few rough edges, the performance metrics in Svelte blew all the others away.

### PR DESCRIPTION
**Pull request by ignatiusmb**
Closes #1010, there's frequent questions too in discord about this topic

Please let me know if there's some wording mistake or if it should be on its own section.

### Tests
- [ ] Run the tests tests with `npm test` or `yarn test`)

### DIFF
--- a/site/content/docs/09-building.md
+++ b/site/content/docs/09-building.md
@@ -14,6 +14,8 @@ The output is a Node app that you can run from the project root:
 node __sapper__/build
 ```
 
+If you're deploying your app outside your machine, make sure to satisfy your `server.js` by installing all your dependencies from `package.json` and to also include your `static` folder.
+
 ### Browser support
 
 Your site is built only for the latest versions of modern evergreen browsers by default. If you are using Rollup, you can use the `--legacy`<sup>1</sup> flag to build a second bundle that can be used to support legacy browsers like Internet Explorer. Sapper will then serve up the correct bundle at runtime<sup>2</sup>.

### CODE REVIEW

File: site/content/docs/09-building.md [hunk 1]
```diff
@@ -14,6 +14,8 @@ The output is a Node app that you can run from the project root:
 node __sapper__/build
 ```
 
+If you're deploying your app outside your machine, make sure to satisfy your `server.js` by installing all your dependencies from `package.json` and to also include your `static` folder.
```
antony, line 17:
```suggestion
If you're deploying your app outside your machine, make sure that you include production dependencies (required by) `server.js` by installing all dependencies from `package.json`:

 ```
   npm --production
 ```
 
   as well as including your `static` directory and it's contents.
```

not sure if we should also include `npm prune`, but I figure that we're not an npm tutorial.

benmccann:
Thanks for adding this. My one suggestion is that I'd expect this content to live under the "Deployment" section instead

ignatiusmb:
Thanks for the wording corrections, please review again if there's still any chance of confusion.

---

@antony I don't know about `npm --production` but according to https://docs.npmjs.com/cli/install, shouldn't it be `npm install --only=prod`?

---

@benmccann I've thought about that too, one of my consideration for putting it directly below `sapper build` is so that people won't miss it, and this is more of a reminder for when using the `build` command, just like `npx serve __sapper__/export` in "Exporting". Should I move it then?

benmccann:
Yeah, it still seems like it belongs in the deployment section to me. It seems like people might go to the deployment section to find what files to copy and they wouldn't know to look in the building section unless they remembered seeing it there earlier. I'm not sure the export example is quite analagous because it just tells you how to run your app, which you might want to do locally, and doesn't really discuss deployment.

habibrosyad:
@ignatiusmb I think what @antony means is `npm install --production`...

ignatiusmb:
Okay, I've converted this PR to a draft for the time being since this will probably be changed again.

I've also added the section in "Deployment" as well. Please let me know which one should be kept.

site/content/docs/11-deploying.md
```diff
@@ -4,6 +4,16 @@ title: Deployment
 
 Sapper apps run anywhere that supports Node 8 or higher.
```
ignatiusmb, line 5:
```suggestion
Sapper apps run anywhere that supports Node 12 or higher.
```

According to https://github.com/sveltejs/sapper-template/issues/206#issuecomment-596485705, Node 8 support has been dropped. Should this be changed to the latest LTS version too?

benmccann: I think saying "Node 10 or higher" here would be most accurate

site/content/docs/09-building.md
```diff
@@ -14,6 +14,14 @@ The output is a Node app that you can run from the project root:
 node __sapper__/build
 ```
 
+If you're deploying your app to a different machine than the one you're building on, make sure that you install your production dependencies (required by `server.js`) by installing all dependencies from `package.json` using `--only=prod` (or `--only=production`):
```
benmccann, line 17: We should remove this so as not to duplicate content from the deploying section

ignatiusmb: Thanks for the reminder, I was still unsure back then but I think the wording in `11-deployment` states it more clearly, will remove this then. But before that, is it better if it was replaced with something like this?
> See [deployment section (link)] for instructions on deploying to a different machine

benmccann: No, I don't think it's necessary and would leave it out

ignatiusmb: Okay then, thanks for the review

site/content/docs/11-deploying.md
```diff
@@ -2,7 +2,17 @@
 title: Deployment
 ---
 
-Sapper apps run anywhere that supports Node 8 or higher.
+Sapper apps run anywhere that supports Node 10 or higher.
+
+### Deploying from sapper build
+
+If you're deploying your app to a different machine than the one you're building on, you will need to copy your `package.json` over and install your production dependencies (required by `server.js`) using `--only=prod` (or `--only=production`):
```
benmccann, line 9: I think it's a bit prescriptive to say that the installation has to happen on the production machine. Also, it's a bit redundant to include the flags here since they're shown below. What about something like?

```suggestion
You will need to have the server's dependencies available on your production machine. You can either install the dependencies on a machine with an empty `node_module/` folder and copy that folder over or you can make the `package.json` available on the production machine and install the dependencies there. Production dependencies can be installed with:
```

site/content/docs/11-deploying.md
```diff
@@ -2,7 +2,17 @@
...
+
+```bash
+npm install --only=prod
+```
+
+Other than `__sapper__/build` directory and `package.json`, you will also need to include your `static` directory and its contents.
```
benmccann, line 15: I don't think you actually need the `package.json` to be deployed to the prod machine

```suggestion
You will also need to run `npm run build` and then make the `__sapper__/build` and `static` directories available on the production machine (either by building it there or by copying over the resulting files).
```

ignatiusmb: I agree, it's the same case as the first statement where it can be both copied or built directly. I think we should also say how to run the app (like building section) to replace `npm start` that `package.json` included, something like

> You can then start your app by running `node __sapper__/build`

Aside from that, perhaps it's better to just make sure `__sapper__/build` is included at the start of this section since `npm run build` has been mentioned in building section.

site/content/docs/11-deploying.md
benmccann: If the folder isn't empty to start then I imagine that you could have dev dependencies in there. I haven't actually tested to see if you run with the production flag whether it will remove those, but I'm assuming it'd leave them there. And yes, I accidentally left off an `s`, which should be included

site/content/docs/11-deploying.md
```diff
@@ -2,7 +2,17 @@
...
 
-Sapper apps run anywhere that supports Node 8 or higher.
+Sapper apps run anywhere that supports Node 10 or higher.
+
+### Deploying from sapper build
+
+If you're deploying your app to a different machine than the one you're building on, you will need to copy your `package.json` over and install your production dependencies (required by `server.js`) using `--only=prod` (or `--only=production`):
+
+```bash
+npm install --only=prod
```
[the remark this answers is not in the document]
benmccann: Yes, you should copy over your `package.json` and `package-lock.json`. I would recommend that even with `npm install`

site/content/docs/11-deploying.md
ignatiusmb: I'm also not sure if the production flag for `npm i` would remove dev packages or not, but I guess this is where `npm ci --prod` would be useful because it automatically removes preexisting `node_modules` folder before it begins its install.

benmccann: Ah, that's good to know. I guess in that case with the `npm ci --prod` command we could drop the part about starting with an empty folder

site/content/docs/11-deploying.md
benmccann: > You can then start your app by running `node __sapper__/build`

That makes sense to me

> Aside from that, perhaps it's better to just make sure __sapper__/build is included at the start of this section since npm run build has been mentioned in building section.

I think it's okay to mention that `npm run build` should be run. I'd rather be more explicit than leave out a step

site/content/docs/11-deploying.md
ignatiusmb: Okay, I've added the changes here with a slight tweak. Resolving this in favor of the updated changes made

site/content/docs/11-deploying.md
```diff
@@ -2,7 +2,21 @@
...
+
+### Deploying from sapper build
+
+We will need to have `server.js` dependencies available on the production machine. You can either install the dependencies locally or have the `package.json` and `package-lock.json` available on the production machine and install the dependencies there. Production dependencies can be installed with:
+
+```bash
+npm ci --prod
+```
+
+Make sure to run `npm run build` and have `__sapper__/build` directory generated before proceeding. We will also need to include the `static` directory in the production machine, you can then start your app with:
```
ignatiusmb, line 15: > before proceeding.

I'm still not 100% sure about this, is this clear enough to imply that `npm run build` can be installed locally or in the production, just like `npm ci` command, or should it be stated again with something like

```suggestion
Make sure to run `npm run build` and have `__sapper__/build` directory generated, just like with the production dependencies, this can happen locally or on the production machine. We will also need to include the `static` directory in the production machine, you can then start your app with:
```

site/content/docs/11-deploying.md
```diff
@@ -2,7 +2,21 @@
 title: Deployment
 ---
 
-Sapper apps run anywhere that supports Node 8 or higher.
+Sapper apps run anywhere that supports Node 10 or higher.
+
+### Deploying from sapper build
+
+We will need to have `server.js` dependencies available on the production machine. You can either install the dependencies locally or have the `package.json` and `package-lock.json` available on the production machine and install the dependencies there. Production dependencies can be installed with:
```
benmccann, line 9: The text switches back and forth between "we" and "you", which reads funny. Let's stick with one or the other for consistency. It seems like the rest of the docs you "you" more than "we"

```suggestion
You will need to have `server.js` dependencies available on the production machine. You can either install the dependencies locally or have the `package.json` and `package-lock.json` available on the production machine and install the dependencies there. Production dependencies can be installed with:
```

site/content/docs/11-deploying.md
```diff
@@ -2,7 +2,21 @@
 title: Deployment
 ---
 
-Sapper apps run anywhere that supports Node 8 or higher.
+Sapper apps run anywhere that supports Node 10 or higher.
+
+### Deploying from sapper build
```
benmccann, line 7: I feel like it's gotten a bit muddled with interleaving stuff about building on one machine vs another. What if we just cover building and running and then talk about possibly transferring the files to another machine.

### Deploying the build output

You can install the production dependencies and build Sapper with:

```bash
npm ci --prod
npm run build
```

You can now run Sapper with:

```bash
node __sapper__/build
```

If you'd like to build on one machine and run on another, you can copy the `__sapper__/build`, `static`, and `node_modules` directories. For example, you could zip those directories and then transfer the zip to another machine.

```bash
zip -r build.zip static __sapper__/build node_modules
```

Conduitry:
I'm still worried about giving people step-by-step instructions that attempt to be generic. I'd probably prefer something that just indicates "you need your Node dependencies, `__sapper__/build`, and `static`".

benmccann:
I think having something about only needing production Node dependencies and not all dependencies is useful. It makes builds faster and smaller and is easy to overlook.

We could certainly drop my last line about the example zip if you prefer. I think it works across all platforms, so thought it would be a good example, but obviously people are going to have their own preferences there

Conduitry:
Yeah, telling people to only install production dependencies is a good idea.

There's no `zip` on Windows, and it's not there by default on my Ubuntu installation, so that's probably not something we want to direct people towards using.

benmccann:
Makes sense. I edited my comment above. Let me know if you think it makes more sense now

site/content/docs/11-deploying.md
ignatiusmb: Thanks for the summary, I took a page out of this but decided to keep the title as is, since
1. Deploying the build output implies that user has already built the output beforehand
2. It has been covered in building section and assuming from the recent comments, we're trying to focus on how to run the app
3. Running `npm run build` after `npm ci --prod` would fail and exit with an error because preexisting `node_modules` are removed and sapper module as `devDependencies` would not exist

I've removed the build command and reordered the sentences a bit, please let me know again if there's more improvements to be made.

ignatiusmb:
I've removed the previously mentioned instructions and only kept the install command as a quick note. Thanks @Conduitry for the clarification and @benmccann for the summarized version. Please let me know if there's more to improve.

Issue #1328, opened 2020-07-18
Problem with Service Worker on Chromium.


ljonya:
I created a [template sapper project](https://github.com/sveltejs/sapper-template), updated and installed packages (`ncu -u && npm i`). Then I added any 40-50mb images to the `static` directory. This causes the site to hang after updating.  `DevTools -> Application -> Service Workers` says `Status: trying to install`.
![work problem](https://user-images.githubusercontent.com/12491743/87854837-3c3bad80-c91d-11ea-9a31-6cd4bb820a16.png)

An example project with a problem (there are several images from Wikipedia in the static directory): https://github.com/ljonya/sapper-app

The problem appears at first run (`npm run dev`), sometimes you need to update something in `.svelte` files, sometimes you need `Unregister` already installed Service Worker in DevTools. 

**Logs**
Logs are empty.

**To Reproduce**
1. Run the example: https://github.com/ljonya/sapper-app
2. In `DevTools -> Application -> Service Workers`: `Unregister` already installed Service Worker.
3. Refresh the page in Chromium and possibly update, for example, `index.svelte`. 

**Information about Sapper Installation:**
- Chromium 84.0.4147.89 (Official build), snap (64 bit)
- Kubuntu 20.04 64 bit
- Sapper version 0.27.16
- Svelte version 3.24.0
- Application uses Rollup

antony:
The service worker in Sapper reads the static dir and caches the contents. This means that the static directory is great for small assets and things that should always be cached, but makes it unsuitable for storing all your sites assets, and you should probably consider customising the service worker or using a CDN.

50mb is very large for a service worker, and I don't doubt that it probably jams your browser. It definitely won't work in Safari https://love2dev.com/blog/what-is-the-service-worker-cache-storage-limit/

So I would suggest actions from here are:

1. Use a CDN for your non-critical assets.
2. Customise your service worker so that it loads assets selectively
3. We might need to better document what the service worker does so that this sort of thing can be avoided.

qwabra:
https://github.com/sveltejs/sapper-template/blob/master/src/service-worker.js#L6-L7

`src/service-worker.js`

exclude `.jpg` files

```js
/** @type {string[]} */
let filesFiltered = files
filesFiltered = files.filter(str => !str.endsWith('.jpg'))
// filesFiltered = files.filter(str => !str.startsWith('images'))

// `shell` is an array of all the files generated by the bundler,
// `files` is an array of everything in the `static` directory
[...]

PR #1367, opened 2020-08-07
clarify default service worker behaviour

antony:
Fixes https://github.com/sveltejs/sapper/issues/1328


site/content/docs/01-structure.md
```diff
@@ -106,6 +106,7 @@ This is a place to put any files that your app uses — fonts, images and so on.
 
 Sapper doesn't serve these files — you'd typically use [sirv](https://github.com/lukeed/sirv) or [serve-static](https://github.com/expressjs/serve-static) for that — but it will read the contents of the `static` folder so that you can easily generate a cache manifest for offline support (see [service-worker.js](docs#src_service-worker_js)).
 
+> Note that the default behaviour of the service worker is to cache all assets from the static directory, so if you have a large number (&gt; 50mb) of files here, you should customise what it caches, otherwise some browsers won't be able to load the service worker.
```
antony, line 109:
```suggestion
> Note that the default behaviour of the service worker is to cache all assets from the static directory, so if you have more than 50mb of files here, you will start to exceed the cache limit for service-workers in some browsers, which can cause the service worker to stop loading. In this instance, it is advisable to customise what files are cached by editing the service-worker yourself.
```

PR #1372, opened 2020-08-08
Don’t prevent the browser’s default behavior corresponding to the Alt key

yuheiy:
In most browsers, clicking links with the Alt key has a special behavior, for example, Chrome downloads the target resource. As with other modifier keys, the router should stop the original navigation to avoid preventing the browser’s default behavior.

When users click a link while holding the Alt key together, the browsers behave as follows.

Windows 10:

| Browser    | Behavior                                    |
|:-----------|:--------------------------------------------|
| Chrome 84  | Download the target resource                |
| Firefox 79 | Prevent navigation and therefore do nothing |
| Edge 84    | Download the target resource                |
| IE 11      | No impact                                   |

macOS Catalina:

| Browser    | Behavior                                    |
|:-----------|:--------------------------------------------|
| Chrome 84  | Download the target resource                |
| Firefox 79 | Prevent navigation and therefore do nothing |
| Safari 13  | Download the target resource                |

### Tests

### DIFF
--- a/runtime/src/app/start/index.ts
+++ b/runtime/src/app/start/index.ts
@@ -77,7 +77,7 @@ function handle_click(event: MouseEvent) {
 	// Adapted from https://github.com/visionmedia/page.js
 	// MIT license https://github.com/visionmedia/page.js#license
 	if (which(event) !== 1) return;
-	if (event.metaKey || event.ctrlKey || event.shiftKey) return;
+	if (event.metaKey || event.ctrlKey || event.shiftKey || event.altKey) return;
 	if (event.defaultPrevented) return;
 
 	const a: HTMLAnchorElement | SVGAElement = <HTMLAnchorElement | SVGAElement>find_anchor(<Node>event.target);


benmccann:
It might be a good idea to submit this PR to page.js as well: https://github.com/visionmedia/page.js/blob/master/page.js#L748

* they're more familiar with this block of code and it would give confidence that this is the correct fix
* it would help us keep this block of code in sync with the upstream version

yuheiy:
I’ve already submitted. https://github.com/visionmedia/page.js/pull/568

Conduitry:
Does the "No impact" for IE 11 mean that typically alt-clicking a link performs regular navigation, and that with this change we'd no longer be intercepting these clicks and it would perform a regular page navigation instead of an SPA-type navigation?

Even if so, I think it's probably overall a good idea to make this change - I just want to be clear on what its effect is.

yuheiy:
That’s right. In IE 11, alt-clicking has the same behavior as when the user is not using alt.

yuheiy:
So IE 11 performs a regular page navigation.

Issue #384, opened 2018-08-29
Mixed Content error when loading /__sapper__ when accessing dev server with https


btakita:
I'm developing an app that uses Auth0. Auth0 requires https to work, even for development.

However, sapper hot module reloading does not work on Chrome & Firefox due to the mixed content error.

It seems like web sockets over a different port are being used. I'm willing to configure the url to access the hot reload service, which I can proxy through nginx or something similar.

thgh:
Did you ever find a solution for this?

Seb35:
I have a first proposition to deal with this issue, although it must be discussed. My working (but hacky) proposition is:

1. to reverse-proxy the connection from HTTPS port 10001 to HTTP port 10000 (e.g. with nginx)
2. to modify sapper/sapper-dev-client.js:
```diff
-        source = new EventSource(`http://${window.location.hostname}:${port}/__sapper__`);
+        source = new EventSource(`http://${window.location.hostname}:${port+1}/__sapper__`);
```

I aggree the `port+1` is arbitrary and not a generic solution. More robust solutions include:

* either add another parameter for the public-visible dev-port (and people can configure HTTPS through the proxy), with a default to the dev-port,
* either manage the https connection inside Sapper (but this is *a lot* of new parameters: public certificate, private key, https versions, ciphersuite preferences, etc.)

thelittlebug:
adding https to sapper is imho no good idea. the proxy does that job better.

but +1 for `either add another parameter for the public-visible dev-port (and people can configure HTTPS through the proxy), with a default to the dev-port`. that would be really helpfull

and the protocoll could be `$(window.location.protocol)`

bdougherty:
I'm wondering if `/__sapper__` needs to be on a separate port at all. It would eliminate the need for any extra configuration if it was served from the main server.

Conduitry:
I had assumed we were using some third-party livereload library that needed to run its own server (on its own port), but as far as I can tell this is not the case.

michaek:
I think there's a good argument for keeping the dev server out of the main server - you can be sure that there's not any unexpected entanglement between the dev and production servers. It seems there's already a CLI flag for `--dev-port` on `sapper dev`, but (unfortunately?) it does affect both the server port binding and the port used in the client.

[...]

PR #1379, opened 2020-08-09
Dynamically set the protocol to connect to the dev server

GaretJax:
This PR relates to https://github.com/sveltejs/sapper/issues/384. It does not address the broader issue of a full https setup, as it may be out of scope or the design warrants further discussion before the implementation can be started, but makes it at least possible to access the dev server when loading the app through an https proxy, by avoiding requests being blocked due to mixed content.

### DIFF
--- a/sapper-dev-client.js
+++ b/sapper-dev-client.js
@@ -13,7 +13,7 @@ function check() {
 export function connect(port) {
 	if (source || !window.EventSource) return;
 
-	source = new EventSource(`http://${window.location.hostname}:${port}/__sapper__`);
+	source = new EventSource(`${window.location.protocol}://${window.location.hostname}:${port}/__sapper__`);
 
 	window.source = source;
 


Conduitry:
This breaks live reloading, as `location.protocol` appears to already include the colon.

benmccann:
Ah, I guess I should have pulled it down and tested it locally. I just pushed a change to remove the extra colon

njbotkin:
I'm having trouble seeing how this is supposed to work.  If I'm running sapper at https://*:3000, but livereload is still http://*:10000, this tries to talk to talk to livereload at https://*:10000 and livereload stops working.  Am I missing a step where the livereload server is made to use https?

GaretJax:
> I'm having trouble seeing how this is supposed to work. If I'm running sapper at https://*:3000, but livereload is still http://*:10000, this tries to talk to talk to livereload at https://*:10000 and livereload stops working. Am I missing a step where the livereload server is made to use https?

You would not be able to connect to livereload on HTTP from a page served over HTTPS anyway. This patch only makes sapper connect to the same protocol as it has been loaded from, but does not take care of actually exposing any server over HTTPS.

njbotkin:
You can mix http and https, you just have you allow the site to load insecure content.  What's the workaround, now?

PR #1386, opened 2020-08-10
Turn on basic type checking

benmccann:
sucrase doesn't actually do any type checking. This switches to using the TypeScript plugin so that we get basic error checking. We could turn on additional checks over time, but I'm really just trying to get some basics in place right now

Thanks to @ajbouh for leading the way here. Many of these changes are backported from his fork of Sapper: https://github.com/metered/sapper


pngwn:
That is the reason sucrase was chosen. Because type checking in dev makes builds unfathomably slow, so we just strip types. Your IDE gives most of the value. 

If we want to type check one final time we should do it in CI as a last quality check, like we do with linting.

benmccann:
Ok. I changed it so that it doesn't type check by default, but does when you run `npm test`, so that it will be run on the CI.

On my (very old) machine it takes 2s to build without type checking and 4.8s to build with type checking, so it is more than twice as slow, though it's still only a couple seconds. The tests already take 40s, so adding 2s more shouldn't make a big difference there.

I think part of the reason I didn't notice is that I don't build the code all that much. I mainly do it when I'm done with my changes want to check the code is correct and either manually test or run the tests where it's a rounding error.

runtime/src/node_modules/@sapper/internal/manifest-server.d.ts
```diff
@@ -0,0 +1,29 @@
...
+  ignore: RegExp[];
+  root_comp?: {
+    default: SSRComponent<unknown>;
+    preload?: Preload<unknown>;
+  }
+  error: SSRComponent
+  pages: Route[]
+};
+
+export type ServerRoute = {
```
dummdidumm, line 17: Some of these are interfaces, some of these are types (not only in this files, others, too). Most of them feel like they could be interfaces. I would try to do it in a consistent manner.

runtime/src/node_modules/@sapper/internal/manifest-server.d.ts
```diff
@@ -0,0 +1,29 @@
...
+  pattern: RegExp;
+  handlers: Record<string, Handler<Req, Res>>;
+  params: (match: RegExpMatchArray) => Record<string, string>;
+};
+
+export interface SSRComponent<Props> {
```
dummdidumm, line 23: Not sure about this semantic-wise, but shouldn't this extend `Component`?

runtime/src/app/types.ts
```diff
@@ -8,21 +8,11 @@ type Child = {
 	component?: Component;
 };
 
-export interface ComponentConstructor {
-	new (options: { target: Node, props: any, hydrate: boolean }): Component;
-	preload: (props: { params: Params, query: Query }) => Promise<any>;
-}
-
 export interface Component {
 	$set: (data: any) => void;
```
dummdidumm, line 12: `$set` is for Props, so maybe you could change this to

```ts
export interface Component<Props = any> {
	$set: (data: Partial<Props>) => void;
	$destroy: () => void;
}
```

runtime/src/node_modules/@sapper/internal/shared.d.ts
```diff
@@ -0,0 +1,11 @@
+export const CONTEXT_KEY: unknown
+
+export interface ComponentModule {
+  default: ComponentConstructor
+  preload: (props: { params: Params, query: Query }) => Promise<any>;
+}
+
+export interface ComponentConstructor {
+  new(options: { target: unknown, props: unknown, hydrate: boolean }): Component;
```
dummdidumm, line 9: If props are the props of the Svelte component, and if you do the proposed change for `Component` in my other comment, then you could change this to

```ts
export interface ComponentConstructor<Props = any> {
  new(options: { target: unknown, props: Props, hydrate: boolean }): Component<Props>;
  preload: (props: { params: Params, query: Query }) => Promise<any>;
}
```

runtime/src/node_modules/@sapper/internal/App.svelte.d.ts
```diff
@@ -0,0 +1,8 @@
+import {
+  ComponentConstructor,
+  SSRComponent
+} from './shared';
+
+declare const App: ComponentConstructor & SSRComponent
```
dummdidumm, line 6: This throws type errors for me when trying it in the TS playground, because the generic for `SSRComponent` is missing.

To get the generics in there and have it types as a class at the same time without writing everything twice, you can do class/interface merging:

```ts
declare interface App<Props> extends ComponentConstructor, SSRComponent<Props> {}
declare class App<Props> {}
```

runtime/src/node_modules/@sapper/internal/manifest-server.d.ts
ajbouh: Ah, that means Component is too general sounding. I think Component should really be DOMComponent, I just didn't realize that the type would split into two when I was first choosing some of the names.

runtime/src/node_modules/@sapper/internal/manifest-server.d.ts
ajbouh: Yes, good call.

benmccann: done. thanks!

runtime/src/node_modules/@sapper/internal/manifest-server.d.ts
benmccann: I combined the client and server components into a single `Component`

runtime/src/app/types.ts
benmccann: I no longer have a `Props` type

runtime/src/node_modules/@sapper/internal/shared.d.ts
benmccann: I'm not quite sure I understand the suggestion to make this a class

runtime/src/node_modules/@sapper/internal/App.svelte.d.ts
benmccann: I just installed VS Code and it led me down a rabbit hole of redoing most of the types. It showed a bunch of errors that the command line build wasn't. Is VS Code doing something different and more powerful than the rollup plugin does?

runtime/src/node_modules/@sapper/internal/manifest-server.d.ts
benmccann: I renamed `Component` to `DOMComponent` since it clarified things a bit to me

runtime/src/node_modules/@sapper/internal/App.svelte.d.ts
benmccann: Ok, I managed to take the files out of `node_modules`, which is making things work better for me

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -39,13 +38,14 @@ export function get_page_handler(
 	function handle_error(req: Req, res: Res, statusCode: number, error: Error | string) {
 		handle_page({
 			pattern: null,
+			resources: [],
```
ajbouh, line 41: This is only used by my rollup plugin, probably shouldn't be here

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -260,9 +260,15 @@ export function get_page_handler(
 				level0: {
 					props: preloaded[0]
 				},
-				level1: {
-					segment: segments[0],
-					props: {}
+				level1: error ? {
```
ajbouh, line 263: I don't know if this should be here or not, as it assumes my changes/simplificates to create_app are present (which they don't seem to be).

runtime/src/internal/manifest-server.d.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export interface ManifestPagePart {
+  name: string | null;
+  file?: string;
+  component: ComponentModule;
+  params?: (match: RegExpMatchArray | null) => Record<string, string>;
+}
+
+export interface ManifestResource {
```
ajbouh, line 34: This is only used by my rollup plugin, probably shouldn't be here

runtime/src/internal/manifest-server.d.ts
```diff
@@ -0,0 +1,65 @@
...
+  server_routes: ServerRoute[];
+  ignore: RegExp[];
+  root_comp: ComponentModule
+  error: SSRComponent
+  pages: ManifestPage[]
+}
+
+export interface ManifestPage {
+  pattern: RegExp | null;
+  resources: ManifestResource[]; // list of assets needed for transitive dependencies
```
ajbouh, line 23: This is only used by my rollup plugin, probably shouldn't be here

runtime/src/internal/shared.d.ts
```diff
@@ -0,0 +1,27 @@
...
+  preload?: Preload;
+}
+
+export interface DOMComponent {
+	$set: (data: any) => void;
+	$destroy: () => void;
+}
+
+export interface ComponentConstructor {
+  new(options: { target: unknown, props: unknown, hydrate: boolean }): DOMComponent;
```
ajbouh, line 18: The actual type for target is Node, which requires the typescript dom library. Unfortunately anything that's in this file needs to work on both node and in the browser.

runtime/src/internal/shared.d.ts
```diff
@@ -0,0 +1,27 @@
+export const CONTEXT_KEY: unknown;
+
+export type Params = Record<string, string>;
+export type Query = Record<string, string | string[]>;
+export type Preload = (props: { params: Params, query: Query }) => Promise<any>;
+
+export interface ComponentModule {
+  default: ComponentConstructor | SSRComponent
```
ajbouh, line 8: Since the ComponentConstructor requires the dom library (which refers to browser-only types), I don't think we can easily have types that are either SSR or DOM. So at some level it's only "ok" to be in this file because the types aren't as accurate as they should be...

runtime/src/server/middleware/get_page_handler.ts
benmccann: done

runtime/src/server/middleware/get_page_handler.ts
benmccann: done

runtime/src/internal/manifest-server.d.ts
benmccann: done

runtime/src/internal/manifest-server.d.ts
benmccann: done

runtime/src/internal/shared.d.ts
benmccann: done

runtime/src/internal/shared.d.ts
benmccann: done

runtime/src/app/app.ts
```diff
@@ -185,11 +194,17 @@ export async function navigate(target: Target, id: number, noscroll?: boolean, h
...
 
-	await render(redirect, branch, props, target.page);
-	if (document.activeElement) document.activeElement.blur();
+	if (redirect) {
+		await goto(redirect.location, { replaceState: true });
+	} else {
+		const { props, branch } = loaded_result;
+		await render(branch, props, target.page);
+	}
+	if (document.activeElement && (document.activeElement instanceof HTMLElement)) document.activeElement.blur();
```
ajbouh, line 207: Might make sense to highlight that this is a fix in the commit message

runtime/src/internal/manifest-client.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  default: ComponentConstructor;
+  preload?: Preload;
+}
+
+export interface DOMComponent {
+	$set: (data: any) => void;
+	$destroy: () => void;
+}
+
+export interface ComponentConstructor {
```
ajbouh, line 17: This should really be DOMComponentConstructor

(but I can do this in a follup up comment)

runtime/src/internal/manifest-client.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  new(options: { target: Element, props: unknown, hydrate: boolean }): DOMComponent;
+}
+
+export interface ComponentLoader {
```
ajbouh, line 21: This should really be DOMComponentLoader

(but I can do this in a follup up comment)

runtime/src/server/middleware/get_server_route_handler.ts
```diff
@@ -18,7 +18,7 @@ export function get_server_route_handler(routes: ServerRoute[]) {
 				// intercept data so that it can be exported
 				res.write = function(chunk: any) {
 					chunks.push(Buffer.from(chunk));
-					write.apply(res, [chunk]);
+					return write.apply(res, [chunk]);
```
ajbouh, line 21: This is technically a fix

runtime/src/internal/manifest-client.d.ts
benmccann: done

runtime/src/internal/manifest-client.d.ts
benmccann: done

runtime/src/app/app.ts
benmccann: this is pretty minor. I'll end up squashing the commits when I merge, so I don't think there's a need. if this were bigger I'd send it as a separate PR

ajbouh: Ok, I don't know precedent for this project. I think helping people comb through commits that might fix their bugs in prod can make them more eager to update ;)

Just a nit on my part

runtime/src/app/app.ts
```diff
@@ -364,8 +373,8 @@ function load_css(chunk: string) {
 	});
 }
 
-export function load_component(component: ComponentLoader): Promise<{
-	default: ComponentConstructor,
+export function load_component(component: DOMComponentLoader): Promise<{
```
ajbouh, line 376: This seems like it's actually returning `Promise<DOMComponentModule>`, right?

Can do this in a follow up

benmccann: updated

runtime/src/app/types.ts
dummdidumm: I meant a `Props` generic, because the `$set` method expects tlan object of input properties of the component, you might want to type this in a generic way.

benmccann: Ok. I think I will save tightening up the types for a follow up (which I think @ajbouh was interested in taking a stab at) and maybe you can lend a set of eyes to that. There's a whole bunch of places that would need to be changed if I'm going to make use of it in a meaningful way. Right now I'm mostly just trying to do the minimal to get the build configuration in and am afraid this PR has already grown pretty large

dummdidumm: Good call 👍

benmccann:
@dummdidumm was there anything else you wanted to take a look at for this one or do you think it's okay to merge?

benmccann:
Thanks @dummdidumm and @ajbouh for taking a look at this one!

PR #1389, opened 2020-08-11
Remove unused dependencies. Upgrade others

benmccann:
This fixes the 151 known security vulnerabilities in `master`

Removed `agadoo` because it's unused
Removed `@types/puppeteer` because the types are now included in the puppeteer package


Conduitry:
These test failures seem to be legitimate - I'm seeing them locally as well - but I haven't looked into them at all.

benmccann:
oops. fixed

PR #1394, opened 2020-08-11
Enable eslint

benmccann:
This ended up being much easier than I remember


src/core/create_manifest_data.ts
```diff
@@ -77,7 +77,7 @@ export default function create_manifest_data(cwd: string, extensions: string = '
 				const route_suffix = basename.slice(basename.indexOf('.'), -ext.length);
 
 				parts.forEach(part => {
-					if (part.qualifier && /[\(\)\?\:]/.test(part.qualifier.slice(1, -1))) {
+					if (part.qualifier && /[()?:]/.test(part.qualifier.slice(1, -1))) {
```
antony, line 80: Are we certain this does the same thing? I'm always wary with the regex fixes.

benmccann: Yeah, I manually tested `/[(]/.test('ab(cd');` and `/[\(]/.test('ab(cd');` in jsconsole and they both returned the same thing. 

You shouldn't need to escape them in a character class. It'd be a bad eslint rule if it were wrong :smile: Though software does have bugs sometimes so good to check!

PR #1395, opened 2020-08-12
CSP Violation Fixes for script-src "Webpack"

vthukral94:
`'nonce-{nonce}'` in CSP header for script-src prevents execution of main file because of the absence of nonce attribute.
### Tests


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -315,7 +318,7 @@ export function get_page_handler(
 					script += `var s=document.createElement("script");try{new Function("if(0)import('')")();s.src="${main}";s.type="module";s.crossOrigin="use-credentials";}catch(e){s.src="${req.baseUrl}/client/shimport@${build_info.shimport}.js";s.setAttribute("data-main","${main}")}document.head.appendChild(s)`;
 				}
 			} else {
-				script += `</script><script src="${main}" defer>`;
+				script += `</script><script${nonce_attr} src="${main}" defer>`;
```
benmccann, line 321: this doesn't need a space between `script` and `${nonce_attr}`?

benmccann:
This is somewhat duplicate of https://github.com/sveltejs/sapper/pull/1232

runtime/src/server/middleware/get_page_handler.ts
antony: it doesn't, since it's done above. However the fact you tripped over this is exactly why I don't do this, and it's super ugly too.

The most readable + still terse way to build this would be to create a map of all required attributes and build the tag from that, something like:
```js
[
  '<script',
  'foo="bar"',
  hasQux ? 'baz="qux"' : [],
  'defer>'
].join(' ')
```

for a lot of attributes I'd then probably use a mapper to build the attribute list too.

vthukral94: @antony I agree this is not a really good way to inject nonce attribute. I didn't want to diverge too much from the current implementation, hence didn't change the way it's being injected. Through this PR I am trying to fix the missing attribute issue,

vthukral94: @benmccann Even after these changes there are still CSP violations as `client/**/main.js` tries to inject styles inside components via javascript without nonce. Is there any reason why aren't we using `emit-css: true` in `svelte-loader` for webpack? This will help avoiding such violations & at the same time improve performance of app.  
I will try to include that as well as part of this PR.

benmccann: I'm really not that familiar with the webpack setup. All the maintainers use Rollup, so it's generally in a better state and [has more features](https://sapper.svelte.dev/faq/)

vthukral94: got it. 
I will switch to roll up `as well` in a week or 2. But, in the mean time I ll try to fix the issue with webpack as well.

vthukral94:
@benmccann I have reduced the scope of this PR to just deal with `srcipt-src`. Fixing `style-src` is gonna take a lot of time at my end which I currently don't have.

benmccann:
This looks to me like it's going to generate `<script${nonce_attr}></script><script${nonce_attr} src="${main}" defer></script>`. It seems pretty hacky to have that empty script tag in there. Most of that is from before your PR, but if we're going to be touching this area I think it would be nice to do that small cleanup rather than just putting more code on top of it

vthukral94:
@benmccann I don't think an empty script tag will be added. I verified it as well. Attaching the screenshot below.
Screen Shot 2020-08-14 at 11 32 52 AM

vthukral94:
Line 296 can give you more context as well https://github.com/sveltejs/sapper/pull/1395/files#diff-9c713b090053bf2158dc7db4484fb401R296

PR #1105, opened 2020-02-28
feat: add ability to follow links in img and source elements

aloker:
This PR supersedes #938. It implements the following new behavior:

- In addition to a-elements, `export` now also follows links found in ``, ` and `<source srcset>` and exports them if they point to server routes. 

- As a side effect, links to other server routes emitting binary files also properly export the binary files. Before, binary files were not correctly exported.

Fixes #1103
Fixes #1104

PR #1398, opened 2020-08-12
in exporting, handle binary files from server routes

### LINKED ISSUE #1103: Exported server routes returning binary files get corrupted
**Describe the bug**
When a route points to a server route that emits a binary file (pdf, image etc.) and the site is `export`ed, the files on disk get corrupted because export only supports text results from server routes.

**To Reproduce**
- Create a server route that returns a binary file (image, PDF)
- Link to the server route inside a Svelte component
- Export the site
- Find the exported file of the server route on disk

Result: the file is corrupt and cannot be opened.

**Expected behavior**
The result of server routes emitting binary data should be properly stored on disk when exporting.

- Sapper master branch (v0.27.10)
- Exported site
- Checked with rollup, but should be same for webpack

**Severity**
Blocking my attempt to load all data including binary assets (documents, images) from a headless CMS.

### LINKED ISSUE #1104: Sapper does not crawl img and source elements
Sapper only follows links in `<a>` elements. 

Sapper should crawl links in the `src` and `srcset` attributes in `` and `<source>` tags and export them to disk if they refer to server routes.

Manually copying the related files to the static directory, but that is not feasible when using Sapper for a headless CMS.

I'm developing a prototype frontend for a headless CMS and this is a huge problem. My server routes load the images from the CMS and transform them using [sharp](https://www.npmjs.com/package/sharp) (resizing, type conversion etc for responsive images). Without this feature I cannot use the automatic conversion and implement an unrelated build step for this.

**Additional context**
PR #938 already proses a partial solution but I have a PR in the available that adds support for srcset attributes.

Conduitry:
Resolves #1103, resolves #1104. Builds on work in and supersedes #938 and #1105.

This is basically #1105, with conflicts with master resolved, code style adjusted, and with another case in `get_server_route_handler` of dangerous buffer-to-string conversion removed.

This lets you follow ``, `<source src>`, and `<source srcset>` attributes when exporting, and to also correctly handle (without encoding corruption) when these are server routes that return binary files.


HighestTide:
This currently results in kinda-expected, but kinda-annoying results if you have endpoints that return images without the route having the correct extension. Would be it be worth considering appending the "correct" extension for image files?

It is pretty simple to work around this by just adding the extension manually on the markup side and removing it in the route handler, but it seems like a simple fix could cover a lot of use cases and prevent confusion with exports not matching the running application. I'm not sure what the Sapper philosophy is when it comes to manipulating stuff like that, though.

Another alternative fix could just be to print a warning during export.

Example url based on the actual example I had: 
`/remote_asset/https%3A%2F%2Fexample.com%2Fasset.svg%3Fcompress%3D50`
`(https://example.com/asset.svg?compress=50)`

Issue #1381, opened 2020-08-09
Provide TypeScript declarations out-of-the-box


benmccann:
Right now people have to copy the below file into their project. We should generate it automatically:

https://github.com/babichjacob/sapper-typescript-graphql-template/blob/main/types/%40sapper/index.d.ts

babichjacob:
Is it the right idea to make this only be done after running a Sapper-oriented fork of [`scripts/setupTypeScript.js`](https://github.com/sveltejs/template/blob/master/scripts/setupTypeScript.js) that gets bundled with the default Rollup and Webpack templates?

benmccann:
I was thinking that it could live in something like [`create_app`](https://github.com/sveltejs/sapper/blob/master/src/core/create_app.ts) and always happen. People don't generally open up their `src/node_modules/@sapper` directory, so it will be hidden from them and not affect them if they're not using TypeScript

benmccann:
I sent https://github.com/sveltejs/sapper/pull/1402 to provide TypeScript declarations automatically

We did get [a request to add a `scripts/setupTypeScript.js` to sapper-template](https://github.com/sveltejs/sapper-template/issues/251). That might not be a bad idea to do as a next step if anyone wants to take a stab at it

martonlederer:
I might take a look at it and see if I can do it @benmccann

PR #1402, opened 2020-08-13
Add TypeScript declarations

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1381

This is one of the biggest blockers to getting people started with TypeScript

This declaration file was posted in https://github.com/sveltejs/sapper/issues/760 and has been used by the community for several months, so it's fairly battle tested and a pretty good starting point even if there's probably room for further improvements. I copied it out of one of the template projects here: https://github.com/babichjacob/sapper-typescript-graphql-template/blob/main/types/%40sapper/index.d.ts

This also enables linting against the test code on my local machine, which was previously blocked by having the types available. It's not yet working on Github Actions though, so I'll have to do some debugging and follow up on enabling it in another change

### DIFF
--- a/runtime/index.d.ts
+++ b/runtime/index.d.ts
@@ -1,0 +1,44 @@
+declare module "@sapper/app"
+declare module "@sapper/server"
+declare module "@sapper/service-worker"
+
+declare module "@sapper/app" {
+	interface Redirect {
+		statusCode: number
+		location: string
+	}
+
+	function goto(href: string, opts: { noscroll?: boolean, replaceState?: boolean }): Promise<unknown>
+	function prefetch(href: string): Promise<{ redirect?: Redirect; data?: unknown }>
+	function prefetchRoutes(pathnames: string[]): Promise<unknown>
+	function start(opts: { target: Node }): Promise<unknown>
+	const stores: () => unknown;
+
+	export {
+		goto, prefetch, prefetchRoutes, start, stores
+	};
+}
+
+declare module "@sapper/server" {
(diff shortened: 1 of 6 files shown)


runtime/index.d.ts
```diff
@@ -0,0 +1,44 @@
...
+		goto, prefetch, prefetchRoutes, start, stores
+	};
+}
+
+declare module "@sapper/server" {
+	import { Handler, Req, Res } from '@sapper/internal/manifest-server';
+
+	interface MiddlewareOptions {
+		session?: (req: Req, res: Resp) => unknown
+		ignore?: unknown
```
pngwn, line 27: This should be typed as:

```ts
type Ignore = string | RegExp | (uri: string) => boolean | Ignore[];
```

runtime/index.d.ts
```diff
@@ -0,0 +1,44 @@
...
+declare module "@sapper/app" {
+	interface Redirect {
+		statusCode: number
+		location: string
+	}
+
+	function goto(href: string, opts: { noscroll?: boolean, replaceState?: boolean }): Promise<unknown>
+	function prefetch(href: string): Promise<{ redirect?: Redirect; data?: unknown }>
+	function prefetchRoutes(pathnames: string[]): Promise<unknown>
+	function start(opts: { target: Node }): Promise<unknown>
```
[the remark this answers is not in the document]
benmccann: done

runtime/index.d.ts
benmccann: done. thanks for giving this a look!

PR #1403, opened 2020-08-13
docs: mention that exporting now follows images

Conduitry:
On master, exporting now also follows images, and the docs should be updated to match. This shouldn't be merged (or at least shouldn't be deployed) until the next version gets released.

### DIFF
--- a/site/content/docs/10-exporting.md
+++ b/site/content/docs/10-exporting.md
@@ -44,7 +44,7 @@ You can also add a script to your package.json...
 
 When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>`, `<img>` and `<source>` elements it finds, and captures any data served by the app.
 
-Because of this, any pages you want to be included in the exported site must either be reachable by `<a>`, `<img>` or `<source>` elements or added to the `--entry` option of the `sapper export` command.
+Because of this, any pages you want to be included in the exported site must either be reachable by `<a>` elements or added to the `--entry` option of the `sapper export` command.
 
 The `--entry` option expects a string of space-separated values. Examples:
 
--- a/site/content/docs/10-exporting.md
+++ b/site/content/docs/10-exporting.md
@@ -42,7 +42,7 @@ You can also add a script to your package.json...
 
 ### How it works
 
-When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>`, `<img>` and `<source>` elements it finds, and captures any data served by the app.
+When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>`, `<img>`, `<link>` and `<source>` elements it finds pointing to local URLs, and captures any data served by the app.
 
 Because of this, any pages you want to be included in the exported site must either be reachable by `<a>` elements or added to the `--entry` option of the `sapper export` command.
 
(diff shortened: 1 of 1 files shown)


site/content/docs/10-exporting.md
```diff
@@ -42,9 +42,9 @@ You can also add a script to your package.json...
 
 ### How it works
 
-When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>` elements it finds, and captures any data served by the app.
+When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>`, `<img>` and `<source>` elements it finds, and captures any data served by the app.
 
-Because of this, any pages you want to be included in the exported site must either be reachable by `<a>` elements or added to the `--entry` option of the `sapper export` command.
+Because of this, any pages you want to be included in the exported site must either be reachable by `<a>`, `<img>` or `<source>` elements or added to the `--entry` option of the `sapper export` command.
```
benmccann, line 47: this sentence starts off talking about "pages", so maybe we shouldn't add ` or <source>` here since you probably wouldn't add them to `--entry`? (because what good is a directly crawled image if it's not reachable from some page)

Conduitry: That's a good point. Do you think we need any changes to this sentence? I'm thinking maybe the addition to the previous sentence is enough.

benmccann: yeah, I think the addition to the previous sentence is probably enough

site/content/docs/10-exporting.md
```diff
@@ -42,7 +42,7 @@ You can also add a script to your package.json...
 
 ### How it works
 
-When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>` elements it finds, and captures any data served by the app.
+When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>`, `<img>` and `<source>` elements it finds, and captures any data served by the app.
```
ignatiusmb, line 45: Tbh, I'm quite confused with this addition. Does it mean externally hosted images (e.g. links from cdn or s3) will be captured and stored locally in export directory? Also, with the addition of `<source>`, does it means it works with video formats too, i.e. mp4, webm, ogg?

Conduitry: No, as with `<a>`, only things that resolve to local paths served by your app are crawled and stored. For `<source>`, this refers only to `srcset` attributes. I'm definitely open to having the docs here explain this in more depth.

benmccann: It's probably a good clarification to add that it only crawls the local images

ignatiusmb: Local images meaning contents from `static`? If so, doesn't sapper already copied the whole folder over?

ignatiusmb: > For `<source>`, this refers only to `srcset` attributes.

Ah, I'm assuming this refers to `srcset` in `<source>` that is a direct descendant from `<picture>`, citing from https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#attr-srcset then, what about something like this?

```suggestion
When you run `sapper export`, Sapper first builds a production version of your app, as though you had run `sapper build`, and copies the contents of your `static` folder to the destination. It then starts the server, and navigates to the root of your app. From there, it follows any `<a>`, `<img>` and `<source>` (only for a direct child of `<picture>` element) elements it finds, and captures any data served by the app.
```

If we were to clarify `` too, perhaps it's better to separate them in another paragraph.

PR #1407, opened 2020-08-14
Switch call handle_error 404 to next function

lamualfa:
### Problem

I don't think the sapper needs to handle 404 it himself. Usually express users have their own 404 handler (Response as Page or JSON). Some Fastify users have complained about the difficulty of integrating with Sapper. This is because when the route is not found, the Sapper responds to the 404 page instead of passing it to the next handler.

#### Example:
```js
// Use express middleware in fastify
fastify.register(fastifyExpress)
    .after(() => {
      fastify.use(require('@sapper/server').middleware());
     
      // Will never be executed because svelte will respond to a 404 page
      fastify.use(require('sirv')(path.resolve('public')))
    })
```

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -377,8 +377,11 @@ export function get_page_handler(
 				return;
 			}
 		}
-
-		handle_error(req, res, 404, 'Not found');
+		
+ 		// Pass the 404 error to the user.
+		// Useful if using a non `express` framework such as fastify.
+		next();
+		return;
 	};
 }
 


antony:
This PR is missing some important things which will prevent us from reviewing or merging it:

1. It hasn't been discussed in an issue or RFC first, which is critical for a change of this type - it is a breaking change which will break **most** users's sites
2. It doesn't have any test coverage
3. It breaks a load of existing tests, and these would need to be fixed, by you, first.

lamualfa:
So, what should I do? close this request or create a fork yourself and try to fix it?

antony:
First it should probably be discussed as an RFC or a github issue to see if there is an appetite for this particular change.

If it looks like there is, you should make a fork, add tests for the change, make the change, and fix all tests which are broken.

Then, open a PR :)

PR #1410, opened 2020-08-15
Add more questions to FAQ

benmccann:
These are two questions that I see a lot in Discord. I'm hoping to save some time answering them :smile:


babichjacob:
(Just to establish the link between these two issues) This could be made unnecessary by successful completion of https://github.com/sveltejs/sapper-template/pull/252 (and Webpack equivalent).

benmccann:
Yeah, my plan was to keep the FAQ question, but update it point to the sapper-template after that PR is merged. I've debated whether it would also be helpful to still include links to other repos like yours as additional examples after that. Any opinion on what you'd like to see happen?

babichjacob:
> Any opinion on what you'd like to see happen?

I don't have an opinion.

site/content/faq/100-version-numbers.md
```diff
@@ -0,0 +1,7 @@
+---
+question: Sapper is not 1.0 yet. Is it production ready and should I use it?
+---
+
+The version number reflects that there are major features we'd like to add still. E.g. TypeScript support was added in Sapper 0.28.
+
+Sapper is operationally stable. For example, some of the Svelte maintainers run their businesses on Sapper.
```
antony, line 7: I think it's worth including something about:

 Sapper is operationally stable. "production readiness" is subjective, and if Sapper fulfils your requirements, then it's production ready. As with any software, a major version number is not an indicator of whether something is suitable for your specific use-case.

site/content/faq/300-typescript.md
```diff
@@ -0,0 +1,8 @@
+---
+question: How do I use TypeScript with Sapper?
+---
+
+The default template is currently setup only with JavaScript support, but here are some templates that community members have created with TypeScript support:
+
+* https://github.com/babichjacob/sapper-typescript-graphql-template
```
antony, line 7: I would probably point to the integrations section of the sveltejs/integrations repository rather than partially duplicating the content of it here.

antony (changes requested):

site/content/faq/300-typescript.md
benmccann: There aren't listed on `sveltejs/integrations` and I'm not sure there's a really good place to put them there since they're specific to Sapper and not Svelte in general

babichjacob: Sub-section "Sapper templates" under "Templates (boilerplates)"

antony: Yeah I suppose that's reasonable. Ok then this list seems fine to stay here.

site/content/faq/100-version-numbers.md
benmccann: done

Issue #1397, opened 2020-08-12
sapper@0.28.0 generates 2 css files with identical content


Ashot-KR:
**Describe the bug**
In `0.28.0` sapper generates files `main.[hash].css` and `client.[hash].css` with identical content.

**To Reproduce**
Create new sapper project from template
```
npx degit "sveltejs/sapper-template#rollup" my-app
cd my-app
npm install
```
Start app `npm run dev`
Open http://localhost:3000 in browser, open browser devtools on network/css
![Screenshot 2020-08-12 at 14 34 18](https://user-images.githubusercontent.com/899218/90011307-0b476200-dcaa-11ea-9c65-6ac477cedef6.png)
`main.[hash].css` and `client.[hash].css` has identical content.

This is css files generated with `sapper@0.27.16` and `rollup@2.23.1` (**UPD: added rollup version**)
![Screenshot 2020-08-12 at 14 33 21](https://user-images.githubusercontent.com/899218/90011461-4c3f7680-dcaa-11ea-893a-3a7b263d3cb9.png)

**Information about your Sapper Installation:**

```
  System:
    OS: macOS 10.15.4
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 615.80 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.7.0 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.7 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    rollup: ^2.3.4 => 2.23.1
    sapper: ^0.28.0 => 0.28.0
    svelte: ^3.17.3 => 3.24.1
```

Ashot-KR:
Screenshot of styles from inspector:
![image](https://user-images.githubusercontent.com/899218/90011798-e69fba00-dcaa-11ea-95e4-be2754279b62.png)

habibrosyad:
@Ashot-KR, what's your Rollup version with sapper v0.27.16?
If you're using Rollup v2, there was an issue here https://github.com/sveltejs/sapper-template/issues/233 which was solved with the release of v0.28.

Ashot-KR:
@habibrosyad `rollup@2.23.1`

habibrosyad:
[...]

PR #1098, opened 2020-02-23
Leftover css should read from unclaimed modules

mapgrid:
Fixes #1076.

Commit [1b9b559](https://github.com/sveltejs/sapper/commit/1b9b559d824178e808ade40088c05e474949b4e4#diff-13ad31a0134fae29917411210deb70bdL219) changed the array that `leftover` CSS pulls from. This causes `main.css` to include all CSS already included by `entry_css_modules` instead of only CSS from unclaimed modules. In practice, it duplicates all CSS to `main.css`.

**Original**

```js
const leftover = get_css_from_modules(Array.from(unaccounted_for));
```

**`1b9b559`**
```js
const leftover = get_css_from_modules(entry_css_modules, css_map, dirs);
```

**Revert (in this PR)**
```js
const leftover = get_css_from_modules(Array.from(unclaimed), css_map, asset_dir);
```

PR #1413, opened 2020-08-17
Correctly calculate leftover css

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1397

It looks like the bug was introduced by https://github.com/sveltejs/sapper/pull/471, which made `leftover` include the css `entry_modules`. I don't understand the issue there or how to test that this doesn't reintroduce the issue that was being addressed

### DIFF
--- a/src/core/create_compilers/extract_css.ts
+++ b/src/core/create_compilers/extract_css.ts
@@ -122,8 +122,7 @@ export default function extract_css(
 	let asset_dir = `${dirs.dest}/client`;
 	if (process.env.SAPPER_LEGACY_BUILD) asset_dir += '/legacy';
 
-	const unclaimed = new Set(client_result.css_files.map(x => x.id));
-
+	const unclaimed = new Set<string>(client_result.css_files.map(x => x.id));
 	const lookup = new Map();
 	client_result.chunks.forEach(chunk => {
 		lookup.set(chunk.file, chunk);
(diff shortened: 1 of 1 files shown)


Conduitry:
There's another related issue that appears to have emerged in either in Sapper 0.28 or upon upgrading Rollup or one of its plugins.

Previously, you were able to simply `import './path/to/some.css'` in any `.svelte` or `.js` file and that CSS file would be included on any page that used that file. This is no longer working on 0.28 with the latest template nor when using this PR.

I suspect it was this type of CSS that #471 was dealing with. If both the `_layout` and a particular page depended on some particular CSS (not a `<style>` tag in a component), it sounds like these styles were getting incorrectly skipped.

Separately, it looks like this PR also causes the same or similar issues as #1098 did - that all of the component styles are now getting inlined into a `<style>` tag in the server-rendered HTML.

Conduitry:
Using the current template but downgrading to Rollup 1 seems to make the `import 'foo.css'` thing partly work again.

With Sapper 0.28.0 and Rollup 1, CSS files imported from components end up in both the server-rendered HTML.

With this PR and Rollup 1, CSS files imported from components only seem to end up in the client-rendered versions of the components. The styles are not applied until js is run.

benmccann:
> Previously, you were able to simply import `'./path/to/some.css'` in any .svelte or .js file and that CSS file would be included on any page that used that file.

Is that a Svelte feature or purely a Sapper feature?

> Modules that are completely tree-shaken will no longer be listed as part of any chunks in generateBundle

That's what https://github.com/sveltejs/sapper/pull/1306 was supposed to address. They were being tree-shaken with Rollup 2, so we added an option telling it to stop tree shaking them

Conduitry:
Just a Sapper feature. The compiler doesn't touch any of the imports in any of your components, and rollup-plugin-svelte doesn't have any special handling of them. I think some handling of this is built into webpack.

PR #1420, opened 2020-08-20
Compile to ES6

benmccann:
This should result in a smaller bundle for modern browsers. For older browsers, the template includes Babel

My main reason for wanting this though is because VS Code doesn't seem to recognize `downlevelIteration` and shows errors everywhere we iterate over a `Set`, which is super annoying. That's fixed if you treat the code as ES6

### DIFF
--- a/src/api/build.ts
+++ b/src/api/build.ts
@@ -32,9 +32,9 @@ export async function build({
 	static: static_files = 'static',
 	dest = '__sapper__/build',
 
-	bundler,
+	bundler = undefined,
 	legacy = false,
-	ext,
+	ext = undefined,
 	oncompile = noop
 }: Opts = {}) {
 	bundler = validate_bundler(bundler);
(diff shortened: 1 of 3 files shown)


antony:
@benmccann I like it, but this absolutely has to be breaking, right?

This is going to end up causing people to have to change their babel config.

benmccann:
I'm not so sure that it will be. The template has `targets: '> 0.25%, not dead'`, which means that Babel should transpile the code appropriately. I can't think of a way this would break users actually. Sapper already doesn't support non-modern browsers by default, and if you are targeting a legacy browser then you already would have had to setup Babel to handle that browser as a target.

Conduitry:
What is this affecting the transpilation of? The Sapper runtime that gets copied into `src/node_modules/@sapper`? I think it makes sense to stop compiling that to ES5 - there's already a lot of non-ES5-compatible stuff that's going to be coming from the compiled Svelte components.

benmccann:
This affects what's in `node_modules/sapper`. The stuff in `src/node_modules/@sapper` gets copied from there by `copy_runtime`, so it would also be affected

Conduitry:
All right, so the code that's run during the building of the Sapper app would also now be ES6, which would also be fine. Makes sense, I think this is a good change.

Issue #1431, opened 2020-08-23
Cleaner separation of router component


benmccann:
The stuff that `start/index.ts` imports from `app` could be moved into `start/index.ts`. We'd then have all the routing info in one place and a clean separation between app and routing. That file could probably be split out into a separate project and used as a standalone router down the road

PR #1434, opened 2020-08-24
Routing refactor

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1431. This is a cleanup that separates out the navigation functionality into a standalone file

This does the following:

* Renames `start/index.ts` to `router/index.ts`
* Moves routing-related methods that were in `app` into `router`
* Makes `app` import and instantiate `router` instead of the other way around

There are no user-visible changes in this PR. This PR looks big, but it's really just moved existing code and hasn't changed much at all. It would probably be hard to review the actual code to verify what's changed, but I hope the above provides a good summary. Since it's probably easier to review the description than the code, please let me know if you have questions


ehrencrona:
I hadn't looked much at `app.ts` before but this makes a lot of sense to me. The code feels a bit tangled at the moment and this is a good step in giving it more structure.

I did notice this doesn't quite disentangle the dependencies: `router/index.ts` depends on `prefetch/index.ts` which in turn depends on `app.ts`. I believe this is a code smell telling us that the prefetching should be moved (to `prefetch/index.ts`). This should be easy: two functions need to move and prefetch would also need an `init`.

Furthermore, I'm a bit worried by `router` not actually doing any routing. Maybe it makes sense in the longer term but for now I believe the name `navigation` might be more fitting.

benmccann:
Agreed we could clean up `prefetch` a bit as well. That'd be a good follow up. I didn't want this one to get any bigger

My plan is to allow registering routes here as a next step. See https://github.com/sveltejs/rfcs/pull/36 for more details on that. I figured it's better to leave this called router for now to avoid renaming multiple times and making this history harder to follow

antony (approved):
This looks very clean and isolates the router perfectly.

PR #1439, opened 2020-08-26
Make dev mode reflect changes to css

benmccann:
When you make a change to a css file the change isn't showing up. The issue was in the lines:

```
that.css_files.push({ id, code });
.
.
.
const f = that.css_files.find(file => file.id === css_module);
```

When a change was encountered it would push the updated code to the end of the array, but then we'd find the first version. I changed it to a map instead and it works now.

Ironically this was pretty much the only line of css handling that didn't change in https://github.com/sveltejs/sapper/pull/1415. This sounded like a new issue from the bug reporter in Discord, but I'm not quite sure why it wouldn't have been happening before since it didn't change. Possibly the duplicate CSS bug covered it up if the original CSS didn't change and the duplicate did.

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -53,7 +53,7 @@ export default class RollupCompiler {
 	warnings: any[];
 	errors: any[];
 	chunks: any[];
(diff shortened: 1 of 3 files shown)


antony (approved):
Can confirm this works with my minimal reproduction!

Issue #1448, opened 2020-08-26
Route chunks are not included in preload headers


Jayphen:
Preload headers are included for all JS/CSS chunks loaded on any route, except for the route chunk itself.

For example, these are the preload headers for the document at https://sapper-preload-repro.vercel.app/blog/why-the-name

`</client/client.1af273ec.js>;rel="modulepreload";as="script", </client/assets/client-24271426.css>;rel="preload";as="style", </client/assets/[slug]-17a7b34f.css>;rel="preload";as="style"`

Note that the `[slug]-[hash].css` chunk from the route is preloaded, but not the `.js` file.

If you run Pagespeed against the above page (https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fsapper-preload-repro.vercel.app%2Fblog%2Fwhy-the-name), you'll get the following message indicating that the performance score has been reduced because of this:

![CleanShot 2020-08-26 at 23 53 24](https://user-images.githubusercontent.com/329184/91360765-56f42280-e7f7-11ea-85cc-6c0f8c5789ed.png)

The reproduction is here - it's just the base sapper template configured to deploy to vercel, and a couple of styles added https://github.com/Jayphen/sapper-preload-repro

benmccann:
Hmm. `[slug].hash.js ` is not listed in `__sapper__/dev/build.json` under the route `blog/[slug].svelte` as I would expect it to be. That's where the preload headers are pulled from

benmccann:
Ah, it seems this is a result of calculating the dependency tree in a new way and I didn't notice this line:

https://github.com/sveltejs/sapper/blob/de87668c9c9429f2033d249b52a9cae5c4d4b7f9/src/core/create_compilers/RollupCompiler.ts#L132

Issue #1447, opened 2020-08-26
CSS from dynamically imported components missing (^0.28.1)


Jayphen:
When dynamically importing a component, that component's styles are not loaded. This regression was introduced in 0.28.1.

I have a repo that reproduces this here: https://github.com/Jayphen/sapper-preload-repro/pull/2/files

The above is based off the base svelte/sapper template, with no modifications.

It is deployed to Vercel here: https://sapper-preload-repro-git-dynamic-import-css.jayphen.vercel.app.

If you check out the above branch and fix the version to 0.28.0, you'll see the text in the header is styled as it should be.

Note that `preserveEntrySignatures` is still in the config in this repro, unlike the one I mentioned in https://github.com/sveltejs/sapper/issues/1446#issuecomment-681135859

benmccann:
Dang I thought this was something I tested. There's even a test for it:
https://github.com/sveltejs/sapper/blob/de87668c9c9429f2033d249b52a9cae5c4d4b7f9/test/apps/css/test.ts#L43

I wonder how these cases are different

benmccann:
Ah, I think I know what's happening now. The issue is that the dynamic import is in the layout (in this case it's actually in the nav, which is used only in the layout). If it were in a route then it would work.

It seems we need to make the css rollup plugin aware of the routes and put all css that doesn't map to a route into the entry chunk's css. I'm not sure how I feel about that though. That's a very tight coupling between Sapper and the Rollup plugin and we were trying to add separation and possibly split the Rollup plugin into its own project. Another solution would be to handle dynamic imports differently depending whether svelte component being imported contains `/routes/` in the path or not. That might bake in a bit of an undocumented assumption in for the short term, but seems a better in the long-term because it's we could make it configurable when we split the Rollup plugin out into its own project.

btakita:
Please fix this issue before the next release. Thank you

matt3224:
I also have this issue since 0.28.1 and also in 0.28.2, it's in all the routes i've used `svelte:component` in, just FYI

Issue #1446, opened 2020-08-26
Some CSS Missing since 0.28.1


benmccann:
**Reproduction 1**
https://github.com/juho/sapper-rollup-css-strip

**Reproduction 2**
On Sapper v0.28.0: https://svelte-materialify.vercel.app/components/alerts/
On Sapper v0.28.2: https://svelte-materialify-g0c0gdngr.vercel.app/components/alerts/

Alert styles aren't being rendered. This is the scss file: https://github.com/TheComputerM/svelte-materialify/blob/sapper%400.28.1/packages/svelte-materialify/src/components/Alert/Alert.scss

**Additional Info**
Components which are not part of the _layout.svelte or are not included by components in _layout.svelte do not have their CSS generated.
Meaning components only included in routes do not have their CSS generated

antony:
That's why I couldn't recreate the issue!

juho:
adding these in rollup config makes the repro work:
`preserveEntrySignatures: false` in the client section and `preserveEntrySignatures: "strict"` in the server section

Jayphen:
You can reproduce this by cloning the default sapper template and removing `preserveEntrySignatures` from the rollup conf

I made a repro here: https://github.com/Jayphen/sapper-preload-repro/tree/broken-css and deployed it here https://sapper-preload-repro-git-broken-css.jayphen.vercel.app/blog/why-the-name

TheComputerM:
Here is the reproduction: https://github.com/TheComputerM/sapper-bug

benmccann:
Thanks. I can reproduce using juho/sapper-rollup-css-strip. I don't have `yarn` installed and have never used it so I will use juho/sapper-rollup-css-strip to test since TheComputerM/sapper-bug is heavily reliant on `yarn`

TheComputerM:
@benmccann I have updated my repo to use npm.

kaipaysen:
A Workaround would be to add the component to _layout.svelte and wrap it within display:none like this:
```html
<!-- workaround until https://github.com/sveltejs/sapper/issues/1446 is fixed -->
<div style="display: none">
    <KontaktBar />
</div>
```

PR #1452, opened 2020-08-28
Fix multiple asset handling issues

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1446
Closes https://github.com/sveltejs/sapper/issues/1447
Closes https://github.com/sveltejs/sapper/issues/1448

This makes a few changes:
* replaces `chunk.ts` with `rollup-dependency-tree`, which means this PR removes about 100 lines from the codebase overall. (`rollup-dependency-tree` itself is about 50 lines). Its API helps make it a little easier to implement some of the other features and, while a bit subjective, I think using it helps with readability. Probably most importantly, it handles circular dependencies
* Handle different values of `preserveEntrySignatures` in the user's Rollup configuration
* Handle dynamic imports in layouts. 2.8.1 only handled dynamic imports in routes, which was a regression
* Add comments to explain the various edge cases and logic. The original `extract_css` was written by quite the Rollup expert and handled a couple cases that neither adamb nor myself recognized it was handling. Hopefully the explanation added here will help others be able to understand and maintain this code more easily


package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "html-minifier": "^4.0.0",
     "http-link-header": "^1.0.2",
+    "rollup-dependency-tree": "0.0.10",
```
Conduitry, line 24: Is there a reason for this to be an external dependency? Not that it matters _a lot_, since it's only going to be in play when building the Sapper app, but, especially since this is already pinned to a particular version, would it make sense to bundle this instead?

benmccann: It could be useful to other projects, which was part of it. But the main reason I did it was that I was having a very difficult time getting the test infrastructure setup within Sapper without breaking the existing tests. The library has some rollup output saved as JSON that I then load and test that it's handled correctly. Sapper's test setup really didn't like that I was trying to import JSON as a module and I couldn't find any combination of settings that would allow me to do that without breaking the setup of existing tests

PR #1468, opened 2020-09-01
preload typings

### LINKED ISSUE #1463: [TypeScript] Provide typings for preload function (especially `this` context)
People are having a hard time trying to figure out how to type the preload function correctly, especially when using `this` (like in `this.fetch(...)`). [It is possible to do by adding a `this` type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions) to the `preload` function, but this may not be widely known.

Provide typings for the `preload` function for people to import.

```html
<script context="module">
    import type { PreloadContext, PreloadPage, PreloadSession } from "@sapper/preload";

    export async function preload(this: PreloadContext, page: PreloadPage, session: PreloadSession) {
       ...
       this.fetch(...) // <-- works without type errors
    }
</script>
```

Also note this in the official docs so it's easy to discover how to do this.

(I'm happy to help / provide a PR if you give me pointers on where to add this)

Don't provide the functions through some `this` context, but through a third parameter of the function: `export async function preload(page, session, {fetch, error, redirect}) { ...}`. This would be a breaking change of course.

**Additional context**
 [I also started thinking](https://github.com/sveltejs/language-tools/issues/497) about adding something to the language-server but ditched that because you can type it yourself, and because the `this`-context could be enhanced by the user in the future ([if I'm interpreting @benmccann correctly](https://github.com/sveltejs/language-tools/pull/499#issuecomment-683423876)).

dummdidumm:
closes #1463

I added the typings to the `d.ts` file and put it in `@sapper/common` which seemed fitting to me since it is used on both client and server. I also added an alternative way of typing the function like @jasonlyu123 suggested, but did not mention it in the docs (I can add that if you want to).
I also added a doc entry.

### DIFF
--- a/runtime/index.d.ts
+++ b/runtime/index.d.ts
@@ -1,6 +1,7 @@
 declare module "@sapper/app"
 declare module "@sapper/server"
 declare module "@sapper/service-worker"
+declare module "@sapper/common"
 
 declare module "@sapper/app" {
 	export interface Redirect {
@@ -34,4 +35,23 @@ declare module "@sapper/service-worker" {
 	export const assets: string[];
 	export const shell: string[];
 	export const routes: Array<{ pattern: RegExp }>;
+}
+
+declare module "@sapper/common" {
+	export interface PreloadContext {
+		fetch: (url: string, options?: any) => Promise<any>;
+		error: (statusCode: number, message: Error | string) => void;
+		redirect: (statusCode: number, location: string) => void;
+	}
+
+	export interface PreloadPage {
+		host: string;
+		path: string;
+		params: Record<string, string>;
+		query: Record<string, string | string[]>;
+	}
+
+	export interface PreloadFunction {
+		(this: PreloadContext, page: PreloadPage, session: any): any | Promise<any>;
+	}
 }
--- a/runtime/index.d.ts
+++ b/runtime/index.d.ts
@@ -52,7 +52,7 @@ declare module "@sapper/common" {
 		error?: Error;
 	}
 
-	export interface PreloadFunction {
+	export interface Preload {
 		(this: PreloadContext, page: Page, session: any): object | Promise<object>;
 	}
 }
--- a/runtime/index.d.ts
+++ b/runtime/index.d.ts
@@ -44,7 +44,7 @@ declare module "@sapper/common" {
 		redirect: (statusCode: number, location: string) => void;
 	}
 
-	export interface PreloadPage {
+	export interface Page {
 		host: string;
 		path: string;
 		params: Record<string, string>;
@@ -53,6 +53,6 @@ declare module "@sapper/common" {
 	}
 
 	export interface PreloadFunction {
-		(this: PreloadContext, page: PreloadPage, session: any): object | Promise<object>;
+		(this: PreloadContext, page: Page, session: any): object | Promise<object>;
 	}
 }
--- a/runtime/index.d.ts
+++ b/runtime/index.d.ts
@@ -49,9 +49,10 @@ declare module "@sapper/common" {
 		path: string;
 		params: Record<string, string>;
 		query: Record<string, string | string[]>;
+		error?: Error;
 	}
 
 	export interface PreloadFunction {
-		(this: PreloadContext, page: PreloadPage, session: any): any | Promise<any>;
+		(this: PreloadContext, page: PreloadPage, session: any): object | Promise<object>;
 	}
 }
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -139,7 +139,7 @@ If you use TypeScript and want to access the above context methods, TypeScript w
 
 ```html
 <script context="module">
-	import { PreloadContext, PreloadPage } from "@sapper/common";
+	import type { PreloadContext, PreloadPage } from "@sapper/common";
 
 	export async function preload(this: PreloadContext, page: PreloadPage, session: any) {
 		const { user } = session;
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -139,7 +139,7 @@ If you use TypeScript and want to access the above context methods, TypeScript w
 
 ```html
 <script context="module">
-	import type { PreloadContext, Page } from "@sapper/common";
+	import type { Page, PreloadContext } from "@sapper/common";
 
 	export async function preload(this: PreloadContext, page: Page, session: any) {
 		const { user } = session;
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -141,7 +141,7 @@ If you use TypeScript and want to access the above context methods, TypeScript w
 <script context="module">
 	import type { PreloadContext, Page } from "@sapper/common";
 
-	export async function preload(this: PreloadContext, page: PreloadPage, session: any) {
+	export async function preload(this: PreloadContext, page: Page, session: any) {
 		const { user } = session;
 
 		if (!user) {
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -135,7 +135,7 @@ You can abort rendering and redirect to a different location with `this.redirect
 
 #### Typing the function
 
-If you use TypeScript, you need to type the `this` context. You can type the function like this:
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
 
 ```html
 <script context="module">
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
 	}
 </script>
 ```
+
+#### Typing the function
+
+If you use TypeScript, you need to type the `this` context. You can type the function like this:
+
+```html
+<script context="module">
+	import { PreloadContext, PreloadPage } from "@sapper/common";
+
+	export async function preload(this: PreloadContext, page: PreloadPage, session: any) {
+		const { user } = session;
+
+		if (!user) {
+			return this.redirect(302, 'login'); // TypeScript will know the type of `this` now
+		}
+
+		return { user };
+	}
+</script>
+```
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -139,7 +139,7 @@ If you use TypeScript and want to access the above context methods, TypeScript w
 
 ```html
 <script context="module">
-	import type { PreloadContext, PreloadPage } from "@sapper/common";
+	import type { PreloadContext, Page } from "@sapper/common";
 
 	export async function preload(this: PreloadContext, page: PreloadPage, session: any) {
 		const { user } = session;
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -141,7 +141,7 @@ If you use TypeScript and want to access the above context methods, TypeScript w
 <script context="module">
 	import type { Page, PreloadContext } from "@sapper/common";
 
-	export async function preload(this: PreloadContext, page: Page, session: any) {
+	export const preload: Preload = async function(this, page, session) {
 		const { user } = session;
 
 		if (!user) {


site/content/docs/04-preloading.md
```diff
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
...
 </script>
 ```
+
+#### Typing the function
+
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
+
+```html
+<script context="module">
+	import { PreloadContext, PreloadPage } from "@sapper/common";
```
benmccann, line 142: Do we need to do `import type` here?  (https://svelte.dev/faq#what-about-typescript-support)

dummdidumm: Good catch, yes we do

dummdidumm: Add the types

site/content/docs/04-preloading.md
```diff
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
...
+
+#### Typing the function
+
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
+
+```html
+<script context="module">
+	import type { PreloadContext, PreloadPage } from "@sapper/common";
+
+	export async function preload(this: PreloadContext, page: PreloadPage, session: any) {
```
benmccann, line 144: should we try to add a type for `session` in this example as well?

dummdidumm: From reading the docs I though this is user-defined. Do you mean something like this?

```ts
import type { PreloadContext, PreloadPage } from "@sapper/common";
import type { Session } from './your-own-session-interface';

export async function preload(this: PreloadContext, page: PreloadPage, session: Session) {
...
```

benmccann: Oh, nevermind then. There's also a `session` store and I thought that there would be a store interface we'd want to use here, but looking at the docs it looks like this might be different. I'm not that familiar with `session` myself yet. I think we can leave this as is

runtime/index.d.ts
```diff
@@ -35,3 +36,22 @@ declare module "@sapper/service-worker" {
...
+	}
+
+	export interface PreloadPage {
+		host: string;
+		path: string;
+		params: Record<string, string>;
+		query: Record<string, string | string[]>;
+	}
+
+	export interface PreloadFunction {
```
benmccann, line 54: How/where would this be used?

runtime/index.d.ts
```diff
@@ -35,3 +36,22 @@ declare module "@sapper/service-worker" {
...
 }
+
+declare module "@sapper/common" {
+	export interface PreloadContext {
+		fetch: (url: string, options?: any) => Promise<any>;
+		error: (statusCode: number, message: Error | string) => void;
+		redirect: (statusCode: number, location: string) => void;
+	}
+
+	export interface PreloadPage {
```
benmccann, line 47: I wonder if we should call this just `Page`?

runtime/index.d.ts
dummdidumm:
```ts
export const preload: PreloadFunction = (page, session) => {
    this... // <- works, too
};
```

runtime/index.d.ts
dummdidumm: I thought about this, too, and decided against it since `Page` is such a generic word that I fear it might clash with other people's namings. But I totally don't mind changing this to `Page`.

runtime/index.d.ts
```diff
@@ -35,3 +36,22 @@ declare module "@sapper/service-worker" {
 	export const shell: string[];
 	export const routes: Array<{ pattern: RegExp }>;
 }
+
+declare module "@sapper/common" {
+	export interface PreloadContext {
+		fetch: (url: string, options?: any) => Promise<any>;
```
Jayphen, line 42: Would it be appropriate to use the `node-fetch` types here instead of typing opts and res as any?

```ts
    fetch: (url: RequestInfo, init?: RequestInit) => Promise<Response>;
```

benmccann: That seems like it would probably work for `Response`:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/07b2759140ad390e14801cfcc5ed9ff94182e7bb/types/node-fetch/index.d.ts#L167

For the options, `node-fetch` extends the whatwg/fetch standard options with its own:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/07b2759140ad390e14801cfcc5ed9ff94182e7bb/types/node-fetch/index.d.ts#L46

That would only work on the server and I'm not sure we'd want to expose the underlying implementation to that degree. Perhaps we should just define our own interface for that? 
```
export interface FetchOpts {
    body?: BodyInit;
    headers?: HeadersInit;
    method?: string;
    redirect?: RequestRedirect;
    signal?: AbortSignal | null;
}
```

Or maybe we can find types for the browser `fetch` that we can use instead?

Jayphen: Ah, I didn't realise node-fetch was a superset of the fetch standard. In that case it's safer to use the equivalent signature for native fetch

https://microsoft.github.io/PowerBI-JavaScript/modules/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.html#fetch

Jayphen:
I think this looks good. Apart from the typing of `fetch` as I commented above, I've been using basically the same types in my own apps. I'm unaware of a better way to handle the types here.

runtime/index.d.ts
Jayphen: … in which case I think `fetch: WindowOrWorkerGlobalScope["fetch"]` would work

runtime/index.d.ts
```diff
@@ -35,3 +36,22 @@ declare module "@sapper/service-worker" {
...
+		error: (statusCode: number, message: Error | string) => void;
+		redirect: (statusCode: number, location: string) => void;
+	}
+
+	export interface PreloadPage {
+		host: string;
+		path: string;
+		params: Record<string, string>;
+		query: Record<string, string | string[]>;
+	}
```
benmccann, line 52: There's a new `error` property being added in https://github.com/sveltejs/sapper/pull/1498. We should probably go ahead and merge that one and then add it here

runtime/index.d.ts
ehrencrona: Note that this definition already exists as `Page` in `runtime/src/app/types.ts`

runtime/index.d.ts
ehrencrona: I think we should require that it return an object (rather than `any`). I assume returning say a string is not supported.

ehrencrona:
It would be good if we could also use these types in the actual sapper code to guarantee that we conform to them. I did a stab at that [in this branch](https://github.com/sveltejs/sapper/compare/master...ehrencrona:preload-typings) and it did reveal that we actually don't :) (it's possible to pass in a `Request` rather than just a `string` as the first parameter of `fetch`; also the headers can be of different types, though I didn't solve that cleanly).

I did wind up duplicating `PreloadContext` into `types.ts` because I couldn't figure out how to import it from `index.d.ts`.

@dummdidumm Feel free to merge my branch into this one if you agree with the changes.

ehrencrona:
Sorry, one last comment. I find myself unable to use these typings in project code. I set up a TypeScript Sapper project using [these instructions](https://dev.to/mhaecker/use-typescript-with-svelte-sapper-45n8), but when importing either `@sapper/app` or `@sapper/common` I get `Cannot find module '@sapper/common' or its corresponding type declarations.ts(2307)`.

I haven't been able to figure out how this is supposed to work. My guess would have been that we need a `type` attribute in the `package.json`, but changing that doesn't fix it for me.

tommywalkie:
@ehrencrona I had the same issue, I simply had to link the `types` field to `sapper/runtime`...
https://github.com/eksplein/website/blob/0feea3ecd6dd1ddb017ffaebfc94a536bd9af1d7/tsconfig.json#L14
because this is where the `index.d.ts` file is.
https://github.com/sveltejs/sapper/blob/e97846c1a882c7dcdd1a256b753a7f1c58feb1e5/runtime/index.d.ts#L1-L3

dummdidumm:
If I remember correctly there's some copy-script that copies these typings into src/node_modules which should make them available. But that's not part of this PR.

runtime/index.d.ts
benmccann: Yeah, we set the return type to `object | Promise<object>` in https://github.com/sveltejs/sapper/pull/1528

dummdidumm:
How to move forward? I updated the PR with all suggestions except the `fetch` typing which seemed to me that there is no concensus yet.

runtime/index.d.ts
```diff
@@ -35,3 +36,23 @@ declare module "@sapper/service-worker" {
...
+
+	export interface Page {
+		host: string;
+		path: string;
+		params: Record<string, string>;
+		query: Record<string, string | string[]>;
+		error?: Error;
+	}
+
+	export interface PreloadFunction {
```
benmccann, line 55:
```suggestion
	export interface Preload {
```

benmccann:
@dummdidumm thanks for pushing this through. I just had one more question about naming, but this looks good to me

site/content/docs/04-preloading.md
```diff
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
...
 </script>
 ```
+
+#### Typing the function
+
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
+
+```html
+<script context="module">
+	import type { PreloadContext, PreloadPage } from "@sapper/common";
```
benmccann, line 142:
```suggestion
	import type { PreloadContext, Page } from "@sapper/common";
```

site/content/docs/04-preloading.md
```diff
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
...
+
+#### Typing the function
+
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
+
+```html
+<script context="module">
+	import type { PreloadContext, Page } from "@sapper/common";
+
+	export async function preload(this: PreloadContext, page: PreloadPage, session: any) {
```
benmccann, line 144:
```suggestion
	export async function preload(this: PreloadContext, page: Page, session: any) {
```

site/content/docs/04-preloading.md
```diff
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
...
 </script>
 ```
+
+#### Typing the function
+
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
+
+```html
+<script context="module">
+	import type { PreloadContext, Page } from "@sapper/common";
```
benmccann, line 142:
```suggestion
	import type { Page, PreloadContext } from "@sapper/common";
```

site/content/docs/04-preloading.md
```diff
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
...
+
+#### Typing the function
+
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
+
+```html
+<script context="module">
+	import type { Page, PreloadContext } from "@sapper/common";
+
+	export async function preload(this: PreloadContext, page: Page, session: any) {
```
benmccann, line 144: Do we also want to use the function typing in this example?

```suggestion
	export const preload: Preload = async function(this: PreloadContext, page: Page, session: any) {
```

dummdidumm: This would be duplicated typing. The idea is to do either one or the other. Using the function typing means you don't need typings for `this` and the params.

benmccann: ah, right. still I wonder which is the better way to demonstrate it. using the `Preload` type seems a bit safer (it means you didn't put the wrong type for any parameter and adds typing for the return value) and it's also more concise

benmccann:
thanks for all your work on this!! this is a great improvement

runtime/index.d.ts
Jayphen: Just wondering — why wasn’t this typed in the end?

dummdidumm: As mentioned below I felt there was no concensus yet on what the typings should be. Feel free to open another PR for a dedicated discussion around this and possibly changing it.

benmccann: would `WindowOrWorkerGlobalScope` work on the server-side? I'm not that familiar with typescript, but it sounds like a browser thing

btakita:
Thank you for releasing the types! I'm getting an error when referencing the `Page` type. I added `"sapper/common"` to the `tsconfig.json` `types`.

```ts
import { Page } from '@sapper/common'
export type $page_type = Page;
```

```
TS2709: Cannot use namespace 'Page' as a type.
TS4081: Exported type alias '$page_type' has or is using private name 'Page'.
```

site/content/docs/04-preloading.md
```diff
@@ -132,3 +132,23 @@ You can abort rendering and redirect to a different location with `this.redirect
...
 </script>
 ```
+
+#### Typing the function
+
+If you use TypeScript and want to access the above context methods, TypeScript will thow an error and tell you that it does not know the type of `this`. To fix it, you need to specify the type of `this` (see the [official TypeScript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions)). We provide you with helper interfaces so you can type the function like this:
+
+```html
+<script context="module">
+	import type { Page, PreloadContext } from "@sapper/common";
```
ignatiusmb, line 142: Sorry to comment on a merged PR, but both variables doesn't seem to be used below, shouldn't it be `Preload`?

dummdidumm: Good catch, yes this is wrong, it should be `Preload`.

dummdidumm: Fixed on master

PR #1478, opened 2020-09-02
Let `this.fetch` forward cookies received on the server

croigian:
## What does this solve?

Requests performed in preload functions might set cookies in their response headers.
When executed on the server, this data is lost as headers are not forwarded to the client.
Current work-around requires code duplication: one part in a script tag, the other in a middleware to have access to a `res` object.

## How does it solve it?

This PR adds an option to `this.fetch` to let callers ensure cookies are forwarded in the response if run server-side.

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -112,7 +112,7 @@ export function get_page_handler(
 			error: (statusCode: number, message: Error | string) => {
 				preload_error = { statusCode, message };
 			},
-			fetch: (url: string, opts?: any) => {
+			fetch: async (url: string, opts?: any) => {
 				const protocol = req.socket.encrypted ? 'https' : 'http';
 				const parsed = new URL.URL(url, `${protocol}://127.0.0.1:${process.env.PORT}${req.baseUrl ? req.baseUrl + '/' :''}`);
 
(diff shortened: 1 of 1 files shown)


benmccann:
I believe that this method is trying to maintain API compatibility with the browser's `fetch` method, so this probably isn't how we'd want to implement this: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Parameters

croigian:
Got it. What do you think about forwarding received cookies by default then?
The current behaviour leads to different outcomes depending on where preload runs, which as I understand is what `this.fetch` tries to avoid.

benmccann:
It kind of looks to me like that's already being done in the block above in the `if (include_credentials)` block. I guess the question would be what's different about your usecase such that `include_credentials` is `false` - are you making requests across domains?

croigian:
The block above forwards request cookies, this PR is about forwarding the ones set by the response.
An example use case is an API for authentication setting a JWT in a `Set-Cookie` header.
Currently, if the request is performed from `preload`, the cookie is set on the browser only if `preload` runs client-side.

benmccann:
I'd probably want a second pair of eyes on it, but that sounds reasonable. I think if it were done that way I'd expect it to happen inside the `if (include_credentials)` block.

croigian:
The common authentication usecase would fit in this block yes.
However, including credentials in the request and forwarding response cookies seem like different concerns to me.
APIs might store arbitrary cookies on the client, not just for authentication purposes.
This can include requests to endpoints accessible without `credentials: 'include'`.

benmccann:
Thanks for the PR! And sorry it took awhile to review. It's probably best to test if this works as hoped in SvelteKit. We wouldn't want to have it implemented only in Sapper for people to lose functionality when migrating

benmccann:

PR #1482, opened 2020-09-03
Remove Rollup CSS handling

benmccann:
Add `rollup-plugin-postcss` to the template (https://github.com/sveltejs/sapper-template/pull/260) instead of using our own CSS handling

See https://github.com/sveltejs/rfcs/pull/29 for more details


babichjacob:
This is actually breaking some things relative to 0.28.4. See this [reproduction based on the default Sapper Rollup template](https://github.com/babichjacob/sapper-rm-css-repro).

The *only* change (besides upgrading all packages to their latest versions) is 
```html
<script>
	import { tweened } from "svelte/motion";

	const hue = tweened(0);
	$: $hue += 30;
</script>

<!-- .. -->


<h1 style="color: hsl({$hue}deg, 70%, 50%)">Great success!</h1>
```

added to easily discern when client-side JS is and is not working.

So the problems I've seen so far are:
1. Client-side JavaScript is broken because there's an `undefined` in the path that it's being looked for at (i.e. the script tag is `<script src="/client/undefined" type="module" crossorigin="use-credentials"></script>` now). The header text "GREAT SUCCESS!" is supposed to hue-rotate but it'll be frozen instead because of this.
2. CSS, as far as I can tell, is fine but (useless?) errors on the generated CSS are appearing like so:

```
> TODO@0.0.1 dev /Users/babichjacob/Repositories/sapper-rm-css-repro
> sapper dev

✗ client
Identifier directly after number (Note that you need plugins to import files that are not JavaScript)
1: main.svelte-1uhnsl8{position:relative;max-width:56em;background-color:white;padding:2em;margin:0 auto;box-sizing:border-box}
                ^
2: /*# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoiX2xheW91dC5zdmVsdGUiLCJzb3VyY2VzIjpbIl9sYXlvdXQuc3ZlbHRlIl0sInNvdXJjZXNDb250ZW50IjpbIjxzY3JpcHQ+XG5cdGltcG9ydCBOYXYgZnJvbSAnLi4vY29tcG9uZW50cy9OYXYuc3ZlbHRlJztcblxuXHRleHBvcnQgbGV0IHNlZ21lbnQ7XG48L3NjcmlwdD5cblxuPHN0eWxlPlxuXHRtYWluIHtcblx0XHRwb3NpdGlvbjogcmVsYXRpdmU7XG5cdFx0bWF4LXdpZHRoOiA1NmVtO1xuXHRcdGJhY2tncm91bmQtY29sb3I6IHdoaXRlO1xuXHRcdHBhZGRpbmc6IDJlbTtcblx0XHRtYXJnaW46IDAgYXV0bztcblx0XHRib3gtc2l6aW5nOiBib3JkZXItYm94O1xuXHR9XG48L3N0eWxlPlxuXG48TmF2IHtzZWdtZW50fS8+XG5cbjxtYWluPlxuXHQ8c2xvdD48L3Nsb3Q+XG48L21haW4+Il0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQU9DLElBQUksZUFBQyxDQUFDLEFBQ0wsUUFBUSxDQUFFLFFBQVEsQ0FDbEIsU0FBUyxDQUFFLElBQUksQ0FDZixnQkFBZ0IsQ0FBRSxLQUFLLENBQ3ZCLE9BQU8sQ0FBRSxHQUFHLENBQ1osTUFBTSxDQUFFLENBQUMsQ0FBQyxJQUFJLENBQ2QsVUFBVSxDQUFFLFVBQVUsQUFDdkIsQ0FBQyJ9 */
✔ service worker (331ms)
✔ server (1.3s)
> Listening on http://localhost:3002
```

on Node v14.5.0 and macOS 10.15.6.

benmccann:
> The only change (besides upgrading all packages to their latest versions) is

You would also need to update the code to use `rollup-plugin-postcss` as shown in https://github.com/sveltejs/sapper-template/pull/260. It doesn't look like https://github.com/babichjacob/sapper-rm-css-repro has that plugin installed

babichjacob:
> You would also need to update the code to use `rollup-plugin-postcss` as shown in [sveltejs/sapper-template#260](https://github.com/sveltejs/sapper-template/pull/260). It doesn't look like https://github.com/babichjacob/sapper-rm-css-repro has that plugin installed

🤦 Whoops, now things work (at least in that repro).

benmccann:
We've developed a better way of handling css in https://github.com/sveltejs/sapper/pull/1508. We should still split it out into a separate plugin, but I'm going to close this for now

PR #1484, opened 2020-09-03
Allow building Sapper on Windows

benmccann:
Someone reported on Discord that they can't build Sapper because `rm -rf` doesn't work

How is anything passing on the Windows build without this?

https://github.com/sveltejs/sapper/blob/7af69e90d2db5c23247fc1de90ca0fa1a20a5bdb/.github/workflows/ci.yml#L9

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -968,6 +968,17 @@
         "ssri": "^6.0.1",
         "unique-filename": "^1.1.1",
         "y18n": "^4.0.0"
+      },
+      "dependencies": {
+        "rimraf": {
+          "version": "2.7.1",
+          "resolved": "https://registry.npmjs.org/rimraf/-/rimraf-2.7.1.tgz",
+          "integrity": "sha512-uWjbaKIK3T1OSVptzX7Nl6PvQ3qAGtKEtVRjRuazjfL3Bx5eI409VZSqgND+4UNnmzLVdPj9FqFJNPqBZFve4w==",
+          "dev": true,
+          "requires": {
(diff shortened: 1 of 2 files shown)


HansChua:
It seems that the Windows runners includes the Git bash in the path by default. https://github.com/actions/virtual-environments/issues/30#issuecomment-634028154

Conduitry:
It looks like the rebase lost actually using rimraf for these deletions.

benmccann:
oops. fixed!

Conduitry:
Can we use rimraf@2, which is already a transitive dependency, rather than adding a new dependency and making npm install several additional copies of rimraf@2?

Conduitry:
Eh. Maybe that doesn't matter so much.

```
$ npm ls rimraf
sapper@0.28.8 /home/conv/dev/sveltejs/sapper
├─┬ eslint@7.6.0
│ └─┬ file-entry-cache@5.0.1
│   └─┬ flat-cache@2.0.1
│     └── rimraf@2.6.3
├─┬ puppeteer@5.2.1
│ └── rimraf@3.0.2
└─┬ webpack@4.44.1
  └─┬ terser-webpack-plugin@1.4.5
    └─┬ cacache@12.0.4
      ├─┬ move-concurrently@1.0.1
      │ ├─┬ copy-concurrently@1.0.5
      │ │ └── rimraf@2.6.3  deduped
      │ └── rimraf@2.6.3  deduped
      └── rimraf@2.6.3  deduped
```

We already have v3 via Puppeteer.

Issue #1486, opened 2020-09-04
CSS is missing when importing one route from another


sfriedel:
**Describe the bug**
Since 0.28.3 when having a route that imports and renders the component of another route, the resulting page will not include the CSS of the imported route.

**To Reproduce**
Repo: https://github.com/sfriedel/sapper-bug-repro

This is a fresh repo created from the sapper template. I've only added a `index2.svelte` route to demonstrate the problem.

* Run sapper dev server
* Go to http://localhost:3000/
* Go to http://localhost:3000/index2 and compare with previous page

You will see that on index2 the layout changes because the CSS from `index.svelte` does not get loaded.

**Expected behavior**
http://localhost:3000/ and http://localhost:3000/index2 should look identical. 

**Information about your Sapper Installation:**

  System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Memory: 331.77 MB / 15.34 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.9.0 - /usr/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 6.14.8 - /usr/bin/npm
    Watchman: 4.9.0 - /home/sef/.local/opt/watchman/bin/watchman
  Browsers:
    Chrome: 85.0.4183.83
    Firefox: 80.0
  npmPackages:
    rollup: ^2.3.4 => 2.26.10
    sapper: ^0.28.0 => 0.28.4
    svelte: ^3.17.3 => 3.24.1

**Severity**
This is blocking me from upgrading beyond 0.28.2 in 2 projects.

**Additional context**
My usecase for importing routes from other routes is multilinguality. The projects where I use this have their primary language on the root and then have subtrees for each other supported language (like `example.com/de/...`, `example.com/fr/...`). The language prefixed routes simply import and use the component of the top-level route, e.g. `routes/de/index.svelte` imports the `routes/index.svelte` component and uses it. Until 0.28.2 this was working very well.

Hope this can be fixed and thank you for your awesome work on sapper :+1:

benmccann:
Would https://github.com/sveltejs/sapper/pull/1482 fix this for you?

sfriedel:
[...]

PR #1489, opened 2020-09-06
Fix CSS issue due to circular dependency in Sapper

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1486

Fixes the issue seen on svelte.dev as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # sapper changelog
 
+## Unreleased
+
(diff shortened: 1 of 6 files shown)


Conduitry:
Do you have a test case for whatever the situation is here that's causing the bug? Even if we're thinking about removing a lot of this code (or, indeed, especially if we're thinking about removing a lot of this code), it'd be good to have more tests against this reappearing.

Also please use `!==` rather than `!=`.

benmccann:
Yes. Good call. I've made those changes

Issue #1493, opened 2020-09-07
Svelte REPL broken on svelte.dev


benmccann:
`codemirror.css` is not being included. It looks like `Repl.js` dynamically imports `codemirror.js` which imports `codemirror.css`. In 0.28.0 it got stuck into `main.css`

benmccann:
Some of the CodeMirror CSS is showing up and some is missing. E.g. the css below is missing it missing

```
.CodeMirror {
    --background: var(--back-light);
    --base: hsl(45, 7%, 45%);
    --comment: hsl(210, 25%, 60%);
    --keyword: hsl(204, 58%, 45%);
    --function: hsl(19, 67%, 45%);
    --string: hsl(41, 37%, 45%);
    --number: hsl(102, 27%, 50%);
    --tags: var(--function);
    --important: var(--string);
    height: 300px;
    direction: ltr;
}
```

The missing css looks like it's in `node_modules/@sveltejs/svelte-repl/src/codemirror.css`. And then `node_modules/@sveltejs/svelte-repl/src/codemirror.js` contains:

```
import './codemirror.css';
```

benmccann:
I think this file is supposed to be discovered via the following path: `Repl.js` imports `CodeMirror.svelte` imports `codemirror.js` imports `codemirror.css`

benmccann:
The trigger for the problem is that the index route imports the REPL and the repl route dynamically imports the REPL. As a result, `codemirror.js` has already been added to `processed_chunks` and isn't seen as a leftover chunk from a dynamic import.

It seems the original algorithm of just sticking the leftover stuff in the application entry chunk isn't going to work. We could probably solve it by making the existing algorithm even more complicated:
* Find all dynamically imported chunks that are not routes
* Calculate the transitive dependency tree of each dynamically imported route
* Include the css for each of these trees in the application entry chunk

The reason this had not been reported in previous versions of Sapper is because Sapper duplicated all CSS into `main.css` (https://github.com/sveltejs/sapper/issues/1076 https://github.com/sveltejs/sapper/issues/1127)

PR #1494, opened 2020-09-08
Handle dynamic css imports that are also statically imported

benmccann:
Fixes https://github.com/sveltejs/sapper/issues/1493. See the issue for a description of the change

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -144,15 +144,17 @@ export default class RollupCompiler {
 				}
 
 				const css_for_chunk = {};
-				const processed_chunks: Set<RenderedChunk> = new Set();
+				const global_chunks: Set<RenderedChunk> = new Set();
 
 				/**
 				 * Creates a single CSS chunk for the given JS chunks
 				 */
-				function handle_chunks(tree_entry_chunk: RenderedChunk, chunks: Iterable<RenderedChunk>) {
+				function handle_chunks(tree_entry_chunk: RenderedChunk, chunks: Iterable<RenderedChunk>, subtree?: boolean) {
 					const css_modules: Set<string> = new Set();
 					for (const chunk of chunks) {
-						processed_chunks.add(chunk);
+						if (!subtree) {
+							global_chunks.add(chunk);
+						}
 						Object.keys(chunk.modules).filter(k => k.endsWith('.css')).forEach(m => css_modules.add(m));
 					}
 					if (css_modules.size) {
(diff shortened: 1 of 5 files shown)


Conduitry:
Can we add a test for this as well? These were obviously weird edge cases that were never considered, and it would be nice to have some ability to guard against them going forward.

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -167,37 +169,45 @@ export default class RollupCompiler {
...
 					}
+				}
 
-					// Put everything that's leftover into the entry chunk in order to include css
-					// that is in a dynamically imported chunk
-					const chunks_for_tree = js_deps(tree_entry_chunk,
-						{ walk: ctx => !subtree || ctx.chunk.fileName !== entry_chunk.fileName });
-					const unused = new Set(chunks_for_tree.filter(x => !processed_chunks.has(x)));
-					handle_chunks(tree_entry_chunk, unused);
+				function isRoute(file_path: string) {
```
Conduitry, line 180: `is_route` please, to keep with the snake case used elsewhere.

benmccann: done

benmccann:
Ok, I updated the tests so that they fail on `master` and pass with this PR. I didn't need to write a new test but just added a new route that would cause existing tests to fail

Conduitry:
Oh nice, I see that this also makes `rollup-dependency-tree` a dev/bundled dependency.

benmccann:
My reason for moving it was that it's only used in building the application by Rollup / the Sapper build process and so I was thinking it was a dev dependency. E.g. if someone runs `npm ci --prod` then I figured this would stop it from being a dependency that's shipped to production. It works just fine either place I have it and I don't have much of an opinion on where to put it. Were you thinking I should move it back?

Conduitry:
No no, it's nice having it be bundled in instead. My understanding from the conversation [here](https://github.com/sveltejs/sapper/pull/1452#discussion_r479437894) was that you were running into some issues when doing that, but it didn't seem very worthwhile to try to get to the bottom of it, because Sapper itself is only a dev dependency of apps that are written in it, and so built apps didn't have any additional dependencies.

I'm a bit confused by parts of your last comment. I would expect that the only difference for someone installing `sapper` into their app would be whether they get a bundled copy of `rollup-dependency-tree` or their package manager has to separately request and install it.

benmccann:
I think we may have been talking about different things when you said "external" in https://github.com/sveltejs/sapper/pull/1452#discussion_r479437894. I originally thought you meant was there a reason to make it a library and get it from npm instead of just putting the source code in Sapper. But now I understand you mean `devDependencies` vs not, so I was answering a bit of a different question on the other thread.

Feel free to ignore my last comment too :smile: I get a bit confused sometimes about Sapper being both a builder and a runtime in the same project

Anyway, I don't think there's much difference whether this is in `devDependencies` or `dependencies`, so I'm happy to have it in either place

PR #1498, opened 2020-09-09
Store `error` in page store (when an error page is triggered)

### LINKED ISSUE #948: Make layout know if it's in error page or home page

ehrencrona:
To fix #948, store the current error in the `page` store to allow the layout to know whether the current page is an error page (as per [suggestion](https://github.com/sveltejs/sapper/pull/962#issuecomment-637513569) by @Conduitry).

My first PR in this project, so very open to feedback :) 

### Tests

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -18,9 +18,9 @@ import {
 	HydratedTarget,
 	Target,
 	Redirect,
+	Branch,
 	Page,
-	PageContext,
-	Branch
+	PageContext
 } from './types';
 import goto from './goto';
 import { page_store } from './stores';
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -57,7 +57,7 @@ stores.session.subscribe(async value => {
 	if (redirect) {
 		await goto(redirect.location, { replaceState: true });
 	} else {
-		await render(branch, props, dest.page);
+		await render(branch, props, Object.assign({error: props.error}, dest.page));
 	}
 });
 
@@ -110,7 +110,7 @@ function handle_error(url: URL) {
 
 	};
 	const query = extract_query(search);
-	render([], props, { host, path: pathname, query, params: {} });
+	render([], props, { host, path: pathname, query, params: {}, error });
 }
 
 async function handle_target(dest: Target): Promise<void> {
@@ -127,7 +127,7 @@ async function handle_target(dest: Target): Promise<void> {
 		await goto(redirect.location, { replaceState: true });
 	} else {
 		const { props, branch } = hydrated_target;
-		await render(branch, props, dest.page);
+		await render(branch, props, Object.assign({error: props.error}, dest.page));
 	}
 }
 
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -18,9 +18,9 @@ import {
 	HydratedTarget,
 	Target,
 	Redirect,
-	Branch,
 	Page,
-	PageContext
+	PageContext,
+	Branch
 } from './types';
 import goto from './goto';
 import { page_store } from './stores';
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -112,14 +112,14 @@ function handle_error(url: URL) {
 
 	};
 	const query = extract_query(search);
-	render([], props, { host, path: pathname, query, params: {}, error, status });
+	render([], props, { host, path: pathname, query, params: {}, error });
 }
 
 
 function buildPageContext(props: any, page: Page): PageContext {
-  const { error, status } = props;
+  const { error } = props;
 
-  return { error, status, ...page };
+  return { error, ...page };
 }
 
 async function handle_target(dest: Target): Promise<void> {
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -18,7 +18,9 @@ import {
 	HydratedTarget,
 	Target,
 	Redirect,
-	Page
+	Page,
+	PageContext,
+	Branch
 } from './types';
 import goto from './goto';
 import { page_store } from './stores';
@@ -57,7 +59,7 @@ stores.session.subscribe(async value => {
 	if (redirect) {
 		await goto(redirect.location, { replaceState: true });
 	} else {
-		await render(branch, props, Object.assign({error: props.error}, dest.page));
+		await render(branch, props, buildPageContext(props, dest.page));
 	}
 });
 
@@ -110,7 +112,14 @@ function handle_error(url: URL) {
 
 	};
 	const query = extract_query(search);
-	render([], props, { host, path: pathname, query, params: {}, error });
+	render([], props, { host, path: pathname, query, params: {}, error, status });
+}
+
+
+function buildPageContext(props: any, page: Page): PageContext {
+  const { error, status } = props;
+
+  return Object.assign({ error, status }, page);
 }
 
 async function handle_target(dest: Target): Promise<void> {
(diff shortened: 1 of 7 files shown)


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -237,7 +237,8 @@ export function get_page_handler(
 							host: req.headers.host,
 							path: req.path,
 							query: req.query,
-							params
+							params,
+							error
```
benmccann, line 241: It looks to me like this could also be a `string` here. If you look 10 lines down then you see how we standardize the `error` to an `Error` type. We should probably do that earlier and use the standardized error in both places

Should we also return `status` here as well?

benmccann (commented):
thanks! And thanks for all your work on the prettier plugin too!

runtime/src/app/app.ts
```diff
@@ -54,7 +54,7 @@ stores.session.subscribe(async value => {
 	if (redirect) {
 		await goto(redirect.location, { replaceState: true });
 	} else {
-		await render(branch, props, dest.page);
+		await render(branch, props, Object.assign({error: props.error}, dest.page));
```
ehrencrona, line 57: I wasn't sure if it was ok to use modern JS here; would it be ok to switch to the spread operator instead?

runtime/src/server/middleware/get_page_handler.ts
ehrencrona: > It looks to me like this could also be a string here. 

True; I didn't see the conversion code. I think I'd prefer relaxing the type of `error` to `any`. It's consistent with how TypeScript typically types errors and there is generally [no guarantee that what you `catch` is of any particularly type](https://github.com/microsoft/TypeScript/issues/8677#issuecomment-220385124), so `Error | string` isn't correct.

The conversion is also a bit funky: `{ message: error }` is not an instance of `Error`.

I find only two places where a string is thrown: [on a 404](https://github.com/sveltejs/sapper/blob/master/runtime/src/server/middleware/get_page_handler.ts#L371) where the string is `Not found` and [on an error in preload](https://github.com/sveltejs/sapper/blob/master/runtime/src/server/middleware/get_page_handler.ts#L213) that gets converted into just the message. I'd propose changing them to `new Error('Not found')` and the actual caught error from preload rather than just the message.

This is technically a breaking change, but it's highly unlikely anyone depends on this behavior. 

Then we could also remove `|| new Error('Unknown error in preload function')` [in `handle_error`](https://github.com/sveltejs/sapper/blob/master/runtime/src/server/middleware/get_page_handler.ts#L43) that doesn't seem right, given that the function is called from other places than just `preload`. 

Should I go ahead with the changes described above?

> Should we also return `status` here as well?

It doesn't seem particularly useful to the layout, but on the other hand it does go together with the `error` object in a way. So we probably should. I'll add it.

(Though I think I would have preferred the status code as a property on a specialized error subclass, so we could throw e.g. `NotFoundError` and then `e.status == 404`, but that's a bigger change)

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -230,24 +231,28 @@ export function get_page_handler(
 				error.stack = sourcemap_stacktrace(error.stack);
 			}
 
+			const pageContext: PageContext = {
+				host: req.headers.host,
+				path: req.path,
+				query: req.query,
+				params,
+				error: error ? error instanceof Error ? error : { message: error } : null,
```
benmccann, line 239: I think to keep the types simpler and more consistent I would change this to `{ name: 'PreloadError', message: error }` and then change the type of `PageContext.error` to be just `Error` instead of `Error | { message: string; }`

ehrencrona: Ok, sure! Please also check my comments and suggestions above on this.

runtime/src/server/middleware/get_page_handler.ts
benmccann: I'd probably rather we continue to accept a `string`. It's both a bit more concise for many use cases and avoids making this a breaking change. I think that we can coalesce it to a single `Error` type later in the code as mentioned below.

Good catch on the fallback error message. It should probably be something like "Unknown error" instead of "Unknown error in preload function". Hopefully that error would never actually be shown - the developer would have to pass an empty string or similar which I would hope no one would do

runtime/src/app/app.ts
benmccann: I think it should be because this is just for modern browsers and legacy browsers should go through babel. Surprisingly I'm not seeing it done anywhere in the codebase except for server-side code

runtime/src/server/middleware/get_page_handler.ts
ehrencrona: Ok, I've updated the message.

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -230,24 +231,32 @@ export function get_page_handler(
 				error.stack = sourcemap_stacktrace(error.stack);
 			}
 
+			const pageContext: PageContext = {
+        host: req.headers.host,
```
benmccann, line 235: there are spaces instead of tabs here

ehrencrona: Oops. Updated.

benmccann (approved):

runtime/src/app/types.ts
```diff
@@ -4,9 +4,15 @@ export interface HydratedTarget {
 	redirect?: Redirect;
 	preload_error?: any;
 	props: any;
-	branch: Array<{ Component: DOMComponentConstructor, preload: (page) => Promise<any>, segment: string }>;
+	branch: Branch;
 }
 
+export type Branch = Array<{
```
benmccann, line 10: this type is a bit different from the one defined in https://github.com/sveltejs/sapper/pull/1528/files#diff-d1d1a1a1a8896e90a29cc23a07d69ed6R10

site/content/docs/07-state-management.md
```diff
@@ -14,7 +14,7 @@ Inside a component, get references to the stores like so:
 ```
 
 * `preloading` contains a readonly boolean value, indicating whether or not a navigation is pending
-* `page` contains a readonly `{ host, path, params, query }` object, identical to that passed to `preload` functions
+* `page` contains a readonly `{ host, path, params, query, error, status }` object. The same value is passed to `preload` functions, though `status` is not present at that time and `error` is only set on error pages.
```
benmccann, line 17: I'm wondering now if my suggestion to add `status` maybe wasn't a good one. I hadn't thought through the fact that it wouldn't be available yet in `preload`. I wonder if it's that useful in this case and maybe that's too much of a caveat to actually add it

benmccann:
Looks like this will need to be rebased since your other PR was merged

site/content/docs/07-state-management.md
ehrencrona: Yeah, I wouldn't mind removing it again. I generally don't find it particularly useful and the concept of an http status code doesn't really make sense on the client side. I'll take it out again.

runtime/src/app/types.ts
ehrencrona: i updated it with the types from the typings PR now.

PR #1508, opened 2020-09-14
Inject CSS loading with Rollup

benmccann:
This has rollup inject css loading into the js files instead of putting it into `manifest-client` and having the application load it from there. Thanks to Rich for writing most of this code!


Conduitry:
As of bd02d865e8bb18e5c85daa17a52d0725c400f9c7 I'm seeing lots of missing styles on svelte.dev in production builds when using this.

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -97,84 +130,90 @@ export default class RollupCompiler {
...
-				// rollup-plugin-svelte adds an import statement to the js file which references the css file
-				// that won't be able to be compiled as js, so we remove it here and store a copy to use later
-				if (/\.css$/.test(id)) {
-					that.css_files[id] = code;
-					return {code: '', moduleSideEffects: 'no-treeshake'};
+			renderDynamicImport({ format, moduleId, targetModuleId, customResolution }) {
+				if (targetModuleId) {
+					return {
+						left: 'Promise.all([import(',
+						right: `), /* ___${targetModuleId}___ */]).then(x => x[0])`
```
Conduitry, line 137: I think I'd be more comfortable if this has a slightly more unique string in the comment.

Conduitry:
For posterity: It was discovered that what is going on in production builds is that Terser is removing the `/* ___foo___ */` comments, which is breaking the handling of styles. A suggestion was to use unknown global variables for these placeholders instead. A hacky but seemingly working implementation of this is:

```diff
diff --git a/src/core/create_compilers/RollupCompiler.ts b/src/core/create_compilers/RollupCompiler.ts
index 806dd7c..d0ae434 100644
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -127,14 +127,14 @@ export default class RollupCompiler {
 					entry_point = inputs[0].file;
 				}
 			},
-			renderChunk(code: string, chunk: RenderedChunk) {	
+			renderChunk(code: string, chunk: RenderedChunk) {
 				that.chunks.push(chunk);
 			},
 			renderDynamicImport({ format, moduleId, targetModuleId, customResolution }) {
 				if (targetModuleId) {
 					return {
 						left: 'Promise.all([import(',
-						right: `), /* ___${targetModuleId}___ */]).then(x => x[0])`
+						right: `), ___SAPPER___${Buffer.from(targetModuleId).toString('hex')}___]).then(x => x[0])`
 					};
 				} else {
 					return {
@@ -176,7 +176,8 @@ export default class RollupCompiler {
 					let chunk_has_css = false;
 
 					if (chunk.code) {
-						chunk.code = chunk.code.replace(/\/\* ___(.+?)___ \*\//g, (m, id) => {
+						chunk.code = chunk.code.replace(/___SAPPER___([0-9a-f]+)___/g, (m, id) => {
+							id = Buffer.from(id, 'hex').toString();
 							const target = <OutputChunk>Object.values(bundle).find(c => (<OutputChunk>c).facadeModuleId === id);
 
 							if (target) {
```

src/core/create_compilers/RollupCompiler.ts
benmccann: updated this with the fix for terser compatibility

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -69,13 +88,27 @@ export default class RollupCompiler {
...
 		let entry_point: string | undefined;
 
 		const that = this;
-		const sourcemap = mod.output.sourcemap;
 
-		// TODO this is hacky, and doesn't need to apply to all three compilers
-		(mod.plugins || (mod.plugins = [])).push({
+		// TODO this is hacky. refactor out into an external rollup plugin
+		(mod.plugins || (mod.plugins = [])).push(css_chunks({ injectImports: true }));
+		if (!JSON.stringify(mod.input).includes('/client.')) {
```
Rich-Harris, line 108: wouldn't this mean that any chunks that were dynamically imported *not* from the manifest wouldn't have the `import(...)` rewriting logic applied? e.g. if a page lazily loads some component upon interaction

benmccann: This line is pretty hacky and could probably be done in a cleaner way, but basically I'm saying to apply it only to the client bundle. I was thinking that you can't apply the rewriting in the server bundle because `document` won't exist there, right?

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -69,13 +88,27 @@ export default class RollupCompiler {
...
 		let entry_point: string | undefined;
 
 		const that = this;
-		const sourcemap = mod.output.sourcemap;
 
-		// TODO this is hacky, and doesn't need to apply to all three compilers
-		(mod.plugins || (mod.plugins = [])).push({
+		// TODO this is hacky. refactor out into an external rollup plugin
+		(mod.plugins || (mod.plugins = [])).push(css_chunks({ injectImports: true }));
+		if (!JSON.stringify(mod.input).includes('client.')) {
```
Conduitry, line 108: My attention was drawn here because it was the line involved in the fix for Windows - and the new version of it makes me even more uncomfortable than the old version.

Why are we looking at `JSON.stringify(mod.input)`? Are there multiple fields within `mod.input` where this filename could occur? And is there a more precise regex we could use here, like `/\/client\/client\.\w+\.js$/` or something?

benmccann: yeah, this whole section is a bit icky. it wasn't the "TODO this is hacky" comment that brought your attention here? :smile: 

ok, I just updated it to be a bit better

I hope that we can refactor this out into an external plugin (either make it part of `rollup-plugin-css-chunks` or a new one) very shortly which would make this go away

Conduitry:
I'm trying to tidy this up a little bit by having `inject_styles.mjs` be a file that lives in `runtime/internal/` and is copied over to `src/node_modules/@sapper/internal/` and is imported from `@sapper/internal/inject_styles` (rather than manually generating and emitting it during the bundling) but I'm running into a couple of weird test failures. Is this something that you'd already explored and then rejected?

benmccann:
I think the CSS code splitting stuff should be refactored out into an external Rollup plugin. Possibly made part of `rollup-plugin-css-chunks` (the author of that plugin seems open to the idea). I think that's much easier to do with the code how it currently is than if we were to move it

PR #1512, opened 2020-09-14
Proposal for configurable error handling

### LINKED ISSUE #742: Custom `internal server error` messages
Sapper will currently bail with `<pre>Internal server error</pre>` when errors are thrown while rendering error pages.

This was done to prevent the `node` process from hanging in an infinite loop, but we should probably look into handling it more gracefully.

At the very least, we could support customizing the error with a static `template-internal-error.html` akin to `template.html` when nothing else can be done.

Thoughts?

### LINKED ISSUE #812: Proper error handling for server routes instead of sending all errors as a text response

### LINKED ISSUE #955: Allow server side redirect on 404 / 500 error.
I can't figure out how to redirect to the index page on a `404 error` without using client side routing after `onMount`. 

A way for the server to detect a status code and redirect to the desired route . 

Perhaps implementing some sort of `preload`  `this.redirect` in the `_error.svelte` template that also has access to the status code so the redirect can be based on the status code.

I think for certain apps, i.e. a timesheet app for logging employee time, having all 404 routes lead to the index (which is the timesheet page) leads to a cleaner feeling app. On some apps, going to a 404 and having to click out of it can feel clunky or like something is broken (which in a way is), but redirecting to main page makes the app feel more polished.

### LINKED ISSUE #993: sapper.middleware catches all server errors, would rather pass through

ehrencrona:
This is a proposal that fixes #955, fixes #812, fixes #993, fixes #960, and fixes #742 

The suggestion is to allow for an `onError` parameter to the `middleware` function. `onError` is a function that is notified on any server-side error. It receives the request, the status code, the error and whether the error was in a server route or a page.

It can be used to log errors on the server side, e.g. in Sentry, but it can also be used to replace the standard error response, if desired. To do this, it receives a function called `customizeResponse`. If called, the response can be modified and the standard response (which is either `_error.svelte` for pages or a plain JSON response for server routes) is not generated.

See the tests for an [example of how to use `onError`](https://github.com/sveltejs/sapper/blob/f6417d42ee690a4d633bf91186b407b0326c13c1/test/apps/errors/src/server.js#L9)

I added tests, but did not update documentation, since I assume there will be change requests and it's better to update it if/when there is agreement on the functionality.

This is quite similar to #813 but a bit more flexible in that it
 * works for both server routes and pages
 * can be explicitly used to just listen to errors rather than modifying the response
 * adds tests :)

### Tests

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -13,7 +13,7 @@ import { OnError, sendErrorResponse, RouteType } from './on_error';
 
 export function get_page_handler(
 	manifest: Manifest,
-	session_getter: (req: Req, res: Res) => Promise<any>, 
+	session_getter: (req: Req, res: Res) => Promise<any>,
 	onError?: OnError
 ) {
 	const get_build_info = dev
@@ -32,10 +32,10 @@ export function get_page_handler(
 		console.error(error);
 
 		sendErrorResponse({
-			
-			defaultResponse: () => {				
+
+			defaultResponse: () => {
 				const message = dev ? escape_html(error.message) : 'Internal server error';
-				
+
 				res.statusCode = 500;
 				res.end(`<pre>${message}</pre>`);
 			},
@@ -55,7 +55,7 @@ export function get_page_handler(
 					parts: [
 						{ name: null, component: { default: error_route } }
 					]
-				}, req, res, statusCode, error);	
+				}, req, res, statusCode, error);
 			},
 			routeType: RouteType.Page,
 			statusCode: 500,
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -9,10 +9,12 @@ import { sourcemap_stacktrace } from './sourcemap_stacktrace';
 import { Manifest, ManifestPage, Req, Res, build_dir, dev, src_dir } from '@sapper/internal/manifest-server';
 import { PreloadResult } from '@sapper/internal/shared';
 import App from '@sapper/internal/App.svelte';
+import { OnError, sendErrorResponse, RouteType } from './on_error';
 
 export function get_page_handler(
 	manifest: Manifest,
-	session_getter: (req: Req, res: Res) => Promise<any>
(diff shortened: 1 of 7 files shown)


benmccann:
Thanks for this!

This one might take a little while to get in. Most PRs that add some type of configuration are currently on hold while we figure out exactly what we want to do about configuration. If you ping me on Discord (I have the same username there), I can share some more details

The test failures might be real on this one and should probably be looked at

ehrencrona:
> The test failures might be real on this one and should probably be looked at

They were. I'd got used to ignoring any timeouts. :( Fixed it.

soullivaneuh:
Is there anything to do to get this feature merged?

benmccann:
It's unlikely we'll be merging many new features to this repo. We're focused on [SvelteKit](https://svelte.dev/blog/whats-the-deal-with-sveltekit), which is a successor to Sapper containing much of Sapper's code that folks will be able to migrate to without too much difficulty

benmccann:
Thanks again for this PR! Unfortunately we never really had an agreed upon way of handling configuration in Sapper which made it hard to get this in. But that's been fixed in SvelteKit with the addition of `svelte.config.js` and adapters, so we'd be happy to help make sure this is possible in SvelteKit if you want to take a stab at it there

PR #1513, opened 2020-09-14
Support IDE-friendly custom extensions for sapper, extensions with multiple "." dots

samdesota:
- [ ] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs

### Tests

TL;DR there's a bug / feature missing in the sapper route walking logic that means you can't use custom extensions with multiple "." dots in them ex. ".route.svelte"

Sometimes it's nice to keep some components that are only used within a single route under that routes folder, so for example if you use `ProductItem.svelte` under `routes/products/index.svelte` it may make sense to keep it under `routes/products/ProductItem.svelte`. Of course, that will also create a route ProductItem.svelte which isn't what we want.

Sapper supports custom extensions, so you could use `.route` so you have `products/ProductItem.svelte` and `product/index.route`, now you can safely keep components wherever you'd like. However, at our organization we've noticed that having svelte files not ending with `.svelte` causes a number of issues for IDEs and prettier for example, usually this can be resolved with additional configuration but it's ideal and clearer if everything uses the same extension, so instead of using `.route` we wanted to use `.route.svelte` to have our cake and eat it too. 

However svelte identified routes using the custom extensions by using `path.extname`, which only returns the last extension segment, meaning we couldn't use extensions with multiple "." dots. In this PR I changed the logic to use `basename.endsWith` instead.

Thank you very much, let me know if there's any additional reasons that this feature may cause an issue or if there's any feedback on the PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ __sapper__
 dist
 /runtime/app.mjs
 /runtime/server.mjs
+/.vscode/settings.json
--- a/src/core/create_manifest_data.ts
+++ b/src/core/create_manifest_data.ts
@@ -57,15 +57,19 @@ export default function create_manifest_data(cwd: string, extensions: string = '
 				const file = path.relative(cwd, resolved);
 				const is_dir = fs.statSync(resolved).isDirectory();
 
-				const ext = path.extname(basename);
+				const file_ext = path.extname(basename);
 
 				if (basename[0] === '_') return null;
 				if (basename[0] === '.' && basename !== '.well-known') return null;
-				if (!is_dir && !/^\.[a-z]+$/i.test(ext)) return null; // filter out tmp files etc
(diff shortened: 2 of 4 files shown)


Conduitry:
This is released in 0.28.10.

paulovieira:
> it's nice to keep some components that are only used within a single route under that routes folder, so for example if you use `ProductItem.svelte` under `routes/products/index.svelte` it may make sense to keep it under `routes/products/ProductItem.svelte`. Of course, that will also create a route ProductItem.svelte which isn't what we want.

This PR is useful, thanks for your contribution!

I would just like to add that it's also possible to have a specific component next to the route component and without a new route being created.

That is, we can have:

- `routes/products/index.svelte`
- `routes/products/_components/ProductItem.svelte`

Since the directory starts with the underscore, sapper will not created routes for those components:

https://sapper.svelte.dev/docs#File_naming_rules

PR #1516, opened 2020-09-15
Increase timeouts on slow tests

ehrencrona:
Certain tests seem to frequently time out on CI. Raise the timeout of those tests.


antony:
I feel like finding the root cause of this timeout would be the optimal solution here, but there are reports we can generate regarding test speed which would probably be the way to do this, so, sure.

PR #1519, opened 2020-09-15
Update 11-deploying.md

markbrocato:
Adds a link to the guide for deploying on the Moovweb XDN. This is just a documentation change.

Full Disclosure:
I'm the head of the platform team at Moovweb.  We love svelte and sapper!  We're actually building our developer tools in svelte right now. Sapper was one of the easiest frameworks to get working on our serverless platform, the Moovweb XDN. The XDN makes it easy to deploy sapper apps to super fast infrastructure.  We'd love to see the XDN listed along side Vercel as an alternative.


pngwn:
Have you considered renaming your product to moonweb? I think it could really work for you.

pngwn:
We have discussed this and decided that this isn't a trend we would like to see continue, we don't really want to be advertising deployment services on the docs. Vercel is mainly there for historic reason but its removal is also being discussed.

PR #1522, opened 2020-09-17
Cleaned up documentation on `session`

ehrencrona:
I didn't understand the purpose of `session` the first time I read the documentation, so I wanted to make that a little bit clearer. I found it particularly confusing that the first reference you find is "preloading" where there is only a slightly cryptic code sample and no reference to the main section that describes sessions. But also, importantly, the actual purpose of the session was never explicitly stated.

I also did some stylistic changes in that chapter of the documentation while I was at it.

-  [x] lint the project with `npm run lint`

### DIFF
--- a/site/content/docs/07-state-management.md
+++ b/site/content/docs/07-state-management.md
@@ -22,14 +22,13 @@ A component can retrieve the stores using the `stores` function:
 
 `session` is `undefined` by default. To populate it with data, implement a function that returns session data given an HTTP request. Pass it as an option to `sapper.middleware` when the server is started.
 
-As an example, here is how to populate the session with the current user:
+As an example, let's look at how to populate the session with the current user. First, add the `session` parameter to the Sapper middleware:
 
 ```js
 // src/server.js
-express()
+polka()
 	.use(
-		serve('static'),
-		authenticationMiddleware(),
+		// ...
 		sapper.middleware({
 			// customize the session
 			session: (req, res) => ({
@@ -37,12 +36,11 @@ express()
 			})
 		})
 	)
-	.listen(process.env.PORT);
 ```
 
 `req` is an `http.ClientRequest` and `res` an `http.ServerResponse`.
 
-The session data must be serializable. This means it must not contain functions or custom classes, just built-in JavaScript data types. [devalue](https://github.com/Rich-Harris/devalue) is used for serialization.
+The session data must be serializable. This means it must not contain functions or custom classes, just built-in JavaScript data types.
 
 The `session` function may return a `Promise` (or, equivalently, be `async`).
 
--- a/site/content/docs/07-state-management.md
+++ b/site/content/docs/07-state-management.md
@@ -14,7 +14,7 @@ A component can retrieve the stores using the `stores` function:
 ```
 
 * `preloading` contains a read-only boolean value, indicating whether or not a navigation is pending
-* `page` contains information on the current address. See [preloading](docs#Arguments) for details. This store is read-only.
+* `page` contains read-only information about the current route. See [preloading](docs#Arguments) for details.
 * `session` can be used to pass data from the server related to the current request. It is a [writable store](https://svelte.dev/tutorial/writable-stores), meaning you can update it with new data. If, for example, you populate the session with the current user on the server, you can update the store when the user logs in. Your components will refresh to reflect the new state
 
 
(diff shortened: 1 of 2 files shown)


benmccann:
looks like this PR will need to be rebased

ehrencrona:
Rebased!

site/content/docs/07-state-management.md
```diff
@@ -13,22 +13,25 @@ Inside a component, get references to the stores like so:
...
 ### Seeding session data
 
-On the server, you can populate `session` by passing an option to `sapper.middleware`:
+`session` is `undefined` by default. To populate it with data, implement a function that returns session data given an HTTP request. Pass it as an option to `sapper.middleware` when the server is started.
+
+As an example, here is how to populate the session with the current user:
 
 ```js
 // src/server.js
-express() // or Polka, or a similar framework
```
benmccann, line 27: I would probably leave this comment for now so that people don't get confused by this saying `express` when the template uses `polka`. or maybe change `express` to `polka` here to match the template

site/content/docs/07-state-management.md
```diff
@@ -13,22 +13,25 @@ Inside a component, get references to the stores like so:
 </script>
 ```
 
-* `preloading` contains a readonly boolean value, indicating whether or not a navigation is pending
-* `page` contains a readonly `{ host, path, params, query, error }` object. `error` is only set on error pages. This is the same value that is passed to `preload` functions.
-* `session` contains whatever data was seeded on the server. It is a [writable store](https://svelte.dev/tutorial/writable-stores), meaning you can update it with new data (for example, after the user logs in) and your app will be refreshed
+* `preloading` contains a read-only boolean value, indicating whether or not a navigation is pending
+* `page` contains information on the current address. See [preloading](docs#Arguments) for details. This store is read-only.
```
benmccann, line 17:
```suggestion
* `page` contains read-only information about the current route. See [preloading](docs#Arguments) for details.
```

site/content/docs/07-state-management.md
```diff
@@ -37,6 +40,10 @@ express() // or Polka, or a similar framework
 	.listen(process.env.PORT);
 ```
 
-The `session` option is a function which returns data, or a `Promise` of data, which must be serializable (using [devalue](https://github.com/Rich-Harris/devalue)), meaning that it must not contain functions or custom classes, just built-in JavaScript data types.
+`req` is an `http.ClientRequest` and `res` an `http.ServerResponse`.
+
+The session data must be serializable. This means it must not contain functions or custom classes, just built-in JavaScript data types. [devalue](https://github.com/Rich-Harris/devalue) is used for serialization.
```
benmccann, line 45: Maybe just make the reference to `devalue` a footnote. I would think most users shouldn't have to care about what's used internally

```suggestion
The session data must be serializable<sup>1</sup>. This means it must not contain functions or custom classes, just built-in JavaScript data types.
```

Then at the bottom of the page add:

1. [devalue](https://github.com/Rich-Harris/devalue) is used for serialization

site/content/docs/07-state-management.md
```diff
@@ -37,6 +40,10 @@ express() // or Polka, or a similar framework
...
+
+The `session` function may return a `Promise` (or, equivalently, be `async`).
 
-> Note that if you return a `Promise` from this function (or use an `async` function), it will be re-awaited for **every** server-rendered page route.
\ No newline at end of file
+> If `session` returns a `Promise`, every server-rendered page route will wait for it to resolve. This can slow down your application significantly.
```
benmccann, line 49: I prefer the old text here for a few reasons. It starts with "Note that" to show it's something to pay attention to, informed users that `async` functions were also affected by this, it bolded every to call attention to it, it didn't try to predict the impact to the user's application

It was a little verbose though and I think could be made slightly more concise:

```suggestion
> Note that if `session` returns a `Promise` (or is `async`), it will be re-awaited for on **every** server-rendered page route.
```

site/content/docs/07-state-management.md
ehrencrona: Oh. I was certainly confused by it. I thought this comment was part of the template code and wasn't relevant to the documentation. I'll switch to polka to match the sample then.

site/content/docs/07-state-management.md
ehrencrona: I'd suggest just removing it entirely. I agree it's an implementation detail.

site/content/docs/07-state-management.md
ehrencrona: Sure. I just found the bolding shouty. I'm honestly not sure what it is really warning of. I assumed it was talking about the potential performance impact. It would be helpful to spell out why the await is or can be a problem.

Issue #1505, opened 2020-09-12
Sapper legacy mode: CSS not loaded on route change


rodoch:
**Describe the bug**
This issue occurs when using a production build (`npm run build`) of Sapper created with the `--legacy` flag in a browser which relies on the legacy code (i.e. a browser with no native support for `async/await` syntax).

When navigating from Page A to page B the CSS associated with the page B route is not loaded and the contents of page B remain unstyled. This does not affect the styling of any components already loaded on page A.

**Logs**
When testing this in Chrome 54 the console displays the following errors:

```
service-worker.js:1 Uncaught SyntaxError: missing ) after argument list 
(index):1 Uncaught (in promise) TypeError: Failed to register ServiceWorker: ServiceWorker script evaluation failed
```

However, this appears to be unrelated to the problem. Removing the service worker from the project and rebuilding does not alleviate the issue.

**To Reproduce**
I've created a minimal repro here: https://github.com/rodoch/sapper-legacy-repro

This is simply a copy of the sapper-template repo with Sapper & Svelte versions bumped to 0.28.6 and 3.25.0 respectively. To reproduce the problem you need to use a 'legacy' browser. I have reproduced with Chrome  <= 54 and Safari <= 9 (via BrowserStack and on an old Windows machine).

1. Start by opening the URL `localhost:3000/blog`. This page is styled correctly:

![pic](https://user-images.githubusercontent.com/5305694/93002841-af365e80-f531-11ea-8b98-883f8d197552.png)

2. Next, navigate to the homepage. The content of this page is unstyled (e.g. the size of the success kid picture is unconstrained).

![pic](https://user-images.githubusercontent.com/5305694/93002853-cb3a0000-f531-11ea-9c14-dc0f1cc69c82.png)

This behaviour is not evident if you start on the homepage and move to the blog, presumably because a sufficient set of styles have been applied on initial page load when moving in this direction. With more complex apps this behaviour can be observed on all navigations subsequent to the initial page load.

**Expected behavior**
The relevant styles should be applied with each route change.

**Information about your Sapper Installation:**

- The output of `npx envinfo --system --npmPackages svelte,sapper,rollup,webpack --binaries --browsers`

```
 System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 5.08 GB / 15.88 GB
  Binaries:
[...]

PR #1523, opened 2020-09-17
Fix support for legacy browsers

benmccann:
Hopefully fixes https://github.com/sveltejs/sapper/issues/1505

I had some difficulty testing this. It fixes the original error, but I'm getting a new one that I think is unrelated and a result of the template not being setup to support legacy browsers by default. @rodoch @antony would you be able to test this one?


benmccann:
Nevermind. This doesn't work

PR #1524, opened 2020-09-17
Minify and hash inject_styles.js

benmccann:
`inject_styles` can be emitted as a `chunk` instead of an `asset` since it's a `.js` file. This allows it to be minified, which makes a better use of bandwidth. It also makes it have a hash in it's name, which is important for avoiding caching bugs. URLs under the `/client` path are cached. I potentially wanted to change the contents of `inject_styles` while coming up with a fix for https://github.com/sveltejs/sapper/issues/1505 (e.g. as in https://github.com/sveltejs/sapper/pull/1525), but that wouldn't work if the file is cached because users would get the old version of the file

Also, I think it would be nice to have `inject_styles` be processed by Babel. This seems like it would be necessary for that to happen. However, from my testing, it does not appear sufficient

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -15,6 +15,7 @@ import { CompileResult } from './interfaces';
 import RollupResult from './RollupResult';
 
 const stderr = console.error.bind(console);
+const INJECT_STYLES_ID = 'inject_styles';
 
 let rollup: any;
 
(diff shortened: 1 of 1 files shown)


ArekBartnik:
@benmccann page speed also points that inject_styles.js should be preloaded.

benmccann:
Ah, good call out. The work on this is never done :-p

I'm going to go head and merge this one to unblock https://github.com/sveltejs/sapper/pull/1525

PR #1525, opened 2020-09-17
IE 11 support

benmccann:
Closes https://github.com/sveltejs/sapper/issues/1505

Upgrades to the latest version of shimport including the fix: https://github.com/Rich-Harris/shimport/pull/32

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4923,9 +4923,9 @@
       "dev": true
     },
     "shimport": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/shimport/-/shimport-1.0.1.tgz",
-      "integrity": "sha512-Imf4gH+8WQmT1GvxS/x79qpmfnE6m50hyN1ucatX+7oMCgmaF8obZWCPIzSUe6+P+YmXM46lkP2pxiV2/lt9Og=="
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/shimport/-/shimport-2.0.3.tgz",
+      "integrity": "sha512-cI6ZhAblQUqZz2xnQxQpcA6lQUbB5YBSodcxGp+FeN8kgBLtXeiC7ZpM+5eGrH5OMNWDRiGTD0rRUnmuA+T6uw=="
     },
     "sirv": {
       "version": "1.0.6",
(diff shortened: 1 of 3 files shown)


Conduitry:
Is the `inject_styles` file not getting sent through Babel (or, for that matter, Terser)? Is it just being output as-is, untranspiled and unminified?

benmccann:
On `master` yes. But I've sent https://github.com/sveltejs/sapper/pull/1524 to fix that

Conduitry:
Should this be blocked by that then? It looks like several of the changes here are for rewriting in older JS syntax, unrelated to the `import.meta.url` thing.

Conduitry:
Once the inject_styles script is being transpiled and minified correctly via #1524, and once https://github.com/Rich-Harris/shimport/pull/32 is merged and released, is there any of this PR we still need besides updating shimport?

benmccann:
#1524 does minify the `inject_styles` output, but it doesn't seem to transpile it. I would assume if it's getting sent through `terser` it must be sent through `babel` as well, so I'm not quite sure what's happening. It's not that many bytes though and I've already burned so much time on this that I'm not sure it's really worth figuring out - though I do agree it'd be slightly nicer

For line 133, `renderDynamicImport` happens after `transform`, so I think that won't be able to be transformed by Babel and that change to `function` here is necessary

antony:
@benmccann the newest Babel is completely weird, it uses the filename of the babel config to determine what it applies to. I believe we should copy the babel config from the IE project into the official template, and it will fix this, because I literally spent weeks figuring it out.

benmccann:
Probably a good idea to copy your config into the template.

That's a bit different than the issue here though because I've been doing all my testing against [antony/sapper-ie](https://github.com/antony/sapper-ie) and this code still isn't being transpiled. I see the code on the filesystem getting transpiled. Maybe it's some kind of issue with the Rollup Babel plugin where it won't transpile code that doesn't exist on the filesystem or something. I don't really think it's worth spending several hours trying to figure it out though since this works

antony:
@benmccann I've finally had a chance to try this build and I'm afraid it doesn't work.

The reason is that the `shimport` file is an untranspiled file in the `/client/` directory, so IE11 just balks at the first `const`.

Screenshot from 2020-09-21 17-55-51

I'm going to see what happened with the older version - I feel like I remember that shimport was transpiled and loaded from `legacy`, but I don't exactly remember.

antony:
I've raised https://github.com/sveltejs/sapper/issues/1544 to track

PR #1528, opened 2020-09-18
Fixed warnings and added types in app.ts

ehrencrona:
There were some missing types in `app.ts` as well as a couple of compiler warnings. Added type annotations and fixed the warnings.

The `branch` property on `HydratedTarget` seems to have been quite wrong.

### Tests

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -32,7 +32,7 @@ export const initial_data: InitialData = typeof __SAPPER__ !== 'undefined' && __
 let ready = false;
 let root_component: InstanceType<typeof App>;
 let current_token: {};
-let root_preloaded: Promise<any>;
+let root_preloaded: Object | Promise<Object>;
 let current_branch: Branch = [];
 let current_query = '{}';
 
@@ -211,7 +211,7 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 	};
 
 	if (!root_preloaded) {
-		const root_preload = root_comp.preload || (() => {});
+		const root_preload = root_comp.preload || (() => ({}));
 		root_preloaded = initial_data.preloaded[0] || root_preload.call(preload_context, {
 			host: page.host,
 			path: page.path,
@@ -247,7 +247,7 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 
 			const { default: component, preload } = await load_component(components[part.i]);
 
-			let preloaded: any;
+			let preloaded: Object;
 			if (ready || !initial_data.preloaded[i + 1]) {
 				preloaded = preload
 					? await preload.call(preload_context, {
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -32,7 +32,7 @@ export const initial_data: InitialData = typeof __SAPPER__ !== 'undefined' && __
 let ready = false;
 let root_component: InstanceType<typeof App>;
 let current_token: {};
-let root_preloaded: Object | Promise<Object>;
+let root_preloaded: object | Promise<object>;
 let current_branch: Branch = [];
 let current_query = '{}';
 
@@ -247,7 +247,7 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 
 			const { default: component, preload } = await components[part.i].js();
 
-			let preloaded: Object;
+			let preloaded: object;
 			if (ready || !initial_data.preloaded[i + 1]) {
 				preloaded = preload
 					? await preload.call(preload_context, {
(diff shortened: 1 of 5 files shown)


benmccann:
This is great! I just did the same for some server-side stuff here: https://github.com/sveltejs/sapper/pull/1529

runtime/src/app/app.ts
```diff
@@ -62,15 +64,15 @@ stores.session.subscribe(async value => {
 
 export let prefetching: {
 	href: string;
-	promise: Promise<HydratedTarget>;
+	promise: Promise<any>;
```
benmccann, line 67: I'm not sure if this should be `any` because then we lose typing info on line 124 then. `Promise<HydratedTarget>` seemed right to me. was there somewhere that was causing issues?

benmccann:
Looks like this will need to be rebased since your other PR was merged

runtime/src/app/app.ts
ehrencrona: No, true. When I did this I had the impression the promise was only ever waited on for synchronization reasons and the result discarded, but that's not the case. But I'll rebase and this will go away.

ehrencrona:
Rebased!

runtime/src/app/app.ts
```diff
@@ -237,7 +239,7 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 
 			const { default: component, preload } = await load_component(components[part.i]);
 
-			let preloaded;
+			let preloaded: any;
```
benmccann, line 242: from the docs it sounds to me like this should be `object | Promise<object>` - https://sapper.svelte.dev/docs#Return_value

runtime/src/app/app.ts
```diff
@@ -263,10 +265,5 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 }
 
 export function load_component(component: DOMComponentLoader): Promise<DOMComponentModule> {
-	const promises: Array<Promise<any>> = [component.js()];
-	return Promise.all(promises).then(values => values[0]);
-}
-
-function detach(node: Node) {
-	node.parentNode.removeChild(node);
+	return component.js();
```
benmccann, line 268: thanks for cleaning this up. now that it's so much simpler maybe we can just get rid of `load_component`? that'd help reduce the coupling between `app` and `prefetch_routes`

runtime/src/app/app.ts
ehrencrona: Wow, that opened up a can of worms. The definition seems wrong; most likely it should `Object` and not `object` (a `Number` is an `object` but not an `Object`).

ehrencrona: Can preload return `null` / `undefined`? 🤔

runtime/src/app/app.ts
```diff
@@ -201,7 +203,7 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 	};
 
 	if (!root_preloaded) {
-		const root_preload = root_comp.preload || (() => {});
```
ehrencrona, line 204: I changed `() => {}` to `() => ({})`, in other words from a `void` return type to an empty object. If our types are correct that must have been what's intended. I hope it's correct...

runtime/src/app/app.ts
ehrencrona: Definitely!

runtime/src/app/app.ts
Conduitry: They _can_but I'm not sure what they behavior surrounding that should be - #843

benmccann: I think you might have the capitalization backwards: https://2ality.com/2020/01/typing-objects-typescript.html#object-(lowercase-%E2%80%9Co%E2%80%9D)-in-typescript%3A-non-primitive-values

ehrencrona: Wow, these are TypeScript subtleties I was not aware of. You're right, numbers etc are not included in `object`, so 

```
// compiles
const O: Object = 1;
// Type 'number' is not assignable to type 'object'.ts(2322)
const o: object = 1;
```

On the other hand
 
```
const O: Object = 1;
// compiles fine
const o: object = O;
```

So I think it's strictly speaking that both `object` and `Object` are assignable from `Number`, but that TS will do type coercion from `number` to `Number` only when encountering `Object`, not `object`.

Interestingly enough, this compiles too

```
const o: object = Object.create(null);
const O: Object = o;

console.log(O.toString);
```

but prints `undefined`. So `O` doesn't actually have have the `toScript` method that its type is supposed to guarantee.

I'm not sure where this leaves us. I'll switch back to `object`, though I do think it's a bit too obscure.

PR #1539, opened 2020-09-20
added modulepreload to check for adding rel=preload to output scripts...

riccardolardi:
…, also added crossorigin=use-credentials

Referencing https://github.com/sveltejs/sapper/issues/919


riccardolardi:
Why does the macOS-latest test fail?

Conduitry:
Probably just a flaky test - there are lots of timeouts on macOS tests.

What are the package-lock.json updates in this PR?

riccardolardi:
@Conduitry ouch that was a false commit. I've reverted it to the current master version and recommited. Still seems the last line `}` is recognized as a change... why's that

pngwn:
Could be a new line at the end. GitHub won’t show me the diff on my phone, so I cannot check.

riccardolardi:
Reverted the wrongly committed file to its master branch version.

riccardolardi:
Thanks @benmccann !

PR #1542, opened 2020-09-21
Refactored prefetching code to reduce dependencies

ehrencrona:
I moved the prefetching code that was distributed over `app.ts` and `router/index.ts` into one place (`prefetch/index.ts`).

This removes the dependency from router on prefetching. Unfortunately, there is still a circular dependency between `app.ts` and `prefetch` but breaking it makes the code more complicated so I dropped that.

### Tests


benmccann (approved):
fantastic! thank you!

PR #1547, opened 2020-09-21
Update TypeScript FAQ

benmccann:
Thanks to @mhatvan for providing TypeScript support in the Sapper template!

### DIFF
--- a/site/content/faq/300-typescript.md
+++ b/site/content/faq/300-typescript.md
@@ -2,7 +2,6 @@
 question: How do I use TypeScript with Sapper?
 ---
 
-The default template is currently setup only with JavaScript support, but here are some templates that community members have created with TypeScript support:
+If you're a Rollup user, you can run `node ./scripts/setupTypeScriptRollup.js`, which is provided in the template, after first starting a new project.
 
-* https://github.com/babichjacob/sapper-typescript-graphql-template
-* https://github.com/Zimtir/SENT-template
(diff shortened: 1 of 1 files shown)


site/content/faq/300-typescript.md
```diff
@@ -2,7 +2,6 @@
 question: How do I use TypeScript with Sapper?
 ---
 
-The default template is currently setup only with JavaScript support, but here are some templates that community members have created with TypeScript support:
+If you're a Rollup user, you can run `node ./scripts/setupTypeScriptRollup.js`, which is provided in the template, after first starting a new project.
```
Conduitry, line 5: I think ideally part of the 'generate the Rollup and webpack from the master branch' thing would leave a Rollup-specific script (without 'Rollup' as part of its name) in the Rollup template and (for now) wouldn't leave any script in the webpack template.

I do realize this isn't the right place for this discussion/request.

benmccann: Ah, that's a good call out. @mhatvan do you think that's something you could do in a follow up? It looks like the webpack vs rollup specific stuff is handled here: https://github.com/sveltejs/sapper-template/tree/master/_template

mhatvan: I am not very familiar with shell scripts unfortunately.

ehrencrona:
Should we add this to the "getting started" section of the documentation too? I'd expect a significant proportion of users to want TypeScript, so it would be good to mention it more prominently.

benmccann:
This was addressed in https://github.com/sveltejs/sapper/pull/1557

PR #1551, opened 2020-09-23
Split client into multiple libraries to avoid circular dependencies

### LINKED ISSUE #1527: Remove circular dependency
> Circular dependency: src/node_modules/@sapper/app.mjs -> src/node_modules/@sapper/internal/App.svelte -> src/routes/_layout.svelte -> src/node_modules/@sapper/app.mjs

We should fix this. `_layout` doesn't need to depend on the whole app. Only on the `stores` for example. We could refactor out the `stores` such that the layout and app can both import them independently without the need for this circular dependency

ehrencrona:
Rather than creating a single `app.mjs` that contains all the client code, create a file for each "module" (corresponding to what used to be subdirectories of `runtime/src/app`)

* Eliminated the subdirectories by moving e.g. `runtime/src/goto/index.ts` to `runtime/src/goto.ts`.
* Renamed `runtime/src/app.ts` to `runtime/src/start.ts` (since that is its default export).
* Created an `app.ts` with the same content as the previous `index.ts` to make existing code work as before
* Created a `stores.ts` for the stores export that was previously in `index.ts`

`runtime/src` directory:

Screen Shot 2020-09-23 at 3 44 30 PM

`src/node_modules/@sapper` directory:

Screen Shot 2020-09-23 at 3 44 50 PM

The point of the whole exercise is to fix #1527  – if, after this change, the client code imports 
```
	import stores from '@sapper/stores';
	import goto from '@sapper/goto';
```

rather than importing from `@sapper/app` (which will still work), the circular dependencies are broken.

As far as I can tell, this works and I can't see any negative effects from it, but it's entirely possible there are aspects I'm not aware of.

(note that it surfaces a new circular dependency: `start.mjs -> prefetch.mjs -> start.mjs`, but it's not too hard to break. i'll do it if there is agreement this is a good idea)

Obviously, this is a rather big change to fix a rather small issue. Waiting for feedback on how to proceed.


benmccann:
Do you think it'd be possible to allow the user to import things via a single line like:
```
import {stores, goto} from '@sapper';
```

ehrencrona:
We couldn't re-export everything from a single package `@sapper`. If it depends on everything and if the project code depends on it, then the project code depends on "everything" again.

But we could export only the stuff that's typically imported by project code, i.e. `goto`, `stores` and (once the dependency is broken) `prefetch`.

Is this really a breaking change? Importing from `@sapper/app` will still work; I don't see what else would change from the project code's perspective.

benmccann:
Thanks! I'm not sure how much it's worth trying to cleanup the codebase at this point since it's basically in maintenance mode, but we do have the circular dependency in SvelteKit still and if you'd like to take a stab at it there I think it'd be great to fix it. Thanks again for all your help on this!

PR #1555, opened 2020-09-24
Warn against accesses to browser globals

### LINKED ISSUE #1554: Detect use of window/document on the server and provide a friendlier message

See https://github.com/sveltejs/sapper/issues/1226

Just be more helpful when people attempt to use window/document in server-side code

1. A FAQ item on the Sapper docs
2. A way of detecting `window`/`document` usage (when undefined, in case somebody is doing something quite odd) on the server which points to the FAQ
3. The FAQ referencing [this docs page](https://sapper.svelte.dev/docs#Making_a_component_SSR_compatible)

Letting people suffer

It comes up a *lot*, it could reduce the support/help overhead and repetition.

ehrencrona:
Detect attempts at accessing `document` or `location` in server-side code to fix #1554.

I see two ways of doing this: one is to set the global `document` and `location` to a proxy that throws if it is accessed. I started implementing this, but then it struck me there will be code doing `typeof window === 'undefined'` to check if they are running server-side so it's not a good solution.

I'm therefore instead catching `ReferenceError: document is undefined` from the client code. This has the advantage of only catching errors that definitely originate in the client code, but it's a bit more brittle; there are other potential errors the code might throw as a result of accesses.

I also updated the documentation to make clearer that component code runs both on the client and server side and couldn't keep myself from doing some general cleanup in the sections i touched.

### Tests


site/content/docs/04-preloading.md
```diff
@@ -2,7 +2,15 @@
...
-As seen in the [routing](docs#Routing) section, page components can have an optional `preload` function that will load some data that the page depends on. This is similar to `getInitialProps` in Next.js or `asyncData` in Nuxt.js.
+Page components can define a `preload` function that runs before the component is created. The values it returns are passed as props to the page.
+
+Preload functions are typically used to load data that the page depends on, hence its name. Having the data available before the page load allows you to avoid flashing while it is fetched.
+
+Preload is the Sapper equivalent to `getInitialProps` in Next.js or `asyncData` in Nuxt.js.
+
+Note that `preload` will run both on the server side and on the client side. It may therefore not reference any APIs only present in the browser.
+
+The following code shows how to load a blog post and passs it to the page in the `article` prop:
```
benmccann, line 13: typoL passs

site/content/docs/04-preloading.md
```diff
@@ -2,7 +2,15 @@
 title: Preloading
 ---
 
-As seen in the [routing](docs#Routing) section, page components can have an optional `preload` function that will load some data that the page depends on. This is similar to `getInitialProps` in Next.js or `asyncData` in Nuxt.js.
+Page components can define a `preload` function that runs before the component is created. The values it returns are passed as props to the page.
+
+Preload functions are typically used to load data that the page depends on, hence its name. Having the data available before the page load allows you to avoid flashing while it is fetched.
```
benmccann, line 7: "Having the data available before the page load allows you to avoid flashing while it is fetched." - I'm not sure that it would necessarily cause flashing to fetch data on the client. Many sites load some amount of data after the client has loaded and just have a portion of the page that's not filled in or has a loading widget or something. I might just remove this sentance

site/content/docs/06-server-side-rendering.md
```diff
@@ -2,19 +2,31 @@
 title: Server-side rendering
 ---
 
-Sapper, by default, renders server-side first (SSR), and then re-mounts any dynamic elements on the client. Svelte provides [excellent support for this](https://svelte.dev/docs#Server-side_component_API). This has benefits in performance and search engine indexing, among others, but comes with its own set of complexities.
+Sapper will render any component on the server side and send it to the client as HTML. It will then run the component again on the client side to allow it to update based on dynamic data. This is known as "hydration." Svelte has excellent support for this.
```
benmccann, line 5: "Svelte has excellent support for this." - Svelte supports it, but I'm not sure I'd say it's excellent. There's lots of issues around it - it's slow and doesn't work well with videos, iframes, etc.

Anyway, I think it's a bit late in the documentation to explain what hydration is because hydration is mentioned a few times before you get to this section

ehrencrona: I just kept "excellent support" from what was there before; I certainly don't mind deleting it. It doesn't help anyone.

I tried searching for hydration in the docs and I can't find any explanation of it. That's why I added it. Is it explain somewhere under a different name? The term is used twice before but without defining it.

site/content/docs/04-preloading.md
ehrencrona: Yeah, this is from trying to preserve the sentence "it runs before the component is created, allowing you to avoid flashes while data is fetched." But you're right; "flashing" doesn't cover it. If anything, it's things jumping around on the page. I'll rewrite it.

ehrencrona:
@benmccann can we merge? I applied your feedback re: the docs.

Conduitry:
I saw a comment on the source issue with getters on the global object that, if it works, sounds like a more robust solution than sniffing exception messages. Did you see or try that?

ehrencrona:
@Conduitry I did consider that. See the description of the PR.

Conduitry:
Gotcha, sorry, I had missed that. Makes sense. :+1:

site/content/docs/06-server-side-rendering.md
```diff
@@ -2,19 +2,29 @@
...
 
-Sapper works well with most third-party libraries you are likely to come across. However, sometimes, a third-party library comes bundled in a way which allows it to work with multiple different module loaders. Sometimes, this code creates a dependency on `window`, such as checking for the existence of `window.global` might do.
+If you need access to these variables, you can run code exclusively on the client side by wrapping it in
 
-Since there is no `window` in a server-side environment like Sapper's, the action of simply importing such a module can cause the import to fail, and terminate the Sapper's server with an error such as:
-
-```bash
-ReferenceError: window is not defined
+```js
+if (typeof window !== 'undefined') {
```
ignatiusmb, line 12: What about `process.browser`? It's provided in both rollup and webpack default template, and I seem to see it being recommended more than checking `typeof window`.

Conduitry: I think there's a bit of an ideological question here of how much we want to separate `sapper` from `sapper-template`. `process.browser` isn't something that's built in to Sapper - it's just something that happens to be replaced at bundle time by the default template. When it's available, `process.browser` _is_ the better thing to use because the replacement happens early enough that unused code can be removed.

ehrencrona:
Can we merge? :P

PR #1557, opened 2020-09-24
Mention TypeScript in the docs under "Getting started"

ehrencrona:
Mentioning the TypeScript conversion script in "Getting started" in the docs (because there is otherwise no mention of it anywhere).


site/content/docs/00-introduction.md
```diff
@@ -49,3 +49,11 @@ npm run dev
 ```
 
 This will scaffold a new project in the `my-app` directory, install its dependencies, and start a server on [localhost:3000](http://localhost:3000). Try editing the files to get a feel for how everything works – you may not need to bother reading the rest of this guide!
+
+#### TypeScript support
+
+Sapper supports TypeScript. If you are using the template with Rollup as described in "Getting started" you can convert your project to TypeScript simply by running:
+
+```bash
+node scripts/setupTypeScriptRollup.js
```
benmccann, line 58: this is probably a good idea, but we should probably implement https://github.com/sveltejs/sapper-template/issues/268 first so that we don't have to remember to go back and update the docs, etc.

benmccann:
Thanks for your work on cleaning this up!

Would you mind also updating `site/content/faq/300-typescript.md` to point to the new template script? We should probably also update that page to point to https://github.com/sveltejs/integrations#sapper-templates-boilerplates for community-maintained templates instead of listing the two that are there now

ehrencrona:
Oh yeah, I hadn't seen the FAQ. Updated it.

PR #1558, opened 2020-09-24
Eliminating the rollup warning "'preload' is not exported"

ehrencrona:
This eliminates the `'preload' is not exported by 'src/routes/_layout.svelte'` warning that Rollup produces on the template project.

Once this has been released the line https://github.com/sveltejs/sapper-template/blob/master/rollup.config.js#L15 that currently suppresses the message can be removed.

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -209,7 +209,7 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 	};
 
 	if (!root_preloaded) {
-		const root_preload = root_comp.preload || (() => ({}));
+		const { preload: root_preload = (() => ({})) } = root_comp;
 		root_preloaded = initial_data.preloaded[0] || root_preload.call(preload_context, {
 			host: page.host,
 			path: page.path,


Conduitry:
Interesting. Do you know whether _not_ emitting the warning in this case is intended Rollup behavior, or might it just be a case they haven't gotten to, and the warning may reappear one day?

ehrencrona:
I suspect the latter. I also haven't found any intended way of actually checking whether an export exists or not, so I suspect it's a case they haven't really thought about. 

(I would have expected `typeof root_comp.preload != 'undefined'` to work, alternatively `typeof root_comp['preload'] != 'undefined'`, but neither eliminates the warning)

Conduitry:
Okay. If that's the case, I think I'd rather stick with the simple thing we have now, and manually suppress this warning in the template.

Conduitry:
Closing this, for the reason I outlined above. This feels like tricking Rollup rather than a proper solution.

ObserverOfTime:
This is more than just a warning in webpack:
```
ERROR in ./src/node_modules/@sapper/app.mjs 485:33-50
export 'root_comp'.'preload' (imported as 'root_comp') was not found in './internal/manifest-client' (possible exports: default)
 @ ./src/client.js 1:0-34 3:0-5

webpack 5.17.0 compiled with 1 error in 2208 ms
> Encountered errors while building app
```

Issue #1561, opened 2020-09-24
document.baseURI is undefined in IE11


antony:
**Describe the bug**

document.baseURI [isn't usable](https://stackoverflow.com/questions/52269994/when-window-document-baseuri-be-undefined) in Internet Explorer, meaning Sapper 0.28.8 can't render pages due to the use of it in `inject_styles.js`.

Instead, an initial SSR render is shown, then a 500 page with the error `cannot get property 'length' of undefined or null reference` is displayed.

**Logs**
```
Unable to get property 'length' of undefined or null reference
 {"description":"Unable to get property 'length' of undefined or null reference","number":-2146823281,"stack":"TypeError: Unable to get property 'length' of undefined or null reference\n at Anonymous function (blob:1C66BD39-459E-4806-9A86-79C51EBAD409:1:316)\n a
```

**To Reproduce**
Checkout new sapper project
Add polyfills
Run in IE11

**Expected behavior**
Pages should render

- The output of `npx envinfo --system --npmPackages svelte,sapper,rollup,webpack --binaries --browsers`

  System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz
    Memory: 169.85 MB / 7.68 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
    Yarn: 1.22.4 - ~/.nvm/versions/node/v12.18.3/bin/yarn
    npm: 6.14.6 - ~/.nvm/versions/node/v12.18.3/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Firefox: 80.0.1
  npmPackages:
    rollup: ^2.3.4 => 2.28.1 
    sapper: ^0.28.0 => 0.28.8 
    svelte: ^3.17.3 => 3.26.0 

- Your browser

IE11

**Severity**
Blocker

**Additional context**

Fixing this could also resolve the issue here:

https://github.com/sveltejs/sapper/issues/1034

Conduitry:
This should be fixed in 0.28.9.

PR #1562, opened 2020-09-24
fix for usage of baseURI which doesn't work in IE11

antony:
`document.baseURI` doesn't exist in IE and causes errors. This attempts to get it and then falls back to looking at the base tag if present.

Fixes https://github.com/sveltejs/sapper/issues/1561 and https://github.com/sveltejs/sapper/issues/1034 as well as an issue where prefetch would also fail in IE as it also depends on baseURI.

Opening in draft because:

- [x] I've not tested it, I will do tomorrow.
- [x] I haven't written any tests for the helper(s)

### Tests

### DIFF
--- a/runtime/src/app/baseuri_helper.ts
+++ b/runtime/src/app/baseuri_helper.ts
@@ -1,10 +1,10 @@
-export function get_base_uri (window_document) {
-  let baseURI = window_document.baseURI;
+export function get_base_uri(window_document) {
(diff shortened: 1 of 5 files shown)


benmccann (approved):
lgtm

antony:
Added unit tests (this is the first time I've used typescript, so feel free to suggest a refactor)

PR #1566, opened 2020-09-27
Change rel=prefetch to sapper:prefetch

### LINKED ISSUE #1401: Rename rel=prefetch to sapper:prefetch
Sapper overrides the browser's default `rel=prefetch` behavior. There seemed to be agreement that we should stop doing this.

ArekBartnik:
Closes #1401

It's a breaking change.


benmccann:
This will be called `sveltekit:prefetch` in SvelteKit, but we might as go ahead and call this `sapper:prefetch` here

zolotokrylin:
Does this change break the existing code, meaning, `rel=prefetch` will not work anymore, right?

PR #1569, opened 2020-09-28
do not enter an infinite loop if layout's preload crashed on error page

### LINKED ISSUE #1506: Calling this.error inside preload on a _layout route hangs the site
**Describe the bug**
What the title says. I'm calling a service from a preload method in a _layout page, if the service fails then I'm calling the this.error function. When this happens on a layout page it seems to get stuck in permanent loading in browser, it never finishes loading.

**Logs**
There's nothing out of the ordinary in the server logs. In the browser the page never gets to load in the first place.

**To Reproduce**

Create a new sapper project using the template, and add this to a _layout.svelte: 

```
<script context="module">
  export function preload({ params, query }) {
    this.error(500, "testing");
  }
</script>
```

If this is done in any other route, non layout, then it correctly loads and shows the error page.

**Expected behavior**
Expected to see the error page.

- The output of `npx envinfo --system --npmPackages svelte,sapper,rollup,webpack --binaries --browsers`

```
npx: installed 1 in 1.225s

  System:
    OS: macOS 10.15.6
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 676.62 MB / 16.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 12.16.1 - /usr/local/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.14.8 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 85.0.4183.102
    Firefox Developer Edition: 81.0
    Safari: 13.1.2
  npmPackages:
    rollup: ^2.3.4 => 2.26.11 
    sapper: ^0.28.0 => 0.28.6 
    svelte: ^3.17.3 => 3.25.0 
```

- Your browser

Tested on latest Chrome and Firefox Developer Edition. I don't think this is a browser issue.

- Your hosting environment (i.e. Local, GCP/AWS/Azure, Vercel/Begin, etc...)

Local testing with `npm run dev`

- If it is an exported (`npm run export`) or dynamic application

**Severity**
How severe an issue is this bug to you? Is this annoying, blocking some users, blocking an upgrade or blocking your usage of Sapper entirely?

Sapper is currently only being evaluated. Not being able to do proper error handling would be a blocker, maybe there's a workaround or maybe I'm doing something wrong.

ehrencrona:
If the layout's preload function throws or calls `this.error` on the error page, we'd enter an infinate loop.

Break that by bailing and returning an unstyled error page.

Fixes #1506

### Tests

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -27,10 +27,10 @@ export function get_page_handler(
 
 	const { pages, error: error_route } = manifest;
 
-	function bail(req: Req, res: Res, err: Error) {
+	function bail(res: Res, err: Error|string) {
 		console.error(err);
 
-		const message = dev ? escape_html(err.message) : 'Internal server error';
+		const message = dev ? escape_html(typeof err === 'string' ? err : err.message) : 'Internal server error';
 
 		res.statusCode = 500;
 		res.end(`<pre>${message}</pre>`);
@@ -97,7 +97,7 @@ export function get_page_handler(
 		try {
 			session = await session_getter(req, res);
 		} catch (err) {
-			return bail(req, res, err);
+			return bail(res, err);
 		}
 
 		let redirect: { statusCode: number, location: string };
@@ -192,7 +192,7 @@ export function get_page_handler(
 			preloaded = await Promise.all(toPreload);
 		} catch (err) {
 			if (error) {
-				return bail(req, res, err);
+				return bail(res, err);
 			}
 
 			preload_error = { statusCode: 500, message: err };
@@ -211,7 +211,13 @@ export function get_page_handler(
 			}
 
 			if (preload_error) {
-				handle_error(req, res, preload_error.statusCode, preload_error.message);
+				if (!error) {
+					handle_error(req, res, preload_error.statusCode, preload_error.message);
+				}
+				else {
+					bail(res, preload_error.message);
+				}
+
 				return;
 			}
 
@@ -357,7 +363,7 @@ export function get_page_handler(
 			res.end(body);
 		} catch (err) {
 			if (error) {
-				bail(req, res, err);
+				bail(res, err);
 			} else {
 				handle_error(req, res, 500, err);
 			}
--- a/test/apps/errors/src/routes/_layout.svelte
+++ b/test/apps/errors/src/routes/_layout.svelte
@@ -3,6 +3,14 @@
 	const { page } = stores();
 </script>
 
+<script context="module">
+  export function preload({ params, query }) {
+		if (query['layoutthrows']) {
+			this.error(500, "layout preload threw");
+		}
+  }
+</script>
+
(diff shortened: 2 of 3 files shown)


test/apps/errors/src/routes/_layout.svelte
```diff
@@ -3,6 +3,14 @@
 	const { page } = stores();
 </script>
 
+<script context="module">
+  export function preload({ params, query }) {
```
benmccann, line 7: spaces here should be tabs

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -211,7 +211,13 @@ export function get_page_handler(
 			}
 
 			if (preload_error) {
-				handle_error(req, res, preload_error.statusCode, preload_error.message);
+				if (!error) {
+					handle_error(req, res, preload_error.statusCode, preload_error.message);
+				}
+				else {
```
benmccann, line 217: We'd normally put this all on one line like: `} else {`

ehrencrona: Fixed it. How about checking in a `.prettierrc` and running it all through prettier? At the moment I have to format by hand because running files through prettier makes too many changes.

benmccann: I've never used prettier, but am open to the idea of at least seeing what code formatted with prettier looks like. Up until this point I'd been avoiding it because I kept seeing so many issues reported in prettier-plugin-svelte, but perhaps you've fixed enough of those now :smile:

test/apps/errors/src/routes/_layout.svelte
benmccann: @ehrencrona it looks like this one still needs to be updated

ehrencrona: sorry, missed that comment

runtime/src/server/middleware/get_page_handler.ts
ehrencrona: I improved the linting rules instead to check for this in #1573

test/apps/errors/test.ts
```diff
@@ -98,6 +98,14 @@ describe('errors', function() {
 		await assertErrorPageRenders(500);
 	});
 
+	it('handles errors occurring in the layout preload', async () => {
+		const res = await r.load('/?layoutthrows=true');
+
+		const responseText = await res.text();
+
+		assert.ok((responseText).includes('Internal server error'), `Expected response "${responseText}" to say "Internal server error"`);
```
benmccann, line 106:
```suggestion
		assert.ok(responseText.includes('Internal server error'), `Expected response "${responseText}" to say "Internal server error"`);
```

Issue #1565, opened 2020-09-26
Allow multiple CSP nonces


benmccann:
According to https://github.com/sveltejs/sapper/issues/1248#issuecomment-699442197 `replace` is only getting called for the first occurrence and we should be allowing multiple occurrences.

The line of code to change is here: https://github.com/sveltejs/sapper/pull/1249/files#diff-9c713b090053bf2158dc7db4484fb401R332

Conduitry:
This should be fixed in 0.28.10.

PR #1570, opened 2020-09-28
Replace all occurences of sapper.cspnonce

benbarbersmith:
This addresses https://github.com/sveltejs/sapper/issues/1565 and allows use of nonces on multiple scripts and stylesheets.

The changes to test/apps/cspnonce/src/template.html fails on master (reflecting the current issue) but passes after the changes in this PR.

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -351,7 +351,7 @@ export function get_page_handler(
 				.replace('%sapper.html%', () => html)
 				.replace('%sapper.head%', () => head)
 				.replace('%sapper.styles%', () => styles)
-				.replace(/%sapper.cspnonce%/g, nonce_value);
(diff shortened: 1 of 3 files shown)


runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -351,7 +351,7 @@ export function get_page_handler(
 				.replace('%sapper.html%', () => html)
 				.replace('%sapper.head%', () => head)
 				.replace('%sapper.styles%', () => styles)
-				.replace('%sapper.cspnonce%', () => nonce_value);
+				.replace(/%sapper.cspnonce%/g, nonce_value);
```
Conduitry, line 354:
```suggestion
				.replace(/%sapper\.cspnonce%/g, () => nonce_value);
```

The `.` should be escaped, and the `() =>` should be preserved as it's there to avoid `.replace`'s special behavior with `$1`, `$2`, etc.

benmccann (approved):

PR #1573, opened 2020-09-30
Linting covering tests and site

ehrencrona:
The lint was previously not particularly comprehensive. It ignored the tests, `.svelte` files, the `site` directory and TypeScript definition files. Including all source files. 

Fixed resulting warnings. Also fixed instances of using space for indenting.

### Tests


Conduitry:
There's a shared ESLint config used in the Svelte repo that we should be using here as well. I haven't compared your suggested additions versus what's already in that shared config, but if you have suggestions for new rules they should go on the shared config repo.

ehrencrona:
@Conduitry Thanks for the tip; I hadn't understood how that worked. There are three rules I'd like to add:

```
// force indents to be tabs
indent: ['error', 'tab'],
// force "} else {" on single line
'brace-style': ['error', '1tbs'],
// check for unused variables (except in function parameters)
'@typescript-eslint/no-unused-vars': ['error', { args: 'none' }]
```

If there are no disagreements about adding these, I'll lint the other sub-projects according to these rules and fix any warnings. Only once that is done and the fixes merged, I'll make a PR to the `eslint-config`. That should prevent builds from breaking.

Does that sound reasonable?

antony:
@ehrencrona There is a full process for adding new rules (one by one) which can be kicked off here - https://github.com/sveltejs/eslint-config/issues/new/choose

benmccann:
The indentation rule we purposefully left out because [it's broken](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/indent.md) and was doing weird things. It's a shame because it'd probably be the highest value rule if it were working correctly. The other two seem worth raising PRs for voting on

runtime/index.d.ts
```diff
@@ -15,9 +15,7 @@ declare module "@sapper/app" {
 	export const stores: () => unknown;
 }
 
-declare module "@sapper/server" {
-	import { Handler, Req, Res } from '@sapper/internal/manifest-server';
```
benmccann, line 19: I'm surprised these are removed. It looks like `Req` and `Res` are used a few lines later

package.json
```diff
@@ -37,15 +37,15 @@
     "@types/node": "^10.0.0",
     "@types/node-fetch": "^2.5.7",
     "@types/puppeteer": "^3.0.1",
-    "@typescript-eslint/eslint-plugin": "^3.0.2",
-    "@typescript-eslint/parser": "^3.0.2",
+    "@typescript-eslint/eslint-plugin": "^4.3.0",
```
benmccann, line 40: it looks like these upgrades don't actually change anything in the code currently being checked. I went ahead and pushed the upgrades to separate them out and make it a bit easier to review

runtime/index.d.ts
ehrencrona: You're right. It's also weird it doesn't give any errors anywhere with them removed. I'll put them back.

ehrencrona:
Ok, I've removed the new rules and what remains is extending linting to the tests and the `site` directory. That and the fixes of warnings (and ex-warnings) still makes sense.

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -161,12 +161,16 @@ export function get_page_handler(
 
 		try {
 			const root_preload = manifest.root_comp.preload || (() => {});
-			const root_preloaded: PreloadResult = root_preload.call(preload_context, {
+			const root_preloaded: PreloadResult = root_preload.call(
+				preload_context,
```
benmccann, line 165: why is anything under `runtime/src` changing? I think this file was already linted and this change isn't really necessary. Personally, I prefer the previous indentation on this file, which was more concise

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -378,7 +378,7 @@ export default class RollupCompiler {
 
 				default:
 					console.log(`Unexpected event ${event.code}`);
-			}
+				}
```
benmccann, line 381: this doesn't seem to be properly aligned anymore

test/apps/encoding/src/routes/fünke.json.js
```diff
@@ -4,6 +4,6 @@ export function get(req, res) {
 	});
 
 	res.end(JSON.stringify(
-		"I'm afraid I just blue myself"
+		'I\'m afraid I just blue myself'
```
benmccann, line 7: This change shouldn't be necessary if you're using 5.2.0 of the lint config. See https://github.com/sveltejs/eslint-config/pull/8

runtime/src/app/router/index.ts
```diff
@@ -19,11 +19,11 @@ export function set_cid(n: number) {
...
 
-const _history = typeof history !== 'undefined' ? history : {
-	pushState: (state: any, title: string, href: string) => {},
-	replaceState: (state: any, title: string, href: string) => {},
-	scrollRestoration: ''
-};
+const _history: History = typeof history !== 'undefined' ? history : {
+	pushState: () => {},
+	replaceState: () => {},
+	scrollRestoration: 'auto'
```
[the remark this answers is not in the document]
ehrencrona: Not directly. It's from adding types. This object literal wasn't even a `Partial<History>` because `scrollRestauration` is not allowed to have empty string as value, so I set it to `auto` which is the browser default.

runtime/src/server/middleware/get_page_handler.ts
ehrencrona: Not sure. I probably ran prettier over this block for whatever reason. I'll roll it back.

src/core/create_compilers/RollupCompiler.ts
ehrencrona: true. eagle eyes!

test/apps/encoding/src/routes/fünke.json.js
ehrencrona: i was using the old version. rolling this back.

benmccann:
It looks like this PR needs to be rebased

benmccann (approved):

PR #1577, opened 2020-10-01
Bump timeout to 30 seconds, using port-authority

alexbjorlig:
According to the source code here https://github.com/Rich-Harris/port-authority/blob/bf12e69618cc59dbc0d46a5132902df8a584e39f/src/wait.ts#L3,
we can specify a timeout - and a default timeout of 5 seconds is not
high enough for many bigger express applications (see https://github.com/sveltejs/sapper/issues/730).

This PR is simply inspired by the dialog in pull request: https://github.com/sveltejs/sapper/pull/985,
but I thought it would just be simpler to create a new one.

Closes https://github.com/sveltejs/sapper/pull/985

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -237,7 +237,7 @@ class Watcher extends EventEmitter {
 					const restart = () => {
 						this.crashed = false;
 
-						return ports.wait(this.port)
+						return ports.wait(this.port, {timeout: 30000})
 							.then((() => {
 								this.emit('ready', {
 									port: this.port,


alexbjorlig:
Ok - but do we have another workaround? It's a relatively big issue right? We can't reliably run sapper with our express server because of the timeout being too low

Conduitry:
As I said in the comment linked Ben linked above, the main thing that could be done would be to make the dev livereload server start up right away, rather than waiting for the regular server to be ready. If someone has a giant app or a computer with limited resources, they could always end up being later than any timeout. Polling periodically might be nice if we want to fix the error message displayed to the user, but the important part is to not prevent the dev livereload server from starting.

alexbjorlig:
> the main thing that could be done would be to make the dev livereload server start up right away, rather than waiting for the regular server to be ready

How would you like that to be done? Should this [function](https://github.com/sveltejs/sapper/blob/43fd91bc4495adf33d397a012c031c5dc7c83e23/src/api/dev.ts#L235-L263) just emit 'ready' right away, instead of waiting for the port - or is it more complicated than that?

I would be happy to help on this, but might need some guidance 😅

benmccann:
Maybe we should focus on making sure this works as expected in SvelteKit instead since this one seemed a bit difficult to handle correctly. Would love if you can test the startup in SvelteKit and see if it works as expected for you

Issue #1576, opened 2020-10-01
Preload tags causing troubles - warning in console + google cannot crawl page anymore


seppegadeyne:
**Describe the bug**
The new feature from 0.28.9, the added preload tag is returning a warning in the console and breaks Google from crawling the website succesfully. 

**Logs**
Warnings in console: https://bit.ly/30o5JG5
Google error loading files: https://bit.ly/2GnziAg
Console warnings google: https://bit.ly/36nHNWQ
Staging: Sapper 0.28.9: https://straffesites-nieuwbouwinspanje-r8ntu3td8.vercel.app/
Live: Sapper 0.28.8: https://nieuwbouwinspanje.com/

**To Reproduce**
Upgrade to Sapper 0.28.9. 

**Expected behavior**
The client library is loading async those index bundles. Only preloading the client file should be enough; there should not be preload tags for those other files. 

**Information about your Sapper Installation:**

System:
OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
Memory: 696.18 MB / 15.50 GB
Container: Yes
Shell: 5.0.17 - /bin/bash

Binaries:
Node: 10.19.0 - /usr/bin/node
npm: 6.14.8 - /usr/local/bin/npm

Browsers:
Chrome: dev

npmPackages:
rollup: ^2.3.5 => 2.28.2 
sapper: 0.28.8 => 0.28.8 (working version --> 0.28.9 breaking)
svelte: ^3.25.0 => 3.29.0 

- Browser: 
Google Chrome Version 87.0.4270.0 (Official Build) dev (64-bit)
Firefox 82.0b5 (64-bits)

- Hosting environment:
Vercel (exported)

**Severity**
It is severe, Google is not able to crawl the page succesfully.

benmccann:
The issue here looks to be the `crossorigin="use-credentials"` that was added in https://github.com/sveltejs/sapper/pull/1539. You could test removing that to see if it fixes the issue

There should indeed be `preload` tags. By preloading assets relevant for the page we improve page speed. The reason the assets aren't used by the page appears to be because the `crossorigin` attribute is incorrect

[...]

PR #1578, opened 2020-10-01
Preload modules as modulepreload

seppegadeyne:
Found a solution rel="preload" is not suited for modules, so using rel="modulepreload" is working fine in this setup with the credentials attribute as it is. Have been testing it and I don't get any console warnings anymore.

See issue: Preload tags causing troubles - warning in console + google cannot crawl page anymore [#1576](https://github.com/sveltejs/sapper/issues/1576)

### DIFF
--- a/src/api/export.ts
+++ b/src/api/export.ts
@@ -229,7 +229,7 @@ async function _export({
 			link.refs.forEach((ref: Ref) => {
 				if (ref.rel === 'preload' || ref.rel === 'modulepreload') {
 					body = (body as string).replace('</head>',
-						`<link rel="preload" as=${JSON.stringify(ref.as)} href=${JSON.stringify(ref.uri)} ${ref.as === 'script' ? 'crossorigin="use-credentials"' : ''}></head>`);
+						`<link rel=${ref.rel} as=${JSON.stringify(ref.as)} href=${JSON.stringify(ref.uri)} ${ref.as === 'script' ? 'crossorigin="use-credentials"' : ''}></head>`);
 				}
 			});
 


benmccann:
Is the `crossorigin` attribute necessary though? It sounds like perhaps it's not and maybe we should remove it

seppegadeyne:
The `crossorigin` is needed because `<script src="/client/client.5c473246.js" type="module" crossorigin="use-credentials"></script>` is using that. They should all use the same crossorigin. 

_For <script> and <link> tags, you can set the credentials mode with the crossorigin attribute. However, it turns out that a <script type="module"> with no crossorigin attribute indicates a credentials mode of omit, which doesn't exist for <link rel="preload">. This means that you would have to change the crossorigin attribute in both your <script> and <link> to one of the other values, and you might not have an easy way of doing so if what you're trying to preload is a dependency of other modules._

src/api/export.ts
```diff
@@ -229,7 +229,7 @@ async function _export({
 			link.refs.forEach((ref: Ref) => {
 				if (ref.rel === 'preload' || ref.rel === 'modulepreload') {
 					body = (body as string).replace('</head>',
-						`<link rel="preload" as=${JSON.stringify(ref.as)} href=${JSON.stringify(ref.uri)} ${ref.as === 'script' ? 'crossorigin="use-credentials"' : ''}></head>`);
+						`<link rel=${ref.rel} as=${JSON.stringify(ref.as)} href=${JSON.stringify(ref.uri)} ${ref.as === 'script' ? 'crossorigin="use-credentials"' : ''}></head>`);
```
Conduitry, line 232: For consistency, this should have been something like `rel=${JSON.stringify(ref.rel)}`, although that doesn't matter immensely since the exported HTML is going to be minified anyway.

benmccann: ah, yeah. I'll update it

Issue #1254, opened 2020-06-04
Allow process.env.PORT to be defined via .env file


martinburger:
I would like to use [dotenv](https://github.com/motdotla/dotenv) to configure the port number the app listens on via a `.env` file.

I minimal implementation of that approach follows these steps:

1. Start new project using `npx degit "sveltejs/sapper-template#rollup" sapper-dotenv-port`.
1. Install dotenv via `npm install dotenv`.
1. Create `.env` file with content `PORT=5000`.
1. Change file `src/server.js` as follows:

```patch
--- a/src/server.js
+++ b/src/server.js
@@ -6,6 +6,9 @@ import * as sapper from '@sapper/server';
 const { PORT, NODE_ENV } = process.env;
 const dev = NODE_ENV === 'development';

+const result = require('dotenv').config({ debug: true });
+console.log('dotenv result:', result);
+
 polka() // You can also use Express
        .use(
                compression({ threshold: 0 }),
```

I would expect that the Sapper app listens on port 5000. However, it listens on default port 3000 instead. The console output shows the reason:

```console
❯ npm run dev

> TODO@0.0.1 dev /[...]/sapper-dotenv-port
> sapper dev

✔ server (952ms)
✔ client (959ms)
> Listening on http://localhost:3000
[dotenv][DEBUG] did not match key and value when parsing line 2:
[dotenv][DEBUG] "PORT" is already defined in `process.env` and will not be overwritten
dotenv result: { parsed: { PORT: '5000' } }
✔ service worker (28ms)
```

As far as I understand the cause of `process.env.PORT` being already defined, this is due to the following two lines in the Sapper code base:

1. `PORT: this.port` at [src/api/dev.ts#L287](https://github.com/sveltejs/sapper/blob/c9a62fa9f8de240de0421682852476621a310d0b/src/api/dev.ts#L287)

```javascript
this.proc = child_process.fork(`${dest}/server/server.js`, [], {
  cwd: process.cwd(),
  env: Object.assign({
    PORT: this.port
  }, process.env),
  stdio: ['ipc'],
  execArgv
});
```

2. `process.env.PORT = process.env.PORT || ${opts.port || 3000};` at [src/cli.ts#L179](https://github.com/sveltejs/sapper/blob/dca0de9a5f54a751c1ec3f616d8245a06f922603/src/cli.ts#L179)

```javascript
fs.writeFileSync(launcher, `
  // generated by sapper build at ${new Date().toISOString()}
[...]

PR #1580, opened 2020-10-02
Allow process.env.PORT to be defined via .env file

karlos1337:
As commented in #1254 current config system doesn't support defining PORT env var via .env file, it's an easy fix, installing dotenv package and requiring it in the same file that port is defined it's enough since its already have in count process.env.PORT.


peopledrivemecrazy:
I can test this tomorrow.

antony:
I don't think we specifically want to have a dependency on `dotenv`. I also feel that having a dependency on a specific configuration mechanism is not ideal.

I've provided a workaround in your original issue.

PR #1581, opened 2020-10-02
Ensure CSS is applied to nested route layouts

### LINKED ISSUE #1579: CSS not applied to nested route layouts (server-rendered only)
**Describe the bug**
When using a nested _layout.svelte component CSS does not appear to be applied to any markup defined within that layout component (the markup within the layout remains unstyled). When the component is subsequently hydrated the client JS does inject the relevant styles. This makes the bug hard to spot but its effects can be clearly seen if:

- You are using a slower or throttled connection (you will get a flash of unstyled content), or;
- JS is disabled in the browser, in which case the styles are never applied.

Some important clarifications:

- This bug affects both markup and styles defined directly within the layout component as well as any child components imported directly into the layout component (i.e. styles defined in these components are not applied either.)
- This bug *does not* affect slotted content within the layout component.
- This bug *does not* affect the top-level layout component; only layouts nested one level down or deeper seem to experience this issue.

**Logs**
There are no relevant error logs or warnings.

**To Reproduce**
Please find a minimal repro here: https://github.com/rodoch/sapper-repro

This is the base Sapper template with two changes:

- Svelte bumped to v3.29.0 & Sapper bumped to v0.28.9
- The `src/blog` directory now contains a nested _layout.svelte file with one styled div.

Again, the surest method to observe the behaviour is to disable JS in your browser or to simulate a throttled connection.

1. Navigate to `/blog`.
2. With JS enabled, you should see an ugly green div (possibly after a FOUC).

pic

3. Without JS enabled, the div remains unstyled and, in this case, less ugly.

pic

**Expected behavior**
The styles defined in the layout component and any of its child components should be included with the server-rendered page.

```
  System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 352.97 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.16.1 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Firefox: 81.0
    Firefox Developer Edition: 78.0
    Safari: 14.0
  npmPackages:
    rollup: ^2.3.4 => 2.28.2 
    sapper: ^0.28.9 => 0.28.9 
    svelte: ^3.29.0 => 3.29.0 
```

- Tested with Chrome, Firefox, Edge latest

- Local environment, Azure deployment

- Dynamic application

**Severity**
Blocker

**Additional context**
I appreciate the work you did on #1505 and I feel bad coming back with another CSS issue! It's definitely getting there though.

I did consider the issue described in #1572 but I am using Rollup 2. I'm curious if this could be in any way related to #1564.

rodoch:
Fixes #1579.

- [X] It's really useful if your PR references an issue where it is discussed ahead of time. In many cases, features are absent for a reason. For large changes, please create an RFC: https://github.com/sveltejs/rfcs
- [X] This message body should clearly illustrate what problems it solves.
- [X] Ideally, include a test that fails without this PR but passes with it.

### Tests
-  [X] Run the tests with `npm test` and lint the project with `npm run lint`


benmccann:
Thanks for this! Would it be possible to put the test in the existing `css` test directory? I.e. copy the `src/routes/nested/` directory into that project. Each new project we adds has to be compiled and start/stop a new server, so it will keep the tests faster if we minimize the number of test projects. If there's some conflict with the existing one then we wouldn't necessarily need to prioritize that though

rodoch:
No problem, that's done!

Issue #1575, opened 2020-10-01
Some bugs related to sapper cli dev, build and export options and flag --cwd


floratmin:
I am trying to integrate sapper into `nrwl/nx`. So i wanted to change the structure of the folders and run into some bugs and issues:
- `sapper build` has no static option, whereas `sapper export` and `sapper dev` have. I tried to change the reference to the static folder in the file `server.js`, but when building, the assets are not found (see my comment in the file `server.js`). Therefore it is not possible to move the static folder if a build is necessary.
- In `sapper build` you specify the output directory directly after `sapper build`, in `sapper dev` and `sapper export` you specify this option with the flag `--build-dir` which is inconsistent.
- I am using the sapper cli with the `--cwd` flag like `--cwd ./apps/web-page`. I have to specify an absolute path for the `--build-dir` flag to make the `--build-dir` option work. 
- When I use `sapper export` with the `--build-dir` option where the `__sapper__` folder points not to the `--cwd` directory then the build part is at the specified `--build-dir` directory, but the exported files are at the specified `--cwd` directory and there is no command line flag to specify a directory for the exported files. (Reproduce with `npm run export:builddir`)
- When using the `--cwd` flag the `rollup.config.js` specified at the path of the `--cwd` option is used. But to make `sapper dev`, `sapper build` or `sapper export` work i need to have a file named `rollup.config.js` in the root directory. This file can be completely empty.
- When using the 'build-dir' option with `sapper build`, starting the server with `node __sapper__/build` fails with `Error: ENOENT: no such file or directory, open '../../__sapper__/build/build.json'.` (Reproduce with `npm run build:builddir`)

In a follow up I will also test the sapper flags `--src` and `--routes`. But for these I have to set up different repositories.

The repo can be found under https://github.com/floratmin/sample-sapper-cli-cwd-build-dir-static-options
To recognize which folder is used for the static content i changed the success kid picture with a success kit picture.

**Expected behavior**
The folder structure should be adoptable to different needs.

**Information about your Sapper Installation:**

- The output of `npx envinfo --system --npmPackages svelte,sapper,rollup,webpack --binaries --browsers`
  System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: (8) x64 Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz
    Memory: 175.39 MB / 11.67 GB
    Container: Yes
    Shell: 4.4.19 - /bin/bash
[...]

Issue #922, opened 2019-10-02
Error building when using the --cwd option


WHenderson:
**Describe the bug**
When using the build option "--cwd", sapper fails trying to save the launcher "index.js".
The code for writing the launcher fails to take into account the cwd option and instead tries to write to an output folder which does not exist.

**Logs** / **Reproduction**
`sapper build test-output --cwd functions --src ../src --routes ../src/routes`
...
`> ENOENT: no such file or directory, open 'C:\Code\OwnPoC\poc-sapper\test-output\index.js'
Error: ENOENT: no such file or directory, open 'C:\Code\OwnPoC\poc-sapper\test-output\index.js'
    at Object.openSync (fs.js:448:3)
    at Object.writeFileSync (fs.js:1210:35)
    at prog.command.describe.option.option.option.option.option.option.option.option.example.action (C:\Code\OwnPoC\poc-sapper\node_modules\sapper\dist\cli.js:588:7)`

** Resolution **
In `cli.js`, replace `const launcher = path.resolve(dest, 'index.js');` with `const launcher = path.resolve(opts.cwd, dest, 'index.js');` to take account of the cwd option.

**Information about your Sapper Installation:**
Sapper v0.27.9

**Severity**
The only workaround for this requires hacking the mentioned file or forking the entire project.

**Additional context**
The objective for using the command line options above is to have sapper build correctly for use with google cloud functions.
The desired structure for google cloud functions is

| Path | Description |
| -- | ----------- |
| /functions/ | cloud functions base |
| /functions/index.js | cloud functions scaffold. This creates the primary entry point into the sapper express app |
| /functions/rollup.config.js | clone of /rollup.config.js because we changed the cwd and no command line option exists to reference the original |
| /functions/test-output/ | The built sapper output |

Note that everything sapper builds for the server needs to load files relative to `/functions` as that is the root as far as cloud functions are concerned. No other folders will be accessible.
This means that build_dir needs to be "test-output" and not "functions/test-output" which leads to all the headaches around cwd/src/routes etc.

Thanks in advance,
Sapper and Svelte are awesome!

mikeparisstuff:
I'm having a similar problem linked here https://github.com/sveltejs/sapper/issues/1074.

Currently working on a fix.

floratmin:
I made a fork which solves this problem. https://github.com/sveltejs/sapper/pull/1584

floratmin:
[...]

PR #1584, opened 2020-10-03
Make sapper flag --cwd for dev and build work when installation is in sub directory

floratmin:
This pull request solves the problem from https://github.com/sveltejs/sapper/issues/922 probably https://github.com/sveltejs/sapper/issues/1074 and some of the problems of https://github.com/sveltejs/sapper/issues/1575 .

With this commit it is possible to have the sapper installation in a sub directory and the build directory in a directory which can be located somewhere else.

The only thing to adjust in the project is the directory for loading the files from the static directory. For this we change the line to `sirv('./$cwd/static', { dev })` in the file `server.js` in the project created trough the sapper template.

The use of the `export_build` variable feels a little hacky to me, perhaps somebody finds a more elegant solution for this.

Many tests fail, but this fix works with a clean template and my sample template.
My sample template for testing the --cwd option can be found here: https://github.com/floratmin/sapper-subdirectory-example


src/core/create_app.ts
```diff
@@ -159,9 +163,11 @@ function generate_server_manifest(
 	manifest_data: ManifestData,
 	path_to_routes: string,
 	cwd: string,
+	root: string,
 	src: string,
 	dest: string,
-	dev: boolean
+	dev: boolean,
+	export_build: boolean
```
benmccann, line 170: Why don't you just pass in the correct value for `root` (i.e. `cwd` or the `root` directory) instead of adding this variable to check if `root` should be overriden?

benmccann:
It looks like this PR has merge conflicts

src/core/create_app.ts
floratmin: I thought that it would be more obvious what is happening to the path, than to morph `cwd` to `root` in `build.js` and using it as `root` in `create_app.ts`. If you prefer this approach I could change this. I could also update my sapper version to resolve the merge conflicts.

src/api/build.ts
```diff
@@ -30,21 +31,24 @@ export async function build({
...
 	bundler = undefined,
 	legacy = false,
 	ext = undefined,
 	oncompile = noop
 }: Opts = {}) {
-	bundler = validate_bundler(bundler);
 
+	const root = path.resolve('.');
 	cwd = path.resolve(cwd);
 	src = path.resolve(cwd, src);
```
benmccann, line 44: will this work? wouldn't it have to be relative to the root?

src/api/build.ts
```diff
@@ -30,21 +31,24 @@ export async function build({
 	output = 'src/node_modules/@sapper',
 	static: static_files = 'static',
 	dest = '__sapper__/build',
+	export_build = false,
```
benmccann, line 34: I'm not particularly a fan of having this `export_build` flag to handle one type of build differently. Could we pass in the `root` instead?

src/api/build.ts
floratmin: I need a reference to the root from where the build is executed. cwd can be supplied as an absolute or relative path. And furthermore in build.ts cwd is reassigned to its absolute path with `cwd = path.resolve(cwd);`. After this there is no possibility to get the root back. Instead of `path.resolve('.')` maybe using `process.cwd()` would be more clear. I chose the former because it is more in line with the handling of the other paths. An other option would be to use `path.relative(cwd, '.')`, to get the needed reference and then get the root with `path.resolve(cwd, path.relative(cwd, '.'));`. But i think we are here only turning in circles.

src/api/build.ts
floratmin: I removed it and created an extra variable pointing to the root which gets passed to the build function.

floratmin:
I also updated my sample template at https://github.com/floratmin/sapper-subdirectory-example and added scripts with the cwd flag as absolute path. Every combination works. In my nx workspace I also used the additional flags --routes and --static with no problems.

benmccann:
The lint check and tests look like they're failing

floratmin:
@benmccann linting is passing now and from 150 tests only one is failing. To make the tests passing I had to made an adjustment in the sapper code. Otherwise I would have to edit 23 test files and change all occurrences of `build({ cwd: __dirname })` to `build({ cwd: __dirname, root: __dirname })`. I thought that also other people could use tests in this way and therefore preferred this approach.

benmccann:
Thanks. Sorry this got dropped a bit. All our focus has been on SvelteKit. It might be better to make sure anything you want to do is supported there. I'm a bit hesitant to make many changes to Sapper except for bug fixes

benmccann:

PR #1586, opened 2020-10-05
Enable to specify host

brisbanewebdeveloper:
### Summary
- This amendment accelerates us to do dev work in case your development environment does not let you open `http://localhost:3000` right away, or you want to do dev work with an external server.
- In my case, I got stuck at `npm run dev` mentioned in https://github.com/sveltejs/sapper#get-started so I had to dig the code to figure how I can actually Get Started.
  + My Chrome somehow kept redirecting to `https://localhost:3000` from `http://localhost:3000`.
- Although I ended up being able to make it work without this amendment, I thought it is better to let us just suggest to try using this option when someone else claimed about the hardcoded hostname so that we can avoid spending time for the case.

### Use Cases

Amend `package.json` before/after `npm install` at https://github.com/sveltejs/sapper#get-started

#### Use Hostname

```
"scripts": {
  "dev": "sapper dev --host my.external-server",
```

#### Use IP Address

```
"scripts": {
  "dev": "sapper dev --host 192.168.0.100",
```

### Tests

### DIFF
--- a/src/api/dev.ts
+++ b/src/api/dev.ts
@@ -27,6 +27,7 @@ type Opts = {
 	hot?: boolean;
 	'devtools-port'?: number;
 	bundler?: 'rollup' | 'webpack';
+	host?: string;
 	port?: number;
 	ext: string;
 };
@@ -45,6 +46,7 @@ class Watcher extends EventEmitter {
 		output: string;
 		static: string;
 	}
+	host: string;
 	port: number;
 	closed: boolean;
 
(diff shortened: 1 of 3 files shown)


benmccann:
Thanks for putting together this PR! I think this feature is actually present in SvelteKit

Sapper never had a good method of handling configuration, which made it difficult to know how to integrate this feature. However, that's been addressed in SvelteKit by adding a `svelte.config.cjs` file and adapters. I'm going to go ahead and close this PR since we won't be adding the feature to Sapper. You might like to check out SvelteKit instead. We'll be reviewing PRs for missing features there

PR #1590, opened 2020-10-07
Service worker would fail if accessing static assets with query string

### LINKED ISSUE #1587: Sapper 0.28 support for query parameters for static files is broken
**Describe the bug**
After update to Sapper 0.28 links to static files are broken if link has request parameters.
Useful for custom fonts or images caching.

**Logs**
Please include browser console and server logs around the time this bug occurred.

**To Reproduce**
Create new project with 
npx degit "sveltejs/sapper-template#webpack" my-app

run test app and try to open 
http://localhost:3000/successkid.jpg?test

works only once or after hard reload (CtrF5), i.e. trying to refresh using (CtrR or F5) shows page:
This site can’t be reached
The webpage at http://localhost:3000/successkid.jpg?test might be temporarily down or it may have moved permanently to a new web address.

in the same time
http://localhost:3000/successkid.jpg
works fine

The same on sapper 0.27 works fine

**Expected behavior**
Expected to properly work with query params for static files.

**Severity**
Issue blocking update from 0.27 to to 0.28

ehrencrona:
Fixes #1587. The service worker would assume that is the request path matches an asset, then the asset is in the cache. It would then use `CacheStorage.match` to retrieve the request, but that would fail if the query string was different.

I don't think there's any reasonably simple way of writing a test for this unfortunately.

### Tests

### DIFF
--- a/site/src/service-worker.js
+++ b/site/src/service-worker.js
@@ -45,7 +45,7 @@ self.addEventListener('fetch', event => {
 
 	// always serve static files and bundler-generated assets from cache
 	if (url.host === self.location.host && cached.has(url.pathname)) {
-		event.respondWith(caches.match(event.request));
+		event.respondWith(caches.match(event.request, { ignoreSearch: true }));
 		return;
 	}
 


Conduitry:
We can also make this change here if we want, but the real fix for this would be to change the file in sapper-template. This is just the code for Sapper's own website.

ehrencrona:
Oops, that's embarrassing. I just grep'ed for the line and only found this one and jumped to conclusions.

PR #1594, opened 2020-10-09
Fixed #1592 by ensuring `scroll` exists

### LINKED ISSUE #1592: "scroll" is sometimes undefined in scrollTo(scroll.x, scroll.y)
**Describe the bug**
I got this error on our production website, tracked using Sentry:

```
TypeError: Cannot read property 'x' of undefined
  at ? (/client/client.8caa5d97.js:16:2552)
  at Generator.next(<anonymous>)
  at a(/client/client.8caa5d97.js:16:86)
```

Here's the link: https://sentry.io/share/issue/20ae1300a33f48f58457ffce7c96925a/

**Fix**
I was able to zero down on the line here: https://github.com/sveltejs/sapper/blob/master/runtime/src/app/router/index.ts#L223. It seems like sometimes `let scroll = scroll_history[id];` might be undefined, and then calling `scrollTo(scroll.x, scroll.y)` throws an error. A proposed fix for this is to change:

```js
if (popstate || deep_linked) {
  scrollTo(scroll.x, scroll.y);
}
```

to:

```js
if (scroll && (popstate || deep_linked)) {
  scrollTo(scroll.x, scroll.y);
}
```

**Stacktraces**

Screen Shot 2020-10-08 at 13 30 45

- The output of `npx envinfo --system --npmPackages svelte,sapper,rollup,webpack --binaries --browsers`

<details>
<summary>Expand</summary>
<pre>System:
    OS: macOS 10.15.6
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 1002.30 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.5.0 - /usr/local/bin/node
    npm: 6.14.6 - ~/Projects/koj/koj/node_modules/.bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Safari: 14.0
  npmPackages:
    rollup: ^2.29.0 => 2.29.0 
    sapper: ^0.28.10 => 0.28.10 
</pre>
</details>

- Browser details from Sentry:

Screen Shot 2020-10-08 at 13 31 37

- Your hosting environment: AWS EC2 running a Docker container that serves the full-stack Sapper server

- If it is an exported (`npm run export`) or dynamic application: Dynamic

AnandChowdhary:
This PR fixes #1592 by ensuring that the variable `scroll` exists before accessing its `x` and `y` properties.

### Tests

### DIFF
--- a/runtime/src/app/router/index.ts
+++ b/runtime/src/app/router/index.ts
@@ -219,7 +219,7 @@ export async function navigate(dest: Target, id: number, noscroll?: boolean, has
 		}
 
 		scroll_history[cid] = scroll;
-		if (popstate || deep_linked) {
+		if (scroll && (popstate || deep_linked)) {
 			scrollTo(scroll.x, scroll.y);
 		} else {
 			scrollTo(0, 0);


benmccann:
I think we really need a set of steps that can be followed to reproduce this bug to make sure we're fixing the actual issue and not just papering over it. We should probably have a test for this change as well once we have a minimal reproduction.

AnandChowdhary:
I agree, I haven't been able to locally reproduce this either, I like others only got to capture it in production in #1592. I think I tried going back and forward in the browser really fast when it was thrown.

The idea behind this PR is to just not try scrolling if the `scroll` value doesn't exist, which is something we should do anyway, right?

benmccann:
@rcabralc thanks for coming up with a way to reproduce this error (https://github.com/sveltejs/sapper/issues/1592#issuecomment-711433134). Given your description, I think this is probably a reasonable fix. What are your thoughts on it?

rcabralc:
I think this is reasonable. Scrolling is not very critical, so it's probably fine just skip it in this case.

AnandChowdhary:
@rcabralc, wonderful! @benbarbersmith, do you think you can review and merge this? 😄

benmccann:
I've finally gotten some time to come back to this and read through things in more detail. It looks like `pushState` will let you push any arbitrary serializable state, so I think the best fix for this would be to push the scroll location when we call `pushState` or `replaceState` instead of storing it in `scroll_history`, which could be entirely removed. Then when we navigate offsite we won't lose the scroll history and when `handle_popstate` is called we will have the scroll location available.

This PR is definitely better than the status on `master`, but I'm also worried of losing track of this as something that should get implemented and would prefer to fix it the right way while it's fresh in everyone's minds.

Issue #1556, opened 2020-09-24
Improve TypeScript type definition for stores


albertms10:
**Is your feature request related to a problem? Please describe.**
The `stores` function from `@sapper/app` has an `unknown` return type, although there is a known object returning some predefined values such as `preloading`, `page`, `session`, as seen in the [Sapper Docs](https://sapper.svelte.dev/docs#Stores).
```svelte
<script>
  import { stores } from '@sapper/app';
  const { preloading, page, session } = stores(); // => `unknown`, destructured items are `any`

  const delayedPreloading = derived(
    preloading,
    (currentPreloading, set) => { // as preloading is treated as `any`, TypeScript cannot infer types for fn parameters
      // ...
    }
  );

  // ...
</script>
```

**Describe the solution you'd like**
Adding types for these stores would improve DX.

**Describe alternatives you've considered**
Just ignore TypeScript warning about the `any` type in use cases.

benmccann:
This is defined here:

https://github.com/sveltejs/sapper/blob/f5032a524e2b674e7c81a1f5606314b8af474289/runtime/index.d.ts#L15

TheComputerM:
Can I try to fix this issue and open a PR?

albertms10:
Of course!

PR #1601, opened 2020-10-13
add types for stores

TheComputerM:
Fixes https://github.com/sveltejs/sapper/issues/1556

### Tests


runtime/index.d.ts
```diff
@@ -39,15 +46,9 @@ declare module '@sapper/common' {
...
 	}
 
-	export interface Page {
-		host: string;
-		path: string;
-		params: Record<string, string>;
-		query: Record<string, string | string[]>;
-		error?: Error;
-	}
+	export { PageContext as Page } from '@sapper/app/types';
```
TheComputerM, line 49: Also imported PageContext from `@sapper/app/types` as to not re-declare it.

ehrencrona (approved):
LGTM :)

bjon:
`import { PageContext } from "@sapper/app/types";` doesn't seem to work with vscode. It still shows:
```
(alias) stores(): {
    preloading: Readable<boolean>;
    page: Readable<any>;
    session: Writable<any>;
}
import stores
```

bjon:
server.ts is created like this:
```
polka() // You can also use Express
	.use(
		sirv('static', { dev }),
		sapper.middleware()
	)
	.listen(PORT, err => {
	});

```
so `opts` in middleware, should be optional. 

`export function middleware(opts?: MiddlewareOptions): Handler;`

TheComputerM:
> `import { PageContext } from "@sapper/app/types";` doesn't seem to work with vscode. It still shows:
> 
> ```
> (alias) stores(): {
>     preloading: Readable<boolean>;
>     page: Readable<any>;
>     session: Writable<any>;
> }
> import stores
> ```

I think PageContext has to be a type instead of an interface for this to work, please correct me if I am wrong.

bjon:
@TheComputerM Both type or interface should work. I just don't understand the "magic" behind `@sapper/app/types`. 
How can vscode access it?

TheComputerM:
So, shall I convert `@sapper/app/types` to `./src/app/types`?

bjon:
That file doesn't exist either?

The only way I can get it to work, is to declare the namespace, directly in the index.d.ts
```
declare module '@sapper/app/types' {
  export interface PageContext {
    host: string;
    path: string;
    params: Record<string, string>;
    query: Record<string, string | string[]>;
    error?: Error;
  }
}
```

But now we're back to square one again :)

TheComputerM:
This is the [file](https://github.com/sveltejs/sapper/blob/5f4097a870e11b05a13e485e824ea73bebc53908/runtime/src/app/types.ts)

bjon:
Yes, but only in the source. I can't find it anywhere in `node_modules/sapper` or `src/node_modules/@sapper`.

If its not included in the sapper dist, or auto-generated @sapper, I dont see how we can import it.

TheComputerM:
So how do we fix this?

bjon:
If `@sapper/app/types`  and `@sapper/internal/manifest-server` are supported to work, it might be a bug when sapper generate the code?

bjon:
One workaround would be to add
```
declare module '@sapper/app/types' {
  export interface PageContext {
    host: string;
    path: string;
    params: Record<string, string>;
    query: Record<string, string | string[]>;
    error?: Error;
  }
}
```
and remove it, when the import works?

bjon:
But im no expert here. I just want auto-complete to work :)

TheComputerM:
Ok, If someone approves this, I will add this.

bjon:
And also, `export function start(opts: { target: Node }): Promise<void>;` should be `export function start(opts: { target: Element }): Promise<void>;`, cause `document.querySelector` returns an `Element`
```
sapper.start({
  target: document.querySelector(`#sapper`),
});
```

bjon:
To avoid compile error its should probably be `export function start(opts: { target: Element | null }): Promise<void>;`.
`document.querySelector` returns `Element | null`

AnandChowdhary:
> To avoid compile error its should probably be `export function start(opts: { target: Element | null }): Promise<void>;`.
> `document.querySelector` returns `Element | null`
> 
> or maybe better, change sapper-template to convert TS to
> 
> ```
> sapper.start({
>   target: document.querySelector(`#sapper`) as Element,
> });
> ```

This is what we do, maybe it's better to ensure it exists instead of typecasting?

```ts
const target = document.querySelector("#sapper");

if (target)
  sapper.start({
    target,
    hydratable: true,
  });
```

bjon:
Found one more. In `@sapper/app` goto(opts), should be optional

ehrencrona:
I looked into how the types work: in the `tsconfig.json` `@sapper` is defined as an alias for the directory `runtime/src`:

```
"paths": {
	"@sapper/*": ["runtime/src/*"]
}
```

This is the reason why `@sapper/internal` references work in Sapper code, but not in project code. In project code there simply is no such thing; the only thing resolved under `@sapper` is what we put in the `index.d.ts`.

Therefore, those references from `index.d.ts` to `@sapper/internal` must be removed. I created #1604 for this. Once that is merged, it should be possible to apply this PR on top of those changes (though I'm afraid that PR might cause some controversy and could take time to merge).

ehrencrona:
On the discussion around whether `target` in `start` is `Node` or an `Element`: An `Element` *is* a `Node`; `Node` is the most abstract class. `start` can handle `Node`s so it should be `target: Node`.

Furthermore, I don't think `start` can handle `target` being `null`. If you can't find the DOM node you want to start the app in, that's an error, which is essentially what the type mismatch says.

runtime/index.d.ts
```diff
@@ -1,14 +1,21 @@
...
+	export function goto(href: string, opts?: { noscroll?: boolean, replaceState?: boolean }): Promise<void>;
 	export function prefetch(href: string): Promise<{ redirect?: Redirect; data?: unknown }>;
 	export function prefetchRoutes(pathnames: string[]): Promise<void>;
 	export function start(opts: { target: Node }): Promise<void>;
-	export const stores: () => unknown;
+	export function stores(): {
+		preloading: Readable<boolean>
+		page: Readable<PageContext>
+		session: Writable<any>
+	};
```
bencates, line 18:
```suggestion
	export function stores<Session = any>(): {
		preloading: Readable<boolean>
		page: Readable<PageContext>
		session: Writable<Session>
	};
```

runtime/index.d.ts
```diff
@@ -21,7 +28,7 @@ declare module '@sapper/server' {
 		ignore?: Ignore
 	}
 
-	export function middleware(opts: MiddlewareOptions): Handler;
+	export function middleware(opts?: MiddlewareOptions): Handler;
```
bencates, line 31:
```
	export interface MiddlewareOptions<Session = unknown> {
		session?: (req: Req, res: Res) => Session
		ignore?: Ignore
	}
```

then

```suggestion
	export function middleware<Session = unknown>(opts?: MiddlewareOptions<Session>): Handler;
```

runtime/index.d.ts
```diff
@@ -39,15 +46,9 @@ declare module '@sapper/common' {
...
-		params: Record<string, string>;
-		query: Record<string, string | string[]>;
-		error?: Error;
-	}
+	export { PageContext as Page } from '@sapper/app/types';
 
 	export interface Preload {
-		(this: PreloadContext, page: Page, session: any): object | Promise<object>;
+		(this: PreloadContext, page: PageContext, session: any): object | Promise<object>;
 	}
```
bencates, line 53:
```
	export interface Preload<Props = object, Session = any> {
		(this: PreloadContext, page: PageContext, session: Session): Props | Promise<Props>;
	}
```

bencates (commented):
Might be nice to parameterize these types to let users type the session and preloaded props.

walterCybot (approved):

benmccann:
@TheComputerM this PR will need to be rebased

runtime/index.d.ts
```diff
@@ -28,7 +28,7 @@ declare module '@sapper/server' {
 		ignore?: Ignore
 	}
 
-	export function middleware(opts?: MiddlewareOptions): Handler;
+	export function middleware<Session = unknown>(opts?: MiddlewareOptions<Session>): Handler;
```
bencates, line 31: I don't think this works correctly unless we also parameterize `MiddlewareOptions`

bencates (changes requested):
Sorry, this might be my fault. GitHub would only allow me to format half of my proposed changes as suggestions.

runtime/index.d.ts
```diff
@@ -6,16 +6,23 @@
  */
 
 declare module '@sapper/app' {
+	import { Readable, Writable } from 'svelte/store';
+	import { PageContext } from '@sapper/app/types';
```
benmccann, line 10: `PageContext` is defined lower down in this file in `@sapper/common`

PR #1604, opened 2020-10-15
Do not reference any internal types in runtime/index.d.ts

ehrencrona:
Prompted by the discussion in #1601 I started looking into how types are exposed to project code and realized the `runtime/index.d.ts` cannot work in project code because it references `@sapper/internal` types that are not available to it.

I therefore removed those dependencies by:

 * Moving `PageContext` into the file. I left `Page` in `types` since I'd consider it internal. For it to be possible for Sapper code to reference `PageContext` under the namespace `@sapper/common` there needs to be a `common` source file so I created an empty one. This is a bit ugly, but I see no decent way around it (except perhaps duplicating `PageContext` but there's a risk of the definitions drifting apart).
 * Moving `Req` and `Res`. `Req` was wrongly inherited from `http.ClientRequest`; it should be `http.IncomingMessage`. I removed the fields declared in them that are already defined by the parent class.

The `Preload` type was declared in `shared.d.ts` in addition to `runtime/index.d.ts`, which seems unnecessary, so I removed it.

`shared.d.ts` also contained `Params`, which wasn't used anywhere, and `Query`, that was used in a single file, so I removed those.

We were exporting a `Page` interface in `index.d.ts` that is called `PageContext` in the Sapper code. Since there is another interface called `Page` we can't rename it throughout. Therefore, I'm exporting it under both names but marking `Page` as deprecated. Since the file was essentially broken I think it's unlikely anyone is using `Page` and we should remove it in the next major release IMO.


runtime/src/app/router/index.ts
```diff
@@ -68,7 +67,7 @@ export function init(base: string, handler: (dest: Target) => Promise<void>): vo
 }
 
 export function extract_query(search: string) {
-	const query = Object.create(null);
+	const query: Record<string, string | string[]> = Object.create(null);
```
benmccann, line 70: can we keep the `Query` type?

runtime/index.d.ts
```diff
@@ -12,16 +19,64 @@ declare module '@sapper/app' {
...
+	import { TLSSocket } from 'tls';
 
 	export type Ignore = string | RegExp | ((uri: string) => boolean) | Ignore[];
 
+	/**
+	 * The request object passed to middleware and server-side routes. 
+	 * These fields are common to both Polka and Express, but you are free to 
+	 * instead use the typings that come with the server you use.
+	 */
+	export interface Req extends IncomingMessage {
```
benmccann, line 32: should we call this `Request` instead of `Req`?

benmccann: Actually, if we just call them `Request` and `Response` do we need to specify the types here? See the comment in https://github.com/sveltejs/sapper/pull/1603#issuecomment-709429933

ehrencrona: Yeah, these namings are very confusing. `Request` and `Response` are the types used by `fetch`. They are similar but unrelated to the request / response types used by `http.Server` which are called `IncomingMessage` and `ServerResponse`.

Polka and Express, in turn, each extend the `IncomingMessage` in different ways, but with a core set of attributes that are identical in both. Those are the ones we have in `Req`.

That said, I'd definitely welcome a renaming. How about `SapperRequest` / `SapperResponse`?

runtime/src/app/router/index.ts
benmccann: I think it should also be used in `PageContext`?

runtime/index.d.ts
benmccann: Ah, I see. Ok. `SapperRequest` and `SapperResponse` sound fine to me

ehrencrona: Ok, I renamed them.

runtime/src/app/router/index.ts
ehrencrona: Fair enough. I put it back together with `Params` that I renamed `PageParams`

runtime/index.d.ts
```diff
@@ -12,16 +19,64 @@ declare module '@sapper/app' {
...
 	export interface MiddlewareOptions {
-		session?: (req: Req, res: Res) => unknown
-		ignore?: Ignore
+		session?: (req: SapperRequest, res: SapperResponse) => unknown;
+		ignore?: Ignore;
 	}
 
-	export function middleware(opts: MiddlewareOptions): Handler;
+	export function middleware(
+		opts: MiddlewareOptions
```
bjon, line 78: `opts` is still optional, right? Should probably be `opts?`.

ehrencrona: True. It doesn't really relate to this PR but I'll change it here anyway. Thanks!

runtime/index.d.ts
```diff
@@ -39,15 +94,26 @@ declare module '@sapper/common' {
 		redirect: (statusCode: number, location: string) => void;
 	}
 
-	export interface Page {
+	export type PageParams = Record<string, string>;
+	export type Query = Record<string, string | string[]>;
+	
+	export interface PageContext {
```
benmccann, line 100: I'd probably rather save the nice name `Page` for external usage and rename the internal `Page` type to something like `PageInfo`

ehrencrona: I'd be fine with that, but I generally feel this *isn't* a "page". The page parameters really parametrize a route, so I guess a page is more or less the same as a route, This object is attached to a request, whereas a page feels like something more permanent, at least to me.

What about `PageRequest` (a request for a page)? The request carries the parameters for the page, the `PageParams`. That kind of makes sense.

benmccann: It's not really a `PageRequest` because of the `error`. I guess `PageContext` makes sense

Issue #1179, opened 2020-05-02
page.query not properly encoding params


chovy:
**Describe the bug**
page.query not properly encoding/decoding params

**Logs**
`https://www.amazon.com/Electronic-Arts-Titanfall-Instant-Access/dp/B01H0LFN3K/ref=as_li_ss_tl?ascsub=`

**To Reproduce**

Pass in an encoded url with parameters to ``/?link=<encoded-url-here>``

Use this url: `https://www.amazon.com/Electronic-Arts-Titanfall-Instant-Access/dp/B01H0LFN3K/ref=as_li_ss_tl?&ascsub&linkCode=sl1&tag=foo-20&linkId=394efa83730b482f01461894f50da53f&language=en_US`

it should be `/?link=encodeURIComponent(url)` when you send it

**Expected behavior**

```
<script context="module">
  export async function preload(page, session) {
    let scoops
    const { title, link, text, category } = page.query;

    console.log(link);
    if (text) scoops = 'text';
    
    return { title, link, text, category, scoops };
  }
</script>

```

console.log(link) here does not have any parameters after it. 

**Information about your Sapper Installation:**

Ubuntu 18.04

`    "sapper": "^0.27.0",`

**Severity**

I'd say not properly handling encoded url parameters is a major bug

**Additional context**

Demo link of issue:

`https://coupondeals.life/compose?link=https%3A%2F%2Fwww.amazon.com%2FElectronic-Arts-Titanfall-Instant-Access%2Fdp%2FB01H0LFN3K%2Fref%3Das_li_ss_tl%3F%26ascsub%26linkCode%3Dsl1%26tag%3Dfoo-20%26linkId%3D394efa83730b482f01461894f50da53f%26language%3Den_US`

chovy:
has anyone had a chance to look at this?

Issue #1174, opened 2020-04-28
Non-latin route not recognized


datio:
Hello, I couldn't make the router to accept Greek characters.

The following router path works:
./src/routes/consent/gdpr.svelte

This one doesn't (_getting 404_):
./src/routes/**συναίνεση**/gdpr.svelte

I have tried to URL encode that filename on Ubuntu 18.04 which didn't fix the problem.

benmccann:
Related to or possibly duplicate of https://github.com/sveltejs/sapper/issues/1002

benmccann:
Closing in favor of https://github.com/sveltejs/kit/issues/2000

Issue #1002, opened 2019-11-25
Router fails to resolve URL-encoded links


c0ldra1n:
**Describe the bug**
Sapper router fails to route links with URL-encoded characters. That is, links that contain non-ascii characters like 가나다라(hangul).

**Logs**
``GET http://localhost:3000/posts/%ED%85%8C%EC%8A%A4%ED%8A%B8 404 (Not Found)``
which translates to http://localhost:3000/posts/테스트 and src/routes/posts/테스트.svelte exists.

**To Reproduce**
Use any links/routes that would produce a link with encoded characters. The generated code will include the encoded characters.

**Expected behavior**
The link should resolve according to the url-encoded string.

**Stacktraces**
No stacktraces were logged

**Information about your Sapper Installation:**

- Chrome 78.0.3904.108 (Official Build) (64-bit)

- macOS 10.14.6 (18G1012)

- Hosted locally

- Sapper version 0.27.9

- Svelte version 3.15.0

- Ran as dynamic application on dev environment

- Rollup

**Severity**
Such URL resolving technique is crucial for international audiences / developers of Sapper based websites.

**Additional context**
This issue can be simply resolved by adding a url-encoding function to all route related regex tests. I will add a pull request later on.

PR #1620, opened 2020-10-30
Fix runtime router handling of encoded query parameters

bermi:
Hello Sapper team! Thanks for creating this fantastic framework.

I found an issue where a link such as 

    <a href="/?multiline=multi%0Aline">link</a>

will cause runtime routes to parse the query parameters as

    { multiline: "multi" }

instead of

    { multiline: "multi\nline" }

It looks like sapper@0.6.0 removed the dependency from [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams) with a custom RegExp to accommodate IE11. Using `URLSearchParams` prevents this and potentially future query parsing bugs from happening, so this PR tries to use `URLSearchParams` when available and fixes the old RegExp to allow.

To fix current query parser adding the [RegExp/dotAll `/s`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) would be sufficient but unfortunately, it does not work on IE11 or FF Android. So the `[\s\S]*` RegExp tick is used as opposed to `.*`.

Since `URLSearchParams` uses iterables, I also had to add `dom.iterable` to the `tsconfig.json` file to get the tests passing. Since I am new to `TypeScript` I am not sure what the implication might be for the final sapper client-side bundle.

I have tested the IE11 code by disabling the `URLSearchParams` block. Are IE11 tests executed automatically by the CI pipeline before pushing a release?

### DIFF
--- a/runtime/src/app/router/index.ts
+++ b/runtime/src/app/router/index.ts
@@ -67,17 +67,30 @@ export function init(base: string, handler: (dest: Target) => Promise<void>): vo
 	addEventListener('popstate', handle_popstate);
 }
 
-export function extract_query(search: string) {
+// IE11 does not support URLSearchParams so we'll fall back to a custom
+// RegExp that mimics the standard URLSearchParams method
+const _get_query_array = (search: string): string[][] => {
+	if (typeof URLSearchParams !== 'undefined') {
+		return 	[...new URLSearchParams(search).entries()];
+	}
+	return search.slice(1).split('&').map(searchParam => {
+		// Instead of `.*` we'll use \s\S to allow characters and non characters
+		// such as [\r\n\v\f]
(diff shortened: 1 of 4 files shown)


ehrencrona (approved):
This makes a lot of sense to me!

I'm fairly sure adding `dom.iterable` to the `tsconfig.json` is harmsless.

benmccann:
There are several potentially related bugs open in the tracker: https://github.com/sveltejs/sapper/issues/1179 https://github.com/sveltejs/sapper/issues/1174 https://github.com/sveltejs/sapper/issues/1002. Would this PR fix any of them?

This seems like a good fix. I'm slightly hesitant to merge it because it will make Sapper's code diverge more from SvelteKit's. I expect SvelteKit would not hit this bug because it uses `UrlSearchParams`. But I think overall it's worth it because it adds a test and it'd be nice to eventually port that test to SvelteKit to prevent a regression there

PR #1644, opened 2020-11-23
Supply `req.url` value as `should_ignore` fallback

featherbear:
When used with the Fastify compatibility layer for Express-like plugins ([`middie`](https://github.com/fastify/middie)), the value of `req.path` is undefined which causes an error.

This PR proposes to add `req.url` as a fallback path in the event that `req.path` is undefined - fixing the optional ignore option

### Tests

### DIFF
--- a/runtime/src/server/middleware/index.ts
+++ b/runtime/src/server/middleware/index.ts
@@ -80,7 +80,7 @@ export function compose_handlers(ignore: IgnoreValue, handlers: Handler[]): Hand
 	return !ignore
 		? (req, res, next) => nth_handler(0, req, res, next)
 		: (req, res, next) => {
-			if (should_ignore(req.path, ignore)) {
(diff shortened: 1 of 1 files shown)


benmccann:
Thanks for the proposed change. In SvelteKit we are adding explicit support for other services like Fastify. Since we're fully focused on SvelteKit and this may already be fixed there I'm going to hold off on reviewing this PR

featherbear:
> 
> 
> Thanks for the proposed change. In SvelteKit we are adding explicit support for other services like Fastify. Since we're fully focused on SvelteKit and this may already be fixed there I'm going to hold off on reviewing this PR

No worries! Looking forwards to SvelteKit!

benmccann:
I'm going to go ahead an close this now that SvelteKit is out. Hopefully it will be possible to create a Fastify adapter there

PR #1652, opened 2020-12-01
Allow layout components to be grouped together

### LINKED ISSUE #708: src/routes/_layout/ directory
Having a `src/routes/_layout` directory would allow layout components to be grouped with `src/routes/_layout/index.svelte`.

fehnomenal:
Fixes #708 by trying to load `src/routes/_layout.{ext}` and `src/routes/_layout/index.{ext}` before falling back to the default layout.

Works locally by using `npm link`.

### DIFF
--- a/src/core/create_manifest_data.ts
+++ b/src/core/create_manifest_data.ts
@@ -167,7 +167,7 @@ export default function create_manifest_data(cwd: string, extensions = '.svelte 
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


benmccann:
Thanks for the suggestion. We probably don't want to add any new features to Sapper that aren't in SvelteKit because we don't want people to lose functionality when upgrading. It'd probably be best to raise the idea there and see what people think of it

Issue #1660, opened 2020-12-07
changing imported css will lead to unnecessary dev reload


dionysiusmarquis:
**Describe the bug**
When you are changing the contents of an imported css in dev, you will see the changes immediately, but after a split second the page will be reloaded with the previous version. This is most likely because the js files file name (e.g. `my-component-{hash}.js`) will remain the same even though the corresponding `my-component-{hash}.css` changed. My guess is that the css file gets updated correctly, which will be applied immediately, but then a (unnecessary) page reload get's triggered. Because the js files hash is still the same even though the css contents changed (maybe because of the `inject_css` stuff happening on top of the rollup build pipeline?), the service worker will provide an old, cached version of that js file.

**To Reproduce**
```html
// _layout.svelte
<script>
  import 'src/styles/test.css';
</script>
```

In Browser dev tools:
```js
// client.cbe62f53.js
const components = [
  {
    js: () => Promise.all([…, __inject_styles(["client-f2a00d8f.css"])]).then(function(x) { return x[0]; })
  },
  …
```

after changing `src/styles/test.css` contents and "automated" page reload (wrong, still same css file hash)
```js
// client.cbe62f53.js
const components = [
  {
    js: () => Promise.all([…, __inject_styles(["client-f2a00d8f.css"])]).then(function(x) { return x[0]; })
  },
  …
```

After "manual" page reload without any further css content change (correct css file hash)
```js
// client.cbe62f53.js
const components = [
  {
    js: () => Promise.all([…, __inject_styles(["client-d6af0d49.css"])]).then(function(x) { return x[0]; })
  },
  …
```

**Expected behavior**
I think it would already be resolved if the page wouldn't reload on imported css change. The Service worker cache would be "out of sync" but that shouldn't have any side effects, I guess.

**Information about your Sapper Installation:**
```
  Binaries:
    Node: 14.15.1 - ~/.nvm/versions/node/v14.15.1/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.15.1/bin/npm
  Browsers:
    Chrome: 87.0.4280.88
    Safari: 14.0.1
  npmPackages:
    rollup: ^2.3.4 => 2.34.0 
    sapper: ^0.28.0 => 0.28.10 
    svelte: ^3.17.3 => 3.30.1 
```

**Severity**
It's annoying because you always have to manually reload the page after the unnecessary "automated" page reload

[...]

PR #1665, opened 2020-12-09
Augmented chunk hashes based on imported css code

dionysiusmarquis:

### Tests
-  [ ] Run the tests with `npm test` and lint the project with `npm run lint`

Fixes: #1660 

Currently the chunk hashes don't include "injected" css code coming from script css imports. This is because the injection is happening in bundle hooks and at this stage all depending chunk imports will already be resolved using hashes that don't "react" to css changes. This is especially crucial in caching environments.
To ensure that the chunk hashes also change as soon as imported css gets updated this PR adds a `css_hashing` plugin to the rollup compiler. The `augmentChunkHash` hook is utilised to provide a hash based on all imported css codes.

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -1,3 +1,4 @@
+import { createHash, Hash } from 'crypto';
 import * as path from 'path';
 import color from 'kleur';
 import relative from 'require-relative';
@@ -9,6 +10,7 @@ import {
 	OutputChunk,
 	Plugin,
 	PluginContext,
+	PreRenderedChunk,
 	RenderedChunk,
 	RollupError
 } from 'rollup';
@@ -160,13 +162,44 @@ export default class RollupCompiler {
 
 		const that = this;
 
-		// TODO this is hacky. refactor out into an external rollup plugin
-		(mod.plugins || (mod.plugins = [])).push(css_chunks({
-			entryFileNames: '[name]-[hash].css'
-		}));
-		if (!/[\\/]client\./.test(entry_point)) {
-			return mod;
-		}
+		/**
+		 * Track css code hashes for proper chunk hashing
+		 */
+		const css_hashing = () => {
+			const css_hashes: Record<string, string> = {};
+			return <Plugin>{
+				name: 'sapper-css-hashing',
+				augmentChunkHash(this: PluginContext, chunk: PreRenderedChunk) {
+					// Augment chunk hash based on saved css hashes  
+					let hash: Hash;
+					for (const moduleId in chunk.modules) {
+						const css_hash = css_hashes[moduleId];
+						if (css_hash) {
+							if (!hash) {
+								hash = createHash('sha256');
+							}
+							hash.update(css_hash);
+						}
+					}
+					if (hash) {
+						return hash.digest('hex');
+					}
+				},
+				transform(this: PluginContext, code: string, id: string) {
+					const module_path = path.parse(id);
+					if (module_path.ext === '.css') {
+						// Ignore svelte css files created with `emitCss` enabled. These css files are generated out
+						// of svelte's `<style>` and already part of the code that was used to generate the chunk hash
+						const is_svelte_css = !!this.getModuleInfo(`${module_path.dir}/${module_path.name}.svelte`);
+						if (!is_svelte_css) {
+							// save css code based hashes
+							css_hashes[id] = createHash('sha256').update(code).digest('hex');
+							return { code, map: null, moduleSideEffects: 'no-treeshake' };
+						}
+					}
+				}
+			};
+		};
 
 		/**
 		 * Finds dynamic imports and rewrites them to import the component and its CSS in parallel
@@ -308,6 +341,14 @@ export default class RollupCompiler {
 			}
 		};
 
+		// TODO this is hacky. refactor out into an external rollup plugin
+		(mod.plugins || (mod.plugins = [])).push(css_hashing());
+		mod.plugins.push(css_chunks({
+			entryFileNames: '[name]-[hash].css'
+		}));
+		if (!/[\\/]client\./.test(entry_point)) {
+			return mod;
+		}
 		mod.plugins.push(css_injection);
 		mod.plugins.push(sapper_internal);
 


src/core/create_compilers/RollupCompiler.ts
```diff
@@ -109,6 +111,7 @@ export default class RollupCompiler {
 	errors: any[];
 	chunks: RenderedChunk[];
 	css_files: Record<string, string>;
+	css_hashes: Record<string, string>;
```
benmccann, line 114: can you keep this variable contained within the `css_hashing` plugin?

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -300,6 +333,14 @@ export default class RollupCompiler {
 			}
 		};
 
+		mod.plugins.push(css_hashing);
```
benmccann, line 336: line 339 does `mod.plugins || (mod.plugins = [])`. it seems like this should come after that. maybe something like:
```
mod.plugins = mod.plugins || [];
mod.plugins.push(css_hashing);
mod.plugins.push(css_chunks({ injectImports: true }));
```

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -154,11 +158,40 @@ export default class RollupCompiler {
...
+		const css_hashing: Plugin = {
+			name: 'sapper-css-hashing',
+			augmentChunkHash(this: PluginContext, chunk: PreRenderedChunk) {
+				let hash: Hash;
+
+				const module_ids = Object.keys(chunk.modules).map(module_id => ({ module_id, module_path: path.parse(module_id) }));
+				module_ids.forEach(({ module_id, module_path }) => {
+					if (that.css_hashes[module_id]) {
+						const file = `${module_path.dir}/${module_path.name}`;
+						// ignore svelte css files created with `emitCss` enabled
```
benmccann, line 173: this comment could be expanded. why ignore those files and not other css files? also, how does this check work?

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -154,11 +158,40 @@ export default class RollupCompiler {
...
+		/**
+		 * Track css code hashes for proper chunk hashing
+		 */
+		const css_hashing: Plugin = {
+			name: 'sapper-css-hashing',
+			augmentChunkHash(this: PluginContext, chunk: PreRenderedChunk) {
+				let hash: Hash;
+
+				const module_ids = Object.keys(chunk.modules).map(module_id => ({ module_id, module_path: path.parse(module_id) }));
+				module_ids.forEach(({ module_id, module_path }) => {
```
benmccann, line 170: I think this logic might be clearer if you instead did a `filter` to find all the matching `module_id`s. Then, after that, if there are any matching IDs you can create a hash from those IDs

dionysiusmarquis: I know this current trend to use chained es6 array functions. But to be honest, I never got why I should optimise script language iteration code for "readability" or "easy to understand". This is not an environment where a compiler will optimise this kind of unnecessary multi iteration for you. If you ask me I would only "allow" es6 array functions if they are really just a short variant of a "simple" iteration. I always encourage other devs I work with to first think of a solution with as few iterations as possible and, if needed, use additional variables to save or determine necessary states etc.
I have seen whole app code foundations becoming pretty laggy just because the extensive use of chained array functions, lately.
I can change it, but I wouldn't suggest it tbh (and I don't really want to participate coding this unnecessary multi iterate way, no matter how many data gets processed 😛)

benmccann: I didn't mean to suggest you have to use chained es6 array functions and don't really care if you want to use `for` loops or something else. My main thought here was more about the structure of the operations. It took me awhile to understand what exactly is happening. I think if it were separated into two steps of first collecting the relevant IDs and then second building the hash from those IDs it'd be a bit easier to understand compared to building the hash as we go.

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -154,11 +156,51 @@ export default class RollupCompiler {
...
+		 */
+		const css_hashing = () => {
+			const css_hashes: Record<string, string> = {};
+			return <Plugin>{
+				name: 'sapper-css-hashing',
+				augmentChunkHash(this: PluginContext, chunk: PreRenderedChunk) {
+					let hash: Hash;
+					const module_files = Object.keys(chunk.modules).map(module_id => {
+						/**
+						 * Save `module_file` which represents the id without file extension. This way it's easy to determine
```
benmccann, line 170: I'd probably just use `//` style comments for variables and reserve the other style for method declarations

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -154,11 +156,51 @@ export default class RollupCompiler {
...
-		if (!/[\\/]client\./.test(entry_point)) {
-			return mod;
-		}
+		/**
+		 * Track css code hashes for proper chunk hashing
+		 */
+		const css_hashing = () => {
+			const css_hashes: Record<string, string> = {};
+			return <Plugin>{
+				name: 'sapper-css-hashing',
```
benmccann, line 165: Can this plugin be combined with the `sapper-css-injection` plugin? It seems like the hashing is really necessary because of what that plugin is doing and isn't something we'd use in a standalone fashion

zaxwed-8huzca-timrYq (approved):

src/core/create_compilers/RollupCompiler.ts
dionysiusmarquis: It needs to be before the css chunk plugin and is also critical for the client chunk.

src/core/create_compilers/RollupCompiler.ts
dionysiusmarquis: My main point was to prefer a single iteration approach via variables (in this case `hash`) over multi iteration. I will change it, but I‘m still not a fan of using additional unnecessary iterations to make it "easier to understand".

benmccann: thanks. I think the number of elements being iterated over here will be quite limited, so there should not be any measurable performance impact, which is why I prefer the readability approach

dionysiusmarquis: Ok, done 😬

src/core/create_compilers/RollupCompiler.ts
dionysiusmarquis: 🤔 Actually this exact issue should also apply to `rollup-plugin-css-chunks` as there is also some `chunk.code` stuff happening without a corresponding `augmentChunkHash` hook. So `sapper-css-hashing` should actually fix this for the css chunk plugin as well as the sapper rollup compiler.

benmccann: I was hoping to combine the `sapper-css-injection` plugin with `rollup-plugin-css-chunk`: https://github.com/domingues/rollup-plugin-css-chunks/pull/8. There's no reason that our css injection functionality couldn't be used by other Rollup projects, so I think it'd better live as a project separate from Sapper

dionysiusmarquis: Thinking about it this maybe could have been solved on `rollup-plugin-css-chunks` level. I already tried to update to the newest version, but that had a major update in the mean time. Using the newest version in sapper had noticeable side effects. Fix it there might still need extra work on sapper side, I guess.

benmccann: I just sent https://github.com/sveltejs/sapper/pull/1666 to update to the latest `rollup-plugin-css-chunks`. I had to bump Rollup as well to make it work

src/core/create_compilers/RollupCompiler.ts
```diff
@@ -1,3 +1,4 @@
+import { createHash, Hash } from 'crypto';
```
benmccann, line 1: `lint` is complaining that `Hash` is unused

benmccann:
It looks like this will need to be rebased. I sent a small PR in https://github.com/sveltejs/sapper/pull/1677 and didn't realize it would conflict with this

dionysiusmarquis:
@benmccann This approach doesn't work anymore, since the update of `rollup-plugin-css-chunks` (https://github.com/sveltejs/sapper/commit/8a2769da8ed95c3daacfadb645946fc9822be4b1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R63).

That's because the css files won't be present in the`chunk.modules` anymore. This is most likely because of this change: https://github.com/domingues/rollup-plugin-css-chunks/commit/3632de47adea7a0447e0d53b0a608cecac1307c8#diff-dcdc3e0b3362edb8fec2a51d3fa51f8fb8af8f70247e06d9887fa934834c9122L90

Following: https://github.com/rollup/rollup/issues/3655

dionysiusmarquis:
@benmccann Ok, I rebased and updated the approach. Maybe you could have another look. I had to add `moduleSideEffects: 'no-treeshake'`. I was also able to simplify the `augmentChunkHash` part because I had to determine `is_svelte_css` in `transform` already.

benmccann:
It looks like there's a merge conflict here. This one was pretty complicated to understand and test. I wonder if it might be better to just make sure this works as intended in SvelteKit instead? SvelteKit uses Vite for CSS bundling which is a completely different set of code, so there's a good chance this issue doesn't exist there

dionysiusmarquis:
Everything is working as expected in SvelteKit, I'd say. Running `dev` imported css inside `<script>` will be added "unbundled" as inline css. `build` will put them in the corresponding bundled css, correctly hashed.

Issue #1564, opened 2020-09-25
"phantom" modules appearing with recent rollup versions.


antony:
**Describe the bug**

Essentially - with the latest version and recent rollup, Sapper generates a zero byte js chunk for a component that is exported with other components from a node module, even if it is not used in the application. It occurs where those other components are used.

If the component *is* used, two chunks with the same name are generated, with different hashes, one of which is 1 byte.

The chunk simply contains a newline.

**Logs**
From the console after `npm run build`:

```
667 B notifier.da518abc.js
           └ node_modules/@beyonk/svelte-notifications/src/notifier.js
       1 B ColourPicker.a049e84d.js
   5.37 kB XIcon.d215763e.js
```

Note the `1 B`yte module which has no contents. The `ColourPicker` module is never imported in the app, so shouldn't generate a chunk.

**To Reproduce**

A small reproduction. The module which makes the issue most obvious is private so I've stuck a copy of it in `node_modules` directly for testing.

Simply inspect the code, see that `ColourPicker` is never imported nor used, and is simply the last `import` in `fields.js` of that module, and then `npm run build` to see an empty (1 byte) js chunk generated for it.

Note that reordering the imports changes this behaviour - the last imported module is always created as an empty file - if it happens to be used in the project, then you get two chunks with the same name and different hashes, one of which is empty.

https://github.com/antony/phantom-modules

**Expected behavior**
A chunk is not generated if the module is not used (tree shaken)

**Information about your Sapper Installation:**

- The output of `npx envinfo --system --npmPackages svelte,sapper,rollup,webpack --binaries --browsers`

  System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz
    Memory: 658.07 MB / 7.68 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
    Yarn: 1.22.4 - ~/.nvm/versions/node/v12.18.3/bin/yarn
    npm: 6.14.6 - ~/.nvm/versions/node/v12.18.3/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Firefox: 80.0.1
  npmPackages:
    rollup: ^2.28.2 => 2.28.2 
    sapper: ^0.28.9 => 0.28.9 
[...]

Issue #1545, opened 2020-09-21
0.29 planning


benmccann:
I would like to propose the following breaking changes for 0.29:
* Remove `segment` prop from layouts (https://github.com/sveltejs/sapper/issues/824)
* Fix https://github.com/sveltejs/sapper/issues/1441 / https://github.com/sveltejs/sapper/issues/1538 / https://github.com/sveltejs/sapper/issues/1692. One possible fix might be to revert https://github.com/sveltejs/sapper/pull/1244 though I'm not quite sure how the `service-worker.js` in the template would need to be updated. Or possibly we could prepend the basepath in the manifest which would not be a breaking change and would not require the template to be updated. That might be easier/safer, but could result in a slightly larger manifest (though I'm not sure it'd matter after compression)
* ~~Remove Rollup CSS handling plugin. It's now nicely contained in a plugin that doesn't touch any other code. I recently 
converted [rollup-plugin-css-chunks](https://github.com/domingues/rollup-plugin-css-chunks) to TypeScript and would like to take a look at integrating our plugin there. See the RFC for full details: https://github.com/sveltejs/rfcs/pull/29~~ - SvelteKit as planned will not allow this. Better to keep Sapper aligned
* DONE - Upgrade `rollup-plugin-css-chunks` (https://github.com/sveltejs/sapper/pull/1666)
* DONE - `sapper:prefetch` (https://github.com/sveltejs/sapper/pull/1566)

Conduitry:
I believe there are people currently using the fact that `segment` is undefined on error pages as a workaround for detecting that situation, so I think it would be nice to expose this information officially on the `$page` store in the same version as we remove `segment`. There's a PR open for this, but I have not looked at what it does or how it does it.

benmccann:
Ah, good call out. That PR looks pretty solid. I had suggested at one point earlier that we include `status` there as well, but am now wondering if I didn't think that all the way through and it's not a good idea. E.g. it's probably not relevant in client-side navigation. I'd be interested in what you think: https://github.com/sveltejs/sapper/pull/1498/files#r492144537. Other than that it looks pretty good to me

UltraCakeBakery:
Please leave #824 open for discussion. Maybe even move this to the next release. I think we should first see some real world examples, side by side, to make sure this change doesn't unnecessarily increase the codebase of your average (and/or more complex) sapper application.

TheComputerM:
[...]

PR #1666, opened 2020-12-10
Upgrade rollup-plugin-css-chunks

benmccann:
Fixes https://github.com/sveltejs/sapper/issues/1564

This is a breaking change, so the next release would have to be Sapper 0.29.0. See https://github.com/domingues/rollup-plugin-css-chunks/pull/17


Conduitry:
Is this pending the merge and release of https://github.com/domingues/rollup-plugin-css-chunks/pull/17 ? Would that then be version 3 of that plugin? Or was version 2 the breaking change, and that PR is just fixing the peerdep ranges to match what it already requires?

benmccann:
Version 2 was the breaking change and that PR is just fixing the `peerDependencies`

benmccann:
@Conduitry are you okay if I merge this and we make the next release 0.29 with this breaking change? I'd like to get some more eyes on the new version of the css plugin since it's used by SvelteKit as well

Conduitry:
I'm fine with this change, but are there other breaking changes we should get cued up for a new version?

benmccann:
There were a few other breaking changes I was originally considering before SvelteKit came along: https://github.com/sveltejs/sapper/issues/1545.
There's no PR available for the `segment` prop change, so I'm not sure that's something that would get done.
Reverting #1244 and merging #1566 probably make sense though

PR #1668, opened 2020-12-11
Fixes #1667 prefetching fails if clicking quickly back and forth

### LINKED ISSUE #1667: Following a prefetched link will fail if switching quickly back and forth
**Describe the bug**
If, starting from a page A you click on a link to a page B and then - before B loads - hovering over a preloading link to A, then you will not be able to follow the link to A once B finishes loading. The URL will update, but the previous page will remain active.

**To Reproduce**
In https://github.com/ehrencrona/sapper-prefetch-caching-repro is the Sapper template but with a preload for the home page and for the blog page that takes 2 s to execute.

Go to `/blog` and reload. Then in quick succession click home, blog and home again. Wait two seconds. Now click blog. The URL will update but you will still be on the home page.

**Expected behavior**
I expect the blog to render.

**Severity**
Moderate; occurs on real public sites when preloading is slow.

ehrencrona:
Fixes #1667 

When prefetching a page that is the current page, no props got stored in `prefetching`

~I'm not entirely sure how to best write a test for this. Any pointers?~ Managed to add a test. It's a bit brittle (in the sense that if the prefetching logic changes it might pass even if the bug were to be reintroduced, not in the sense of sporadically failing) but I don't see any better way to handle it.

This is the kind of thing that unit tests might be better at, but the routing/preloading logic has so much global state that's pretty much impossible.

### Tests

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -240,8 +240,7 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 
 			if (!session_dirty && !segment_dirty && current_branch[i] && current_branch[i].part === part.i) {
 				result = current_branch[i];
-			}
-			else {
+			} else {
 				segment_dirty = false;
 	
 				const { default: component, preload } = await components[part.i].js();
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -236,29 +236,35 @@ export async function hydrate_target(dest: Target): Promise<HydratedTarget> {
 
 			const j = l++;
 
+			let result;
+
 			if (!session_dirty && !segment_dirty && current_branch[i] && current_branch[i].part === part.i) {
-				return current_branch[i];
+				result = current_branch[i];
+			}
+			else {
+				segment_dirty = false;
+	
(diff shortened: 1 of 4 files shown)


benmccann:
Do we need to port this fix to Kit as well? https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/internal/renderer/index.js

ehrencrona:
> Do we need to port this fix to Kit as well? https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/internal/renderer/index.js

Good point. Probably, though the code has been rewritten so it's not obvious whether the issue still occurs. I'll try to find time to test that.

ehrencrona:
> > Do we need to port this fix to Kit as well? https://github.com/sveltejs/kit/blob/master/packages/kit/src/runtime/internal/renderer/index.js
> 
> Good point. Probably, though the code has been rewritten so it's not obvious whether the issue still occurs. I'll try to find time to test that.

Update: I ported the test to Kit and it passes without any code changes, so I assume the fix does not need porting. Though I haven't investigated why in detail.

PR #1686, opened 2021-01-01
Update 04-preloading.md

babakfp:
Some weird sentence. It was hard to understand.

### DIFF
--- a/site/content/docs/04-preloading.md
+++ b/site/content/docs/04-preloading.md
@@ -2,13 +2,13 @@
 title: Preloading
 ---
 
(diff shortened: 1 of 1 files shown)


benmccann:
Thanks for the effort. However, this introduces a number of grammar issues, incorrect usage of commas, etc. and I don't see any issues with the current wording for the most part. Additionally, I've already copied and updated all the docs for SvelteKit and would need to port this change there as well. I'm going to go ahead and close this since it's not clear to me what it's fixing and doesn't seem worth further time investment

PR #1695, opened 2021-01-12
Support for dynamically setting the `lang` attribute

ehrencrona:
Currently the `lang` attribute in the `html` tag is hard-coded in `template.html`. This makes it impossible to build multi-language sites with Sapper.

This PR allows routes to export a `lang` function in addition to `preload`. It gets the same path information as `preload` and returns a language code to use for the page. The most specific component wins.

See discussion in #576 

I know Sapper is currently in maintenance mode and that this is tied up in the much larger internationalization question, but this feels like a fairly straightforward extension. At Hyperlab we are currently building two multi-language sites. Without the ability to determine the language dynamically we would need to fork Sapper. 

### Tests

### DIFF
--- a/runtime/src/app/app.ts
+++ b/runtime/src/app/app.ts
@@ -16,6 +16,7 @@ import {
 	HydratedTarget,
 	Target,
 	Redirect,
+	BranchSegment,
 	Branch,
 	Page,
 	InitialData
@@ -136,6 +137,11 @@ async function handle_target(dest: Target): Promise<void> {
 		const { props, branch } = hydrated_target;
 		await render(branch, props, buildPageContext(props, dest.page));
 	}
+
+	
+	const { lang } = hydrated_target;
+
+	if (lang) document.querySelector('html').setAttribute('lang', lang);
 }
 
 async function render(branch: Branch, props: any, page: PageContext) {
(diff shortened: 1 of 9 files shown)


benmccann:
There are three PRs open for dynamically updating the template (most recently https://github.com/sveltejs/sapper/pull/1642). I wonder if coming up with a good solution there might be a more generic solution to a wider class of issues

benmccann:
Actually, do we even need to do anything or is this already supported?

```
<svelte:head>
	<html lang="sw" />
</svelte:head>
```

ehrencrona:
@benmccann That will create an `html` tag inside `head`, but it's the very top-level `html` tag we want to modify. If it wasn't, it would indeed be easier.

benmccann:
Ah, you're right. I saw someone posted that as an example, so assumed it worked 😛 How about doing something like this?

```
onMount(() => {
  document.documentElement.lang = 'sw';
});
```

ehrencrona:
> How about doing something like this?
> 
> ```
> onMount(() => {
>   document.documentElement.lang = 'sw';
> });
> ```

That will work on the client side, but for the benefit of Google indexing in the correct language we need to be able to serve the correct language from the server side.

I guess it would be another road towards implementing support, though, to offer a `document` object on the server side, which per se might not be unreasonable.

benmccann:
Thanks for this PR!! I don't think we'll be adding anything new to Sapper at this point, but this now exists in SvelteKit! https://github.com/sveltejs/kit/pull/670

PR #1697, opened 2021-01-14
Fix scroll position is not restored after changing search param with `goto`

### LINKED ISSUE #1321: Scroll position is not restored after changing search param with `goto`
**Describe the bug**
This is slightly difficult to explain, so my apologies if it's a little bit confusing. I wrote test cases linked below that will help understanding.

Normally, when either clicking a link or using `goto`, the scroll position before the navigation event occurs is cached. On popstate (back/forward), Sapper looks up the destination page in the cache and restores using scrollTo.

It seems that if you use `goto` to change the search params (but leave the pathname the same - common when applying filters or a search query for example), the scroll position is no longer recorded before navigating away from the page. When a popstate event occurs, the scroll position is restored to the point it was at _before the search params were changed_.

So, for example:

1. Visit a page
2. Scroll down 500px and click a button that appends `?q=cheese` to the URL using `goto` (your scrollY will now be 0)
3. Scroll down 1000px and click a link.
4. When on the next page, press back in your browser.
5. 🐛 Your scrollY will be at 500px, instead of at 1000px.

I have tested this (link to test cases below), and the above works as expected as long as the button in step 2 navigates to a different pathname, instead of only changing the query param.

**To Reproduce**
I have written test cases here:
https://github.com/sveltejs/sapper/compare/master...Jayphen:scrollY-after-search-param-change

The first test (restores scroll on popstate events) passes. I couldn't find any other tests in the Sapper repo that actually test scroll restoration on popstate, so I wrote this one.

The second test ('restores scroll on popstate events preceded by search param changes') fails. I've included some comments to hopefully explain what's happening.

**Severity**
Quite severe - we won't launch our ecommerce project until this is fixed. This results in losing your scroll position when viewing a product from a search result page and then going back, which is bad UX.

istarkov:
fixes #1321 

goto must have same order of execution as [handle_click](https://github.com/istarkov/sapper/blob/5757a85d195f9ceb2e329caa5e3e38abd5a1123d/runtime/src/app/router/index.ts#L164-L166)

### DIFF
--- a/test/apps/scroll/test.ts
+++ b/test/apps/scroll/test.ts
@@ -2,7 +2,7 @@ import * as assert from 'assert';
 import { build } from '../../../api';
 import { AppRunner } from '../AppRunner';
 
-describe('scroll', function () {
+describe('scroll', function() {
 	this.timeout(10000);
 
 	let r: AppRunner;
--- a/test/apps/scroll/test.ts
+++ b/test/apps/scroll/test.ts
@@ -2,7 +2,7 @@ import * as assert from 'assert';
 import { build } from '../../../api';
 import { AppRunner } from '../AppRunner';
 
-describe('scroll', function() {
+describe('scroll', function () {
 	this.timeout(10000);
 
 	let r: AppRunner;
@@ -50,10 +50,7 @@ describe('scroll', function() {
 		await r.page.click('[href="another-tall-page"]');
 		await r.wait();
 
-		assert.strictEqual(
-			await r.page.evaluate(() => window.scrollY),
-			0
-		);
+		assert.strictEqual(await r.page.evaluate(() => window.scrollY), 0);
 	});
 
 	it('preserves scroll when a link with sapper:noscroll is clicked', async () => {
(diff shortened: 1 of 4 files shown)


istarkov:
tests are added from the corresponded issue

istarkov:
timeout on ci test, not somehow related to this PR

benmccann:
If we update this we'll also need to take a look at updating SvelteKit which shares this code:
https://github.com/sveltejs/kit/blob/6668f2d9e9edae2430a5616679883f68c165fd62/packages/kit/src/runtime/internal/router/index.js#L168

test/apps/scroll/test.ts
```diff
@@ -2,7 +2,7 @@ import * as assert from 'assert';
 import { build } from '../../../api';
 import { AppRunner } from '../AppRunner';
 
-describe('scroll', function() {
+describe('scroll', function () {
```
benmccann, line 5:
```suggestion
describe('scroll', function() {
```

benmccann:
I was not able to reproduce this in SvelteKit, so no changes should be necessary there

PR #1709, opened 2021-01-29
Fix bad imports in type definitions

benmccann:


runtime/src/internal/manifest-server.d.ts
```diff
@@ -1,12 +1,11 @@
...
-} from './shared';
+import { Preload } from '@sapper/common';
 
 export const src_dir: string;
 export const build_dir: string;
 export const dev: boolean;
 export const manifest: Manifest;
 
+import { SapperRequest, SapperResponse } from '@sapper/server';
 export { SapperRequest, SapperResponse } from '@sapper/server';
```
dummdidumm, line 9: You can omit the `from`-part if you imported those already.
```suggestion
export { SapperRequest, SapperResponse };
```

benmccann: thanks! yes, that's better!

PR #1710, opened 2021-01-29
Make type definitions compatible with Express and Polka

benmccann:
Closes #1706 

The type definition for `SapperRequest` and `SapperResponse` is totally borked. It won't work with Express or Polka. This changes it to use the Express types instead which work on both.

I ignored several issues in the codebase as a result of this change. That's probably not the best solution, but everything's been working just fine and it's not worth investing more time with Sapper eventually going away. My priority is just getting Sapper unbroken

I left the `SapperRequest` and `SapperResponse` types instead of replacing them with the Express types in order to maintain backwards compatibility

PR #1713, opened 2021-02-01
Fixed typings to be compatible with express

### LINKED ISSUE #1706: Typescript issues with v0.29.0 (tested in sveltejs/sapper-template)
**Describe the bug**
I upgraded Sapper in a new `sveltejs/sapper-template` rollup project and see Typescript errors when running `npm i && npm run dev`

**Logs**

(Repeated for client, server, and service-worker)

```
@rollup/plugin-typescript TS2769: No overload matches this call.
  Overload 1 of 2, '(...handlers: RequestHandler<ParamsDictionary, any, any, ParsedQs, Record<string, any>>[]): Polka', gave the following error.
    Argument of type '(req: SapperRequest, res: SapperResponse, next: () => void) => void' is not assignable to parameter of type 'RequestHandler<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'.
      Types of parameters 'req' and 'req' are incompatible.
        Property 'search' is missing in type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>' but required in type 'SapperRequest'.
  Overload 2 of 2, '(pattern: string | RegExp, ...handlers: RequestHandler<ParamsDictionary, any, any, ParsedQs, Record<string, any>>[] | Polka[]): Polka', gave the following error.
    Argument of type 'RequestHandler<ParamsDictionary, any, any, ParsedQs, Record<string, any>>' is not assignable to parameter of type 'string | RegExp'.
      Type 'RequestHandler<ParamsDictionary, any, any, ParsedQs, Record<string, any>>' is missing the following properties from type 'RegExp': exec, test, source, global, and 12 more.

  9 polka() // You can also use Express
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 10  .use(
    ~~~~~~
... 
 13      sapper.middleware()
    ~~~~~~~~~~~~~~~~~~~~~
 14  )
    ~~

  src/node_modules/@sapper/index.d.ts:62:3
    62      search: string | null;
            ~~~~~~
    'search' is declared here.
```

**To Reproduce**
- make a new rollup project from `sveltejs/sapper-template`
- run the Typescript conversion script
- upgrade Sapper to `v0.29.0`
- `npm i && npm run dev`, errors reported in console

```
npx envinfo --system --npmPackages svelte,sapper,rollup,webpack --binaries --browsers
npx: installed 1 in 1.233s

  System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 35.42 GB / 64.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.20.1 - ~/.nvm/versions/node/v12.20.1/bin/node
    npm: 6.14.10 - ~/.nvm/versions/node/v12.20.1/bin/npm
  Browsers:
    Chrome: 87.0.4280.141
    Firefox: 84.0.2
    Safari: 13.1.3
  npmPackages:
    rollup: ^2.3.4 => 2.38.0 
    sapper: ^0.29.0 => 0.29.0 
    svelte: ^3.17.3 => 3.31.2 
```

**Severity**
Blocking upgrade, which is important for us because it seems to fix issues we have with `preload` not firing consistently (https://github.com/sveltejs/sapper/issues/1582)

ehrencrona:
A more conservative version of #1710 that fixes #1706.

There seem to be three issues with the request:

 * the `search` property that's supposed to hold the query string doesn't actually exist. No idea how it wound up in there. It's also not referenced anywhere in the code so easy to remove.
 * `query` can contain a wider range of stuff than `SapperRequest` declares. I grabbed the Express definition but we could also do `Record<string, any>`
 * `socket` assumed it was always an `https` connection

This seems to compile with both Polka and Express when I tested it

### DIFF
--- a/runtime/index.d.ts
+++ b/runtime/index.d.ts
@@ -27,9 +27,11 @@ declare module '@sapper/app' {
 
 declare module '@sapper/server' {
 	import { IncomingMessage, ServerResponse } from 'http';
-	import { TLSSocket } from 'tls';
+	import { Socket } from 'net';
 
 	export type Ignore = string | RegExp | ((uri: string) => boolean) | Ignore[];
+
+	interface ParsedQs { [key: string]: undefined | string | string[] | ParsedQs | ParsedQs[] }
 
 	/**
 	 * The request object passed to middleware and server-side routes. 
@@ -57,16 +59,11 @@ declare module '@sapper/server' {
 		params: Record<string, string>;
 	
 		/**
-		 * The un-parsed querystring
+		 * The parsed query string
 		 */
-		search: string | null;
-	
-		/**
-		 * The parsed querystring
-		 */
-		query: Record<string, string>;
+		query: ParsedQs;
 
-		socket: TLSSocket;
+		socket: Socket;
 	}
 
 	export interface SapperResponse extends ServerResponse {
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -18,6 +18,7 @@ import {
 import App from '@sapper/internal/App.svelte';
 import { PageContext, PreloadResult } from '@sapper/common';
 import detectClientOnlyReferences from './detect_client_only_references';
+import { TLSSocket } from 'tls';
 
 export function get_page_handler(
 	manifest: Manifest,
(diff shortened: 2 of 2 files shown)


ehrencrona:
Come to think of it, #1710 is the better solution. I vaguely remember that the request interfaces were different between Polka and Express, which is the only reason there should be a `SapperRequest` in the first place. If the Express typings work (either that has changed or I was wrong), then it's better to just use them directly.

benmccann:
I actually like yours better. I'll reopen this until we figure out which one to merge

dummdidumm:
I like this one better as well, but take this opinion with a grain of salt as I don't know much of Sapper's typings. What I like more about this is that this one is more self-contained.

PR #1722, opened 2021-02-17
Add defer to script tag created by rollup build

Evertt:
#1123 added defer to the script tag when the bundler is not rollup. I think it was an oversight that it wasn't added to the other 2 lines.

### DIFF
--- a/runtime/src/server/middleware/get_page_handler.ts
+++ b/runtime/src/server/middleware/get_page_handler.ts
@@ -344,9 +344,9 @@ export function get_page_handler(
 			if (build_info.bundler === 'rollup') {
 				if (build_info.legacy_assets) {
 					const legacy_main = `${req.baseUrl}/client/legacy/${build_info.legacy_assets.main}`;
-					script += `(function(){try{eval("async function x(){}");var main="${main}"}catch(e){main="${legacy_main}"};var s=document.createElement("script");try{new Function("if(0)import('')")();s.src=main;s.type="module";s.crossOrigin="use-credentials";}catch(e){s.src="${req.baseUrl}/client/shimport@${build_info.shimport}.js";s.setAttribute("data-main",main);}document.head.appendChild(s);}());`;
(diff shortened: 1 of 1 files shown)


Evertt:
By the way, why does the rollup build get such a strange way of including the script tag?

benmccann:
It has `type=module` which I believe implies `defer`

Evertt:
> It has `type=module` which I believe implies `defer`

Nope, it doesn't. Or at least the problem I'm running into is that if I have other script tags *before* `%sapper.scripts%` that use `defer` then they are not loaded in the right order. As in sapper's scripts are then loaded first, but I need them loaded last. And with the change in my PR it does work as expected.

benmccann:
Hmm. MDN says "The defer attribute has no effect on module scripts — they defer by default.": https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

Evertt:
Well I'm using the latest version of Chrome. Maybe this is a bug in Chrome. You know what, I'll test it in other browsers and I'll get back to you in a minute.

Evertt:
Okay so it's one of those terrible bugs that sometimes pops up and sometimes doesn't. But I'd like to invite you to go to this URL: https://mytryout-246d2.web.app/

Open up your console and refresh a bunch of times without cache (cmd+shift+R on a mac). Sometimes you'll get an error that is because of scripts not being run in the right order. Do you get that too?

Evertt:
[Here's some video evidence of the issue](https://www.youtube.com/watch?v=NWEfvwrgVg4). I'm not sure what to do with it.

benmccann:
There's a perhaps related issue in https://github.com/sveltejs/sapper/issues/1725, but it was suggested there that this PR doesn't fix that issue. Does adding the `onload` handler suggested there help you with your issue at all?

Evertt:
@benmccann yes the onload handler does fix my issue. And yeah I was aware that my PR doesn't actually fix anything. It just seemed to, for a few minutes. 😅

Issue #1437, opened 2020-08-25
accessing session data from server route


tvardy:
I'm not really sure (and had not yet found it in written) how could I access the session data from inside the servers routes.
For the `fileName.svelte` routes it's clearly said how to get such data from both, the preload function, and the component's script...
But what about the `fileName.js` route? I was initially expecting to have something like `req.session` available in my `get(req, res)` or `post(req, res)` functions... but... unfortunately, it's not there... :(

antony:
Hi @tvardy - please ask support questions on [discord](https://svelte.dev/chat) as it is a better medium to help you - github issues is for bug reports and feature requests only. Thanks!

danawoodman:
@tvardy or anyone else finding this via Google (like me), have a look at this article I just put together, think it may help 🤩

https://dev.to/danawoodman/getting-session-data-in-your-sapper-server-routes-51md

PR #1749, opened 2021-03-16
Add a note about providing a middleware to set a user session

danawoodman:
For a new user it might not be clear that they'll need to provide `req.user` via a middleware, so I've added a line to make this explicit that they'll need to create some sort of middleware to set the user for the `sapper.middleware`

Relates to #1437

### DIFF
--- a/site/content/docs/07-state-management.md
+++ b/site/content/docs/07-state-management.md
@@ -27,6 +27,8 @@ As an example, let's look at how to populate the session with the current user. 
 ```js
 // src/server.js
 polka()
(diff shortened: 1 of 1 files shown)


benmccann:
Thanks. It might help some, but I'm afraid it also sort of raises the question of how to create `userMiddleware`. I don't really want to spend much more time on the Sapper docs, so maybe it's better if we just make sure the SvelteKit docs are sufficiently clear on this point

danawoodman:
Ok, up to you. Now that it is in Beta it makes sense to throw energy at it instead 👍

benmccann:
Thanks!

Issue #343, opened 2018-08-06
Security Nonce for sapper.scripts


AlexRiedler:
Issue: when we use sapper.scripts for injection (changed from sapper.main) we now no longer have access to inject nonce for scripts; maybe there is a better way to do this? (CSP policy)

Any ideas would be greatly appreciated :)

Rich-Harris:
This sounds like a good and sensible idea, though I have to confess I've never used `nonce` before 😬 Could you walk me through the process of generating/applying it? Thanks!

AlexRiedler:
Basically on the request, a UUIDv4 is normally generated and put into the CSP header like so:

HTTP Header:
`Content-Security-Policy: default-src 'self'; script-src 'nonce-4AEemGb0xJptoIGFP3Nd'`

and then when the script tag is generated it uses the nonce:

```html
<script type="text/javascript" nonce="4AEemGb0xJptoIGFP3Nd"> ... </script>
````

In javascript I have seen people use helmetjs/csp library, and do something like:

For generating UUID for nonce
```js
var uuidv4 = require('uuid/v4')

app.use(function (req, res, next) {
  res.locals.nonce = uuidv4()
  next()
})
```

and then upstream when generating the actual payload:

```js
return '<script nonce="' + res.locals.nonce + '">alert('hello world');</script>'
```


for the lengthy read troy hunt gives a good read for it: https://www.troyhunt.com/locking-down-your-website-scripts-with-csp-hashes-nonces-and-report-uri/

It basically blocks scripts from  running on the page that do not have the nonce that matches the one the the HTTP Header.

I probably have time to investigate on how to implement this; would it make more sense to somehow add parameters to %sapper.scripts% ; or would it make more sense to support this feature by configuration?

Seb35:
In the meantime this was implemented by @nolanlawson on 7 September 2018 (e3775158) and documented (sveltejs/sapper.svelte.technology@6cde0fae822affbb058697e54fb43ec68b40ad13).

I’ve just tested with Helmet and Polka on my app ([here](https://github.com/Legilibre/archeo-lex.fr/commit/d34f928488da0f7f0387ed8d0ccf9c32b1ab1f72)) and it works: a nonce is added in the CSP by Helmet and the nonce is used in %sapper.scripts%.

Is there anything else to do in this issue? -- btw thanks for this issue and the implementation, that’s an interesting feature.

ShimShamSam:
[...]

PR #1760, opened 2021-04-06
Remove use of eval in non-legacy rollup builds

seanlail:
## Description

As @maximedupre reported in https://github.com/sveltejs/sapper/issues/1622:
> This file contains `new Function` (injected in the browser), which is a form of eval, which causes a CSP error, unless adding `unsafe-eval`

Adding `unsafe-eval` to CSP would then defeat the purpose of using CSP.

It was using `new Function` to do feature detection for imports, but it looks like that is now supported by all non-legacy browsers anyway.

This PR removes the try/catch when building a non-legacy rollup bundle, thereby removing the use of shimport.
This will allow users to remove "unsafe-eval" from any CSP configuration.

I realise Sapper is in maintenance but I feel like this is a blocking issue for production usage.

Closes https://github.com/sveltejs/sapper/issues/1622
Relates to https://github.com/sveltejs/sapper/issues/343

## Test

I came across this issue originally because I added CSP + blocked unsafe-eval and noticed that reactivity did not work.
So I've added a test that does just that. Binds a text input, types in a value and then checks that the updated value is shown.
This test now passes, before the change it failed.

## Questions

- [ ] Should I add the `nonce_attr` in?
- [ ] Should I just use the webpack line? (i.e. `</script><script${nonce_attr} src="${main}" defer>`


package.json
```diff
@@ -19,6 +19,7 @@
     "test": "test"
   },
   "dependencies": {
+    "helmet": "4.4.1",
```
benmccann, line 22: I think we've generally put the test dependencies in `devDependencies`

runtime/src/server/middleware/get_page_handler.ts
```diff
@@ -346,7 +346,7 @@ export function get_page_handler(
 					const legacy_main = `${req.baseUrl}/client/legacy/${build_info.legacy_assets.main}`;
 					script += `(function(){try{eval("async function x(){}");var main="${main}"}catch(e){main="${legacy_main}"};var s=document.createElement("script");try{new Function("if(0)import('')")();s.src=main;s.type="module";s.crossOrigin="use-credentials";}catch(e){s.src="${req.baseUrl}/client/shimport@${build_info.shimport}.js";s.setAttribute("data-main",main);}document.head.appendChild(s);}());`;
 				} else {
-					script += `var s=document.createElement("script");try{new Function("if(0)import('')")();s.src="${main}";s.type="module";s.crossOrigin="use-credentials";}catch(e){s.src="${req.baseUrl}/client/shimport@${build_info.shimport}.js";s.setAttribute("data-main","${main}")}document.head.appendChild(s)`;
```
benmccann, line 349: what was this `try` / `catch` trying to detect? I feel like it was to see what browser you were using and can't simply be removed

Conduitry:
I'd be wary about making this without it being a breaking change. Are there, say, any old version of Chromium on old web TVs that support async/await but do not support ESM? I would imagine so, and I know there are people who use Sapper for these environments.

runtime/src/server/middleware/get_page_handler.ts
seanlail: Looks like it was trying to detect "import" support to add the module script, otherwise it added "shimport".

Conduitry: If there's a way to detect ESM/`import`/`import()` support without a try-catch, then I'd be all for switching to that. Otherwise, as I indicated below, I don't think we can just do this without it being a breaking change. (And it'd be a breaking change that I don't think could be easily worked around for people that need to support these browsers. They'd be stuck on the old version.)

seanlail:
> I'd be wary about making this without it being a breaking change. Are there, say, any old version of Chromium on old web TVs that support async/await but do not support ESM? I would imagine so, and I know there are people who use Sapper for these environments.

@Conduitry Good point, this would be a breaking change. If those people have _not_ targetted legacy then `shimport` could be saving them at the moment.

Could we rewrite that try/catch then to be CSP safe? 🤔

runtime/src/server/middleware/get_page_handler.ts
seanlail: @Conduitry Agreed. I'll have a look 👍

seanlail: Actually, can't we use `nomodule` for backwards compat?

Kapsonfire-DE:
```'noModule' in HTMLScriptElement.prototype```
can be used without try catch to detect module functionality

Conduitry:
Does that safely check for _dynamic_ import support as well? There was a while where Firefox supported `import`/`export` by not `import()`. We still need to bring in Shimport if the browser supports `import ... from ...` but not `import(...)`.

seanlail:
Unfortunately I think 'new Function' is the way to test dynamic.

Could we solve this another way? 

Default to using shimport and then expose a flag for people people to skip/disable it?

It's like the opposite of legacy though.

Conduitry:
If there's not a reasonable way to test for dynamic imports, then I think at this point in Sapper's life, the answer for how to do this in a CSP-compatible way is to use a fork of Sapper. There's not a good way to expose options like this without there being an overbearing number of CLI options.

I would encourage you to follow along with SvelteKit and, when we get to a point where we begin to think about differential legacy builds there, to advocate for whatever features you think are necessary to keep CSP happy.

Kapsonfire-DE:
But whats wrong with try catch?

Kapsonfire-DE:
```try { import('').catch(ex => {}); } catch(ex) { console.log('dynamic export not supported'); }```
I mean this works perfectly without any console output

Conduitry:
I haven't checked, but won't `import('')` be a parsing error on some browsers? If it won't even parse, this won't work, and none of the code will run. The eval was to get around stuff not parsing.

Kapsonfire-DE:
either dynamic import exists, then it returns a promise and its error is catched with .catch(ex), or import doesnt exists and its throws a default error catched by try-catch... if import('') really does a parsing error, then you can use any valid string here
import('notExistant.js')
still same result

Kapsonfire-DE:
oh you are right - giving syntax error on IE

seanlail:
@Kapsonfire-DE @Conduitry Thanks for digging in to this.

I have added a comment to 2 SvelteKit issues and will follow the progress.

I'm going to fork Sapper I think to get around the issue for now, as like you said there won't be many changes to Sapper anyway and hopefully we can move to Kit soon.

Thanks for looking at this anyway 👍

PR #1780, opened 2021-06-14
Make prefetching styles work more reliably

### LINKED ISSUE #1582: Nested directory structure with absent layout causes preload to get stuck at that route level
**Describe the bug**
If you have a series of nested directories in your routes directory, and one of those directories doesn't have its own `_layout.svelte`, and you attempt to navigate to the route *after* that directory, then preload gets stuck, the browser URL changes, but preloads stop being executed (and the preload store gets "stuck" in preloading state) at that point and the application freezes.

**Additional context**

** I've been able to produce this without a missing layout - so further investigation is being done - it may be more serious than I first thought as I can't work around it now **

1. This only happens on the client-side. The server side works as expected.
1. This only happens the first time you try it. If you then go back and try again, the routes load as expected.
1. Direct visits to the problematic route work without issue (i.e. this affects goto, and direct links)
1. This bug first appeared in Sapper 0.28.6

**Logs**
Server:

```
outside preload
innermost preload
account route preload
```

Client:
```
account route preload
_layout.svelte:3 innermost preload
```

**To Reproduce**

https://github.com/antony/sapper-incongruous-layouts-issue

Run app and click `checkout`.

**Expected behavior**
Preload should work

```
System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz
    Memory: 329.94 MB / 7.68 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
    Yarn: 1.22.4 - ~/.nvm/versions/node/v12.18.3/bin/yarn
    npm: 6.14.6 - ~/.nvm/versions/node/v12.18.3/bin/npm
  Browsers:
    Chrome: 85.0.4183.121
    Firefox: 81.0
  npmPackages:
    rollup: ^2.3.4 => 2.28.2 
    sapper: ^0.28.0 => 0.28.9 
    svelte: ^3.17.3 => 3.28.0 
```

**Severity**
Critical. This is completely unexpected behaviour for new users. There is a workaround but it would depend on finding this issue.

wlach:
There was a small race between the element being added to the DOM
and an onload event being triggered, fix this by making sure we have an
onload/onerror handlers set up before adding it to document head. Fixes #1582.

NOTE: All development is currently focused on Sapper's successor, SvelteKit: https://kit.svelte.dev/
We are not adding new features to Sapper. While we might fix some particularly pressing or easy to
address bugs, it's highly unlikely that most PRs will be reviewed.

### Tests

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -45,13 +45,10 @@ export default function(files) {
 			link = document.createElement('link');
 			link.rel = 'stylesheet';
 			link.href = href;
+			link.onload = link.onerror = fulfil;
 			document.head.appendChild(link);
-		}
-		if (link.sheet) {
-			fulfil();
 		} else {
-			link.onload = function() { return fulfil() };
-			link.onerror = reject;
+			fulfil();
 		}
 	})}));
 };`.trim();


src/core/create_compilers/RollupCompiler.ts
```diff
@@ -41,18 +41,16 @@ export default function(files) {
...
 			|| document.querySelector('link[rel=stylesheet][href="' + href + '"]');
-		if (!link) {
+		if (link) {
+			fulfil();
+		} else {
 			link = document.createElement('link');
 			link.rel = 'stylesheet';
 			link.href = href;
+			link.onload = fulfil;
+			link.onerror = fulfil;
```
wlach, line 51: I changed this behaviour slightly because it seems better to resolve the promise in the case that we can't load the style than just reject (which will make the application appear to hang).

wlach:
(there's something wrong here, it doesn't fix the problem reliably)

wlach:
Ok on further testing I think this does actually work as expected (I had some difficulty setting up a good test environment).

Open the following in a chrome private tab:

https://irydium.netlify.app/

Vs. the following:

https://deploy-preview-36--irydium.netlify.app/

With the first example, navigating to "examples" doesn't always work. I can't reproduce it 100% reliably but believe this is correct. It's actually a little unclear to me why we're resolving the stylesheet loads as promises at all-- we can't really take any action if they fail.

wlach:
Further refined this to handle the case where a CSS element was already loaded (which would result in the promise never resolving). I believe this might be related to this kludge in the repl (I had a similar problem to what this is attempting to fix):

https://github.com/sveltejs/svelte-repl/blob/2d0ecdf288c117f93a208efaaf04f2e38240fb86/src/Repl.svelte#L99

benmccann:
Is there a reason you don't just use SvelteKit instead?

wlach:
> This would need to be rebased, but is there a reason you don't just use SvelteKit instead? I don't really want to spend time on these edge cases for a deprecated project

Porting to SvelteKit is on my roadmap, though my project does some rather esoteric things via rollup that I'm not sure offhand how to do with SvelteKit so I've been holding off on porting things over.

I'd describe this behaviour more as a bug than an edge case, but I understand if it's not a priority to test and merge. I wound up publishing a fork for myself to hold myself over until I can find the time to port to SvelteKit: https://www.npmjs.com/package/@wlach/sapper

benmccann:

PR #1790, opened 2021-07-12
replace onload with addEventListener in inject_styles

ankitstarski:
When multiple dynamically imported components are loaded, only one of them resolves the promise. Upon further investigation, I found out that those components were loading `.css` files and `inject_styles` function was using `onload` in order to resolve `inject_styles` promise.

If two components use the same `css` file, the first `onload` will be overridden by the second one.

The solution is to replace `onload` with  `addEventListener("load", ...`.

@Rich-Harris  pushed a fix for this in svelte kit ([kit#347](https://github.com/sveltejs/kit/pull/347))

### DIFF
--- a/src/core/create_compilers/RollupCompiler.ts
+++ b/src/core/create_compilers/RollupCompiler.ts
@@ -50,8 +50,8 @@ export default function(files) {
 		if (link.sheet) {
 			fulfil();
 		} else {
-			link.onload = function() { return fulfil() };
-			link.onerror = reject;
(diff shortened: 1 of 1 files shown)


benmccann (approved):
I'm not merging much into Sapper, but this one looks fine since it aligns the code with SvelteKit

benmccann:
Although that code change was from before SvelteKit switched to Vite. That code no longer exists in SvelteKit as Vite is now responsible for it. That being said I'll go ahead and merge this since it looks like Rich approves of the change

PR #1795, opened 2021-09-06
Clean serve logs

babakfp:
Hi
I just made those stuff clean and pretty. That's all😉🙌🏻.

Screenshot 2021-09-06 084806
Screenshot 2021-09-06 084851


benmccann:
Sapper is in maintenance mode and all development is focused on SvelteKit. We'd only merge changes to Sapper for things that are absolutely necessary like security issues. Adding emojis to the console will not qualify :smile:

